PR #28, opened 2018-05-21
convert submodules to https across everything

sagark:
Ready to merge.


sagark:
@davidbiancolin ready for review

I tested on a fresh machine with no keys setup, it works as expected. Also updated docs.

PR #49, opened 2018-07-19
Model clock-domain crossings between transformed-RTL model and endpoints

davidbiancolin:
Associated midas PR: ucb-bar/midas#90. See that PR for more detail.

As for FireSim specific stuff:
- Adds Platform Configs for building simulators with clock divison between RC and DRAM models
- Add recipes to run midas synthesizable unittests
- Adds unittests to scala tests 
Some unrelated changes that could be dropped:
- Adds an submodule init only mode to build setup. 
  - this is a little hack for speeding up setup on millennium. I continue to have problems with the fast install toolchain working -- i need to full install (Should probably open an issue). 

Ultimately we probably want to model 1:3 or 1:2 division between Rocket Chip and the DRAM model by default but i'm seeing some weird behavior in the QC-FRFCFS instances that must be addressed before that can even be considered.


davidbiancolin:
@zhemao @sagark this is ready for review

sagark:
Re: fast-install toolchain, I don't think we expect it to work on millennium. We'd have to have a separate package for that.

PR #52, opened 2018-07-22
Bump upstream aws-fpga 1.4.0

sagark:
This is mostly done and works on the FPGA for the simple designs I've tested (rocket chip, singlecore, nic + no nic). This should be merged to master in the beginning of August, since Amazon is **removing** support of the current shell on September 1st.

XDMA conversion is not included here, since it needs to be tested more extensively, but lives on a separate branch. EDMA is still the default and works fine.

This should probably trigger a major version release (FireSim 1.3.0).

TODOs for merge to dev:

- [x] Regenerate all images (in progress)
- [ ] Notify dev branch users of requirement to upgrade AMI
- [x] Do we want a dev branch of aws-fpga also?

TODOs for merge to master (copy this into the WIP dev -> master PR once that is created):
- [ ] We should notify users by email in advance about the fact that they will need to upgrade any existing instances to the new 1.4.0 AMI.
- [ ] Need to test XSim, VCS
- [ ] Mark which issues it resolves
- [ ] Merge branch (dev or other) to master of aws-fpga-firesim

### DIFF
--- a/docs/Initial-Setup/Setting-up-your-Manager-Instance.rst
+++ b/docs/Initial-Setup/Setting-up-your-Manager-Instance.rst
@@ -22,7 +22,7 @@ To launch a manager instance, follow these steps:
    data is preserved when you stop/start the instance, and your data is
    not lost when pricing spikes on the spot market.
 2. When prompted to select an AMI, search in the ``Community AMIs`` tab for
-   "FPGA" and select the option that starts with ``FPGA Developer AMI - 1.3.5``.
+   "FPGA" and select the option that starts with ``FPGA Developer AMI - 1.4.0``.
    **DO NOT USE ANY OTHER VERSION.**
 3. When prompted to choose an instance type, select the instance type of
    your choosing. A good choice is a ``c4.4xlarge``.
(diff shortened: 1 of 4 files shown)


davidbiancolin:
Yeah, we probably want dev branches of the key submodules; merges into master from dev should just be reflected on the submodules i think.

sagark:
ok this is ready for review

docs/Initial-Setup/Setting-up-your-Manager-Instance.rst
```diff
@@ -22,7 +22,7 @@ To launch a manager instance, follow these steps:
    data is preserved when you stop/start the instance, and your data is
    not lost when pricing spikes on the spot market.
 2. When prompted to select an AMI, search in the ``Community AMIs`` tab for
-   "FPGA" and select the option that starts with ``FPGA Developer AMI - 1.3.5``.
+   "FPGA" and select the option that starts with ``FPGA Developer AMI - 1.4.0``.
```
davidbiancolin: Would it be easier just to check for this?

sagark: I think the aws-fpga scripts already check, at least for this change.

PR #54, opened 2018-07-30
Fix XSim and FPGA-level VCS simulation for 1.4.0 shell

sagark:
Ready for review/merge into dev.

### DIFF
--- a/scripts/setup-vcsmx-env.sh
+++ b/scripts/setup-vcsmx-env.sh
@@ -3,10 +3,10 @@
 # This script sets up VCS for FPGA-level RTL simulation on the Millennium
 # cluster at Berkeley
 
-firesim_root=$(dirname $0)/..
+firesim_root=$(pwd)/../
 cwd=$(pwd)
 
-source /ecad/tools/xilinx/SDx/2017.1.op/settings64.sh > /dev/null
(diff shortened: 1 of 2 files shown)


scripts/setup-vcsmx-env.sh
```diff
@@ -3,10 +3,10 @@
 # This script sets up VCS for FPGA-level RTL simulation on the Millennium
 # cluster at Berkeley
 
-firesim_root=$(dirname $0)/..
+firesim_root=$(pwd)/../
 cwd=$(pwd)
 
-source /ecad/tools/xilinx/SDx/2017.1.op/settings64.sh > /dev/null
+source /ecad/tools/xilinx/Vivado/2017.4/settings64.sh > /dev/null
```
davidbiancolin: You didn't need SDx for this? Interesting.

davidbiancolin:
Though we should probably have reviews on aws-fpga-firesim changes too.

scripts/setup-vcsmx-env.sh
sagark: Yeah

sagark:
Yeah, I think we can do that as part of the dev -> master merge? I don't want to create too many branches lol

PR #55, opened 2018-08-02
LoadMem improvements and DRAM initialization

davidbiancolin:
This PR improves loadmem performance by reducing the number of driver-writes per host-beat, from 4 to ~2 which should help with Shell-level RTL simulation speed. 

It also adds a feature to zero-out FPGA DRAM completely, which can be invoked by passing "+zero-out-dram" to the driver. When set, this enables a mode in the loadmem unit that requires no further writes from the driver. Instead, it issues multi-beat writes to the host-memory system as rapidly as it will accept them across the available address space. This takes about ~10 seconds to zero out a 16 GiB channel. 

This changes the simulation memory map, so this driver change may not be compatible with existing AGFIs. How should we proceed?

### DIFF
--- a/deploy/runtools/runtime_config.py
+++ b/deploy/runtools/runtime_config.py
@@ -86,7 +86,7 @@ class RuntimeHWConfig:
         # the sed is in there to get rid of newlines in runtime confs
         driver = self.get_local_driver_binaryname()
         runtimeconf = self.get_local_runtimeconf_binaryname()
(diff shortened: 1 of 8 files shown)


sagark:
I think we can ship this with the 1.4.0 update since that will break AGFI compatibility anyway?

davidbiancolin:
Yeah sounds good.

PR #62, opened 2018-08-08
[RFC] Makefile rework & partial introduction of MIDAS examples

davidbiancolin:
This introduces the [midas-examples](https://github.com/ucb-bar/midas-examples) as subpackage of firesim, leaving out RISC-V mini because we have Rocket-Chip. To make them co-exist with the existing RC-based firesim targets i reworked the Makefiles. I'm hoping this should enable two use cases:

1) Multiple targets co-existing within Firesim with Firesim acting as the "top" in the project.
As before, the user interacts with the build system through sim/Makefile, but based on the desired target-package will `include`:
- `sim/src/main/makefrag/<target-package>/Makefrag` -- a target-specific Makefrag
  - which defines all the target-specific sources and defines a recipes for generating the simulator verilog and memory map
- `sim/Makefrag` -- a target-agnostic Makefrag
  - this defines recipes for building MIDAS-level simulators and drivers, which are really just wrappers around a makefile provided in MIDAS
  - xsim recipes,  vestigial fpga build recipes, and `replace_rtl` are defined here

2) A chip-top project that includes FireSim as a submodule. Here chip-top will have a Makefile that subsumes `sim/Makefile` and `sim/src/main/makefrag/<target-package>/Makefrag` and includes `sim/Makefrag` directly

As part of introducing midas-examples, this change also reworks the sbt-based tests. Here the goal is to make an SBT TestSuite provide a thin-wrapper around the makefiles -- the only thing not proxied out to a makefile is RTL generation (which would invoke another instance of SBT). 

Looking for comments. @zhemao @sagark @colinschmidt 

Some open questions:

1. Manager interactions? 

TODO:
- [x] reintegrate sbt-tests from midas-examples
- [x] fix PointerChaser -- need memory init
- [x] fix Risc & RiscSRAM 
- [x] fix runtime conf generator
- [x] check xsim recipes
- [ ] give GroundTest to its own target package?


davidbiancolin:
@sagark @zhemao This is almost ready to merge and is worth having a look at as it's a large change.

davidbiancolin:
@sagark  @zhemao This is ready for merge

sagark:
lgtm overall

- does this affect the manager at all?
- can you add docs for the new stuff?

davidbiancolin:
No manager changes -- all of the variables touched by the manager are the same. I can start writing docs if we are happy with this.

PR #66, opened 2018-08-13
Editing scan over Generating-Different-Targets

davidbiancolin:
BOOM can generate superscalar pipelines.


sagark:
I'm going to add the fix for the SSH-ing instructions and PR this straight to master since it's just docs.

PR #72, opened 2018-08-16
FireSim 1.3.1 Release - BOOM support, host network fixes

### LINKED ISSUE #12: [BOOM] Add support for BOOM
BOOM Support has been added to FireSim in the 1.2 release. However, there are still target-level bugs that need to be resolved in BOOM before it can be used fully for Linux-based software (Linux currently boots most of the way, but then hangs, which can be reproduced in both regular VCS (no sim components) and on the FPGA). See the log in the next post.

### LINKED ISSUE #23: Fix subnetting for 256+ node networked simulations
In 256+ node simulations, we should be using /16 instead of /24 addresses.

This was fixed in scripts during ISCA, but didn't make it into the scripts here.

### LINKED ISSUE #68: Fix IP address conflict between simulations and hosts
This is only a bug for users trying to SSH into simulated nodes. Doesn't matter for anyone else.

If a user used the aws-setup.py script to setup their account (which they should be doing), host instances get IPs from 192.168.0.0/24 and in all of our scripts, we default simulated nodes to also getting 192.168.0.0/24 addresses. This is a problem when bridging together the VPC network and the simulated network (e.g. when enabling ssh into simulated nodes).

This was unnoticed because my account was not set-up with our aws-setup.py because it didn't exist when I made my account :)

I think the fix is as explained here, but we will also need to propagate the IP address changes to all of our experiment scripts. I haven't had a chance to look, but IIRC the manager does not need to change.

https://groups.google.com/d/msg/firesim/zBu-a2A3xHM/jCaHTkd9AgAJ

sagark:
Ready for review/merge

# FireSim 1.3.1 Release

## Updates:

- Bumps to a stable single-core BOOM, boots Linux and runs SPEC. Resolves #12 
- Fix instructions for accessing the internet from within simulations. eth0 is not a reliable interface name on instances.
- Change IP address range assigned to simulations to prevent conflict with host instance IPs (both were 192.168.0.0 ranges before if you used aws-setup.py to setup aws infrastructure). Resolves #68 
- While we were changing addressing anyway, resolved #23 

## Testing
- Re-ran memcached-thread-imbalance to make sure IP addressing fixes didn't break stuff
- SSH'd into a BOOM and wget-ed google.com from within it

## TODOs

- [x] merge boom agfis PR into here
- [x] Generate new BOOM agfis
- [x] Confirm memcached test output
- [x] Merge submodules to master
- [x] Make re-generated AGFIs public
- [x] Mark which issues it resolves

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 This changelog follows the format defined here: https://keepachangelog.com/en/1.0.0/
 
+## [1.3.1] - 2018-08-18
+
+### Changed
+
+* Update version of BOOM included as a target in FireSim. The included version/configurations now reliably boot Linux and run SPEC. The FireSim NIC/network is also supported.
+* Update IP addressing for simulated nodes to prevent a conflict between host IP range and simulation IP range. Simulations now live in 172.16.0.0/16 instead of 192.168.0.0/24.
+* Update experimental SSH into target instructions to account for different interface names across EC2 instance types. This fixes the ability to access the internet from within simulated nodes.
+
+
 ## [1.3.0] - 2018-08-11
 
 ### Changed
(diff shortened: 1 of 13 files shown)


sagark:
@davidbiancolin this is ready for review/merge

Issue #74, opened 2018-08-17
Remove block device driver messages from dmesg in Linux


sagark:
There are too many.

sagark:
URGENT @zhemao @zhemao

PR #77, opened 2018-08-25
Some improvements from memblade branch

zhemao:
I'd like to merge in a few changes I made when working on the memblade paper that will be useful in general.

* Optimized NIC with split TX/RX IRQs and checksum offloading
* Set DRAM profiling interval in configuration
* Throttle outbound bandwidth from switch
* Allow packets to be split across batches in switch

### DIFF
--- a/target-design/switch/switch.cc
+++ b/target-design/switch/switch.cc
@@ -34,11 +34,12 @@ int switchlat = 0;
 
 #define SWITCHLATENCY (switchlat)
 
-// param: max number of flits that can be sent per batch
+// param: numerator and denominator of bandwidth throttle
 // Used to throttle outbound bandwidth from port
 //
-// THIS IS SET BY A COMMAND LINE ARGUMENT. DO NOT CHANGE IT HERE.
-int maxflits = 0;
+// THESE ARE SET BY A COMMAND LINE ARGUMENT. DO NOT CHANGE IT HERE.
+int throttle_numer = 1;
+int throttle_denom = 1;
 
 // uncomment to use a limited output buffer size, OUTPUT_BUF_SIZE
 //#define LIMITED_BUFSIZE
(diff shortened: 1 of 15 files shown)


target-design/switch/switch.cc
```diff
@@ -11,7 +11,7 @@
 #include <omp.h>
 #include <cstdlib>
 
-#define IGNORE_PRINTF
+//#define IGNORE_PRINTF
```
sagark: why is this commented out?

target-design/switch/switch.cc
```diff
@@ -156,8 +163,8 @@ while (!pqueue.empty()) {
     switchpacket * tsp = pqueue.top().switchpack;
     pqueue.pop();
     uint16_t send_to_port = get_port_from_flit(tsp->dat[0], 0 /* junk remove arg */);
-    printf("packet for port: %x\n", send_to_port);
-    printf("packet timestamp: %ld\n", tsp->timestamp);
+    //printf("packet for port: %x\n", send_to_port);
+    //printf("packet timestamp: %ld\n", tsp->timestamp);
```
sagark: why are these commented out?

sagark:
how well tested is the new NIC code? have you re-run memcached/iperf/etc?

what's the perf improvement of the new NIC stuff?

sagark:
Also while we're at it, can you take care of #74

target-design/switch/switch.cc
zhemao: Whoops. I thought I reverted those changes, but apparently not.

zhemao:
I've run iperf3 with the new NIC code and it works pretty consistently.

I got up to 1.9 Gbps with the code at one point, but since updating rocket-chip I'm now getting just 1.6 Gbps.

zhemao:
Fixed issue #74 and reverted the switch.cc printf changes.

sagark:
@zhemao can you run the full memcached experiment on it and send me the graph?

sagark:
@davidbiancolin does this conflict with your changes at all? if so we should merge your PR first since it's bigger.

davidbiancolin:
The only conflict is firesim_top.cc is moved to firesim/firesim_top.cc i believe. There's nothing serious.

PR #78, opened 2018-08-25
[Manager] Generate a better initial bucket name

davidbiancolin:
Users.filterNot(_ == "sagark") won't have to modify the bucket name after manager init.


sagark:
heh. lgtm

PR #90, opened 2018-09-05
Fix sentence to be three instead of two

abejgonzalez:


sagark:
When you're done working through the tutorial and finding stuff lmk and I'll merge

PR #98, opened 2018-09-13
Re-work main loop, encapsulate FESVR entirely in the Serial endpoint

davidbiancolin:
This is a fairly big change, it can wait until the tracer stuff and other things are pushed into dev.

Overview of changes:
- Reworks endpoint interface
  - Rename done -> terminate. Assert true if the endpoint wants to bring down simulation
  - Add exitcode() -> encodes reason for termination. 0 = pass
- Reworks main loop 
  - add's systematic_scheduler_t, which advances the simulator to earliest cycle possible (removes the need for an explicit stepsize)
  - tasks (task_t (std::function<uint64_t()> that must be run at specific target-cycles can be registered in main class's constructor or init. 
  - tasks return value indicates the number of target-cycles that should be advanced before they should run again
- Moves all FESVR-related code into the serial_t endpoint class. 
- Adds an FMR print that uses new hardware cycle counters

Todos: 
- [x] FPGA-hosted tests (What should be run?)
- [x] Test networked simulation performance

### DIFF
--- a/deploy/sample-backup-configs/sample_config_hwdb.ini
+++ b/deploy/sample-backup-configs/sample_config_hwdb.ini
@@ -9,28 +9,27 @@
 # If you are using an older version of FireSim, you will need to generate your
 # own images.
 
-[firesim-quadcore-nic-ddr3-llc4mb]
-agfi=agfi-0363594a482851ec5
-deploytripletoverride=None
-customruntimeconfig=None
-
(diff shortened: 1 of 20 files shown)


sagark:
looks great to me so far!

davidbiancolin:
@zhemao You should probably look at this too. Profile-interval can now be set to arbitrary non-zero values.

davidbiancolin:
This is ready for review and merge.

PR #106, opened 2018-09-20
Deterministic, latency-pipe block-device model

davidbiancolin:
This PR provides a deterministic fixed-latency model for the block device (with independently programmable read and write latency). 

This should also save a bunch of MMIO when the block device is idle.

Adds target-land blockdev tests to the generated makefrag, and the ScalaTests. We should consider adding more. 

Todo: 
- [x] Make latencies CML args. Legalize against hardware widths.
- [x] Debug features: Add counters for stall causes / bind stall causes to MMIO regs 
- [x] Pick the right base branch. This maybe should go into the main loop rework. 
- [x] Add block device tests to makefrag & scala tests 
- [ ] Consider increasing tracker count? Or at least testing that.

### DIFF
--- a/deploy/sample-backup-configs/sample_config_hwdb.ini
+++ b/deploy/sample-backup-configs/sample_config_hwdb.ini
@@ -10,26 +10,27 @@
 # own images.
 
 [fireboom-singlecore-nic-ddr3-llc4mb]
-agfi=agfi-0c62c26523d4c1a4d
+agfi=agfi-0c050337424731b7d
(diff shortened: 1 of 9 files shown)


davidbiancolin:
This is ready for review. @zhemao you should look at this too.

davidbiancolin:
If we can automate NIC tests, we can add in the NIC targets to the ScalaTests.

sagark:
Have we run any disk-heavy benchmarks on linux?

PR #110, opened 2018-09-21
verilator install

sagark:
Pre-install verilator on manager instances.

Tested on a fresh instance.


sagark:
i used this to run ASM tests for the bump

PR #113, opened 2018-09-23
Bump Rocket Chip to July 3 2018

sagark:
Ready for review but not merge. The last thing left here is merging midas PR to dev and doing a final bump of the submodule.

TODOs:

- BOOM is not yet bumped. @jerryz123 
- All submodules now point to dev, excluding midas which has an outstanding PR.

Notes:
- We no longer submodule a custom sifive-blocks. Instead, set UART baud rate from software in riscv-pk
- UART queue sizes are now increased to 256 instead of 8. Under normal typing, this means characters you type won't be lost. If you paste something > 256, chars may still be lost. Per discussion with @a0u, fixing the UART widget software to use a pty should handle this correctly with SW flow control.
- Software has not been bumped to upstream. But everything still works with existing software, so I think we should do SW in a separate bump.

Testing:
- re-ran memcached, simperf-latency, ping-latency, they all look fine

### DIFF
--- a/target-design/TODO
+++ b/target-design/TODO
@@ -1,0 +1,3 @@
+-have not bumped riscv-tools
+-have not fixed boom
+-probably want to bump linux too
--- a/target-design/TODO
+++ b/target-design/TODO
@@ -1,3 +1,4 @@
 -have not bumped riscv-tools
 -have not fixed boom
 -probably want to bump linux too
+-barstools/mdf are not bumped to master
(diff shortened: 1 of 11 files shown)


target-design/TODO
```diff
@@ -0,0 +1,4 @@
+-have not bumped riscv-tools
+-have not fixed boom
+-probably want to bump linux too
+-barstools/mdf are not bumped to master
```
davidbiancolin: Is there a reason to change the barstools pointer unless there are breakages under new firrtl/chisel?

PR #114, opened 2018-09-23
FireSim Release 1.4.0

### LINKED ISSUE #6: [UART] Use a pty instead of taking over stdin/stdout
This only requires modifying the UART simulation driver (C++).

See @a0u's example: https://gist.github.com/a0u/2d533ca4ea91764a0fa8

### LINKED ISSUE #60: Add deterministic timing model to block device.
This is perhaps the largest remaining source of non-determinism in our sims.

### LINKED ISSUE #70: Fix how sourceme scripts deal with ssh-agent when user is using ssh forwarding

### LINKED ISSUE #74: Remove block device driver messages from dmesg in Linux

### LINKED ISSUE #105: Give target-land device tests to their own recipes in the generated makefrag
We'd like to be easily able to test things like the blockdev, uart, even NIC in midas level simulation, without having to go through a bunch of manual steps.

sagark:
FireSim Release 1.4.0
------------------------

## Rocket Chip Bump 1 of 2
- Rocket Chip is bumped to July 3, 2018
- We no longer submodule a custom sifive-blocks. Instead, set UART baud rate from software in riscv-pk
- UART queue sizes are now increased to 256 instead of 8. Under normal typing, this means characters you type won't be lost. If you paste something > 256, chars may still be lost. Per discussion with @a0u, fixing the UART widget software to use a pty should handle this correctly with SW flow control.
- Software has not been bumped to upstream. But everything still works with existing software, so I think we should do SW in a separate bump.
- As discussed in the meeting, pre-built AGFIs are only provided for 3 rocket-chip configs and 2 boom configs.
- BOOM is bumped.
- All submodules now point to dev

Testing:
- re-ran memcached, simperf-latency, ping-latency, they all look fine

## Rework main loop #98 

- Reworks endpoint interface
  - Rename done -> terminate. Assert true if the endpoint wants to bring down simulation
  - Add exitcode() -> encodes reason for termination. 0 = pass
- Reworks main loop 
  - add's systematic_scheduler_t, which advances the simulator to earliest cycle possible (removes the need for an explicit stepsize)
  - tasks (task_t (std::function<uint64_t()> that must be run at specific target-cycles can be registered in main class's constructor or init. 
  - tasks return value indicates the number of target-cycles that should be advanced before they should run again
- Moves all FESVR-related code into the serial_t endpoint class. 
- Adds an FMR print that uses new hardware cycle counters

## Rocket Chip Bump 2 of 2
- Rocket Chip bumped to master as of Sept 26 2018
- BOOM boots linux and can ping between nodes in a cluster @jerryz123 
- Re-ran ISCA experiments excluding simperf-scale. Results as expected.
- AGFIs public

Submodule status:
- firechip dev branch
  - very close to RC master (yesterday)
  - sifive-blocks master
  - boom master
- firrtl master
- midas dev

Comments to be addressed:
- HasSystemErrorSlave removed from RC. Do we care? Commented out for now.
- re: gitignore: where do the black box resources get dumped? Maybe exclude the whole dir?

## Deterministic Block Dev #106 
- provides a deterministic latency pipe timing model for the blockdev
  - independently programmable read and write latencies (defaults to 4096 cycles)
- adds blockdev tests to generated makefrag and ScalaTests
- resolves #60, resolves #105 

## Build Instance Request Retries #126
- Unify code between `launch_instances()` and `launch_run_instances()` to re-attempt failed spot requests when the availability zone of the initially selected subnet is oversubscribed.

## ssh-agent Forwarding #128
- If running within a `tmux` or `screen` session, ensure that a local `ssh-agent` instance is always used and avoid agent forwarding.
- Resolves #70.

## Model Endpoint CDCs & MIDAS unit tests #49  (ucb-bar/midas#90)
- Adds Platform Configs for building simulators with clock division between RC and DRAM models
- Add recipes to run midas synthesizable unit tests
- Adds unittests to scala tests
- Adds an submodule init only mode to build setup 

## Firesim Software reorg, better scripting
- Unify configurations of Linux images; the same exact image boots in both qemu and firesim. You can now configure/test your images in qemu before launching in firesim.
- Automate launching of configurations in qemu or spike
- Automated system for building working Fedora images

## Manager network mapping improvements, faster fpga clear/flash
- New "FireSimLink" node type used internally in manager to better track information about links. Doesn't affect users.
  - Manager now creates globally unique names for all shared-mem links, these are then put into switchconfigs / passed as a plus arg to fpga-sims. No longer hardcode naming based on slot, etc.
  - These links are automatically materialized into some type of host-level transport based on how the endpoints of the link are mapped to the host (see custom mapping below). 
  - As a result of these changes, you can basically have arbitrary placement of switches on host nodes now
- Custom user topologies can now include a custom mapping function to let them manually map simulated nodes/switches to host nodes. The API is documented in the example custom mapping given in the fat_tree_4ary topology. Topologies that do not supply a custom mapping will use the same old default.
- Speed-up FPGA clear and flashing (use async args to AWS tools). Unlike previous attempts at doing this, it actually seems to work now. Was stress tested in our ARPA-E experiments.
- Switch model improvements
  - Simple "token compression": If a batch is completely empty on a link with socket transport, convey this with only a 64bit flag. This doesn't really help for small sims, but allows large sims where there are many links that cross ec2 instances, where the links are not heavily utilized (e.g. IceNet). Otherwise, you become BW limited instead of latency limited.
  - Multiple uplink support in switches
  - Randomized switching across multiple uplink paths
    - With broadcast support (avoid loops in simple clos cases)
- Pull in various updates on master

-Resolves #74, removed blkdev linux driver printks

## Assertion Synthesis #143
- Assertions can now be synthesized and caught during FPGA-hosted simulation
- Enable by mixing in `WithSynthAsserts` into your PLATFORM_CONFIG
- See Debugging/Assertion-Synthesis
- Introduces a few Config fixes
  - Adds a Config that copies an existing RocketTiles or BoomTiles Field N ways using `up`
  - L2 TLB & HPM counters now correctly generated in multi-core targets 

## Remove macros

Replaced macros in MIDAS with struct generation. Easier to have multiple copies of a particular endpoint now.

## WIP Supernode Merge to Master
- This adds infrastructure to support supernode configurations on firesim master (rather than having a separate branch). The support is not complete, this is just supporting infrastructure.
- Several endpoints now support having multiple copies (also see remove macros above)
- UARTs with id > 0 now use a PTY rather than writing to file. Resolves #6 

## IceNIC improvements

- NIC counts MMIO registered changed to make each count 8 bits instead of 4. This expands the maximum size of the req/resp queues from 16 to 256.
- TX/RX completion interrupts separated into two different interrupts.
- Added interrupt masking

## MISC
- Support multiple widgets with DMA. DMA addresses hard-coded for now.
- Add TracerV Widget. This allows getting the trace coming from the simulated Rocket Chip's trace port into a log file on the host. Can work alongside the NIC due to multiple DMA support.
- Support multiple endpoints using DMA (e.g. SimpleNICWidget + TracerVWidget). This currently requires hardcoding addresses. The next release will support automatically allocating addresses as is already done with MMIO.

## Deferred
- Tracer V docs @sagark 

TODO:
- [x] Wait for CHANGELOG PR merge
- [x] Wait for submodule merges to dev
- [x] wait for final isca workload runs (not running scale)


sagark:
@davidbiancolin @zhemao @alonamid 

this is ready for review. can we get approvals from everyone?

davidbiancolin:
Net +21 lines!

PR #120, opened 2018-09-27
[DO NOT MERGE] Rocket Chip Bump to Sept. 2018

sagark:
running isca paper tests now

- BOOM boots linux and can ping between nodes in a cluster @jerryz123 
- AGFIs public

### DIFF
--- a/sim/.gitignore
+++ b/sim/.gitignore
@@ -4,3 +4,5 @@ output
 test.disk
 tags
 *.timestamp
+AsyncResetReg.v
+firrtl_black_box_resource_files.f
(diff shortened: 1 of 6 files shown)


sagark:
@davidbiancolin this is ready for review/merge

should be fairly painless. I didn't even have to blow away ~/.sbt ~/.ivy2

sim/src/main/scala/firesim/Targets.scala
```diff
@@ -25,7 +25,7 @@ import java.io.File
 class FireSim(implicit p: Parameters) extends RocketSubsystem
     with CanHaveMisalignedMasterAXI4MemPort
     with HasPeripheryBootROM
-    with HasSystemErrorSlave
+//    with HasSystemErrorSlave
```
davidbiancolin: What is this about?

sim/.gitignore
```diff
@@ -4,3 +4,5 @@ output
 test.disk
 tags
 *.timestamp
+AsyncResetReg.v
```
davidbiancolin: where do the black box resources get dumped? Maybe exclude the whole dir?

PR #122, opened 2018-09-29
Add TraceRV alongside NIC

zhemao:


sim/Makefrag
```diff
@@ -67,6 +68,12 @@ $(verilator_debug): $(HEADER) $(DRIVER_CC) $(DRIVER_H) $(midas_cc) $(midas_h)
 verilator: $(verilator)
 verilator-debug: $(verilator_debug)
 
+run-verilator: $(verilator)
```
sagark: How does this differ from the existing run commands?

sagark: Also if we want to keep it, it should be added for vcs too

sim/build.sbt
```diff
@@ -21,7 +21,8 @@ lazy val testchipip = project in file("target-rtl/firechip/testchipip") settings
 lazy val icenet     = project in file("target-rtl/firechip/icenet") settings commonSettings dependsOn (rocketchip, testchipip)
 
 lazy val mdf        = RootProject(file("barstools/mdf/scalalib"))
-lazy val barstools  = project in file("barstools/macros") settings commonSettings dependsOn (mdf, rocketchip)
+lazy val firrtl     = project in file("firrtl") settings commonSettings
```
sagark: we should be using the firrtl installed with publishLocal right?

sim/project/plugins.sbt
```diff
@@ -17,3 +17,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-pack" % "0.9.3")
 
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
+
+addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.6.3")
+
+libraryDependencies += "com.github.os72" % "protoc-jar" % "3.5.1.1"
```
sagark: what are these for?

sim/Makefrag
zhemao: This allows you to run arbitrary binaries, not just the ones that can be symlinked to output/. I was just basing it off the run-xsim rule.

sim/build.sbt
zhemao: That causes some issues when firrtl updates. You end up having to redo the publishLocal. But I guess we have to do it for rocket-chip anyway, so I'll get rid of it.

sim/Makefrag
sagark: Okay, that sounds useful. Can you add it for vcs also? 

@davidbiancolin

sim/project/plugins.sbt
zhemao: It's necessary because I added firrtl to the list of dependencies. If we just use the publishLocal firrtl instead, we don't need it.

sim/build.sbt
sagark: 👍

sim/project/plugins.sbt
sagark: Yeah let's keep things as is.

zhemao:
I'll comment it out then.

sim/project/plugins.sbt
sagark: If there's a good reason to do this, we can move it to a separate PR.

sagark:
It lowers perf right? There's no runtime pause that doesn't hurt perf?

zhemao:
Simulation performance or target performance?

sim/Makefrag
zhemao: I don't have a firesim installation on the millennium mahcines set up for testing, can you or David test the commit I just pushed?

sim/Makefrag
```diff
@@ -41,7 +41,8 @@ CONF_NAME ?= runtime.conf
 TIMEOUT_CYCLES = 100000000
 SIM_RUNTIME_CONF ?= $(GENERATED_DIR)/$(CONF_NAME)
 mem_model_args = $(shell cat $(SIM_RUNTIME_CONF))
-COMMON_SIM_ARGS ?= $(mem_model_args)
+nic_args = +niclog=niclog +linklatency=6405 +netbw=100 +netburst=8
```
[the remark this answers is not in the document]
zhemao: ? But this is firesim's Makefrag. Is there a different Makefrag this should go in?

sagark:
LGTM now pending:
- add docs for TraceRV

PR #123, opened 2018-09-29
Fix path to config files

pranith:


sagark:
Thanks for the PR, however the paths of these files have actually changed as of the last release. If you're using an older release of FireSim, you should be able to switch to the documentation for that previous release on readthedocs.

PR #126, opened 2018-10-04
Attempt all possible subnets when launching build instances

a0u:
The original implementation of `launch_instances()` lacked a retry mechanism to tolerate a failed spot request when the availability zone of the randomly selected subnet is oversubscribed.  This generalizes
`launch_run_instances()` to eliminate code duplication and impose consistent behavior.


deploy/awstools/awstools.py
```diff
@@ -116,12 +72,16 @@ def launch_run_instances(instancetype, count, fsimclustertag, instancemarket, sp
     vpcfilter = [{'Name':'tag:Name', 'Values': [vpcname]}]
     firesimvpc = list(ec2.vpcs.filter(Filters=vpcfilter))
     subnets = list(firesimvpc[0].subnets.filter())
+    random.shuffle(subnets)
```
sagark: Can you remove the randomization?

sagark:
LGTM besides the randomization comment

deploy/awstools/awstools.py
a0u: If randomization is problematic for run instances, then we can add a `randomize=false` optional argument.  Otherwise, it seems somewhat useful to not have all spot requests for build instances skew towards earlier subnets, even if they initially succeed.

sagark: Yeah, let's do that.

PR #132, opened 2018-10-11
NIC bugfixes and improvements

zhemao:
This is a more limited version of the big memblade backport. I've left out the splitting of TX/RX interrupts and checksum offload and only added the expansion of NIC counts to 8 bits, which allows the command queues to expand to up to 256 entries. I've updated the Linux driver and network bandwidth test to reflect this change.

### DIFF
--- a/target-design/switch/switch.cc
+++ b/target-design/switch/switch.cc
@@ -34,6 +34,13 @@ int switchlat = 0;
 
 #define SWITCHLATENCY (switchlat)
 
+// param: numerator and denominator of bandwidth throttle
+// Used to throttle outbound bandwidth from port
+//
+// THESE ARE SET BY A COMMAND LINE ARGUMENT. DO NOT CHANGE IT HERE.
+int throttle_numer = 1;
+int throttle_denom = 1;
+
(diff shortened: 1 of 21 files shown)


target-design/switch/switch.cc
```diff
@@ -34,6 +34,13 @@ int switchlat = 0;
 
 #define SWITCHLATENCY (switchlat)
 
+// param: numerator and denominator of bandwidth throttle
+// Used to throttle outbound bandwidth from port
+//
+// THESE ARE SET BY A COMMAND LINE ARGUMENT. DO NOT CHANGE IT HERE.
+int throttle_numer = 1;
+int throttle_denom = 1;
```
sagark: are these actually set by a command line argument? I don't see it?

sagark:
Can you merge/propagate the merge commits of submodules before you merge this? @zhemao

target-design/switch/switch.cc
zhemao: Line 228. Their pointers are passed to simplify_frac, which is what sets them.

sagark: ah got it. cool

PR #134, opened 2018-10-18
Enable IceNet driver in Fedora kernel config

a0u:
Pending confirmation that the network interface is brought up correctly in Fedora.


sagark:
@NathanTP might be useful to you also

PR #137, opened 2018-10-24
Manager network mapping improvements

sagark:
This pulls in all of the improvements we made to the manager for supporting clos topologies. Most users will primarily be interested in the improved FPGA clear/flashing times.

New stuff:
- New "FireSimLink" node type used internally in manager to better track information about links. Doesn't affect users.
  - Manager now creates globally unique names for all shared-mem links, these are then put into switchconfigs / passed as a plus arg to fpga-sims. No longer hardcode naming based on slot, etc.
  - These links are automatically materialized into some type of host-level transport based on how the endpoints of the link are mapped to the host (see custom mapping below). 
  - As a result of these changes, you can basically have arbitrary placement of switches on host nodes now
- Custom user topologies can now include a custom mapping function to let them manually map simulated nodes/switches to host nodes. The API is documented in the example custom mapping given in the fat_tree_4ary topology. Topologies that do not supply a custom mapping will use the same old default.
- Speed-up FPGA clear and flashing (use async args to AWS tools). Unlike previous attempts at doing this, it actually seems to work now. Was stress tested in our ARPA-E experiments.
- Switch model improvements
  - Simple "token compression": If a batch is completely empty on a link with socket transport, convey this with only a 64bit flag. This doesn't really help for small sims, but allows large sims where there are many links that cross ec2 instances, where the links are not heavily utilized (e.g. IceNet). Otherwise, you become BW limited instead of latency limited.
  - Multiple uplink support in switches
  - Randomized switching across multiple uplink paths
    - With broadcast support (avoid loops in simple clos cases)
- Pull in various updates on master


sagark:
This is now ready for review/merge @davidbiancolin @zhemao

deploy/runtools/run_farm.py
```diff
@@ -24,13 +24,14 @@ def __init__(self):
 class EC2Inst(object):
     # for now, we require that only one switch is mapped to each host
     # this could be overridden based on instance type later
-    SWITCH_SLOTS = 1
+    SWITCH_SLOTS = 100000
```
davidbiancolin: This comment should be updated and this number explained.

PR #138, opened 2018-10-25
firesim-software overhaul

NathanTP:
This branch bumps firesim to the refactored version of firesim-software that includes more comprehensive support for fedora, buildroot, qemu, and spike. There were minimal changes to buildroot, just a slight change in serial port stuff to get it to play nice with qemu.

I've tested:
* The disk based versions of buildroot and fedora on single-node firesim and qemu
* The disk-based version of buildroot on multi-node firesim
* the initramfs versions of both on spike and qemu

note 1: Spike won't work with 'build-setup.sh fast' until another pull request goes through. The prebuilt riscv-tools isn't a submodule, so this will get fixed automatically after it's updated.

note 2: This is pointing at a branch of firesim-software instead of master. If you'd rather I merge firesim-software first and have the submodule point there, that's fine to.

note 3: Once this goes in, I'll update the firesim website


NathanTP:
Symlinks and build scripts in workloads/deploy all compile now (except checksum which doesn't compile on dev either). I ran ping-latency and it worked fine, I assume all the others still work (I didn't actually change any configs, just symlinks).

sagark:
I also ran memcached-thread-imbalance and it worked

davidbiancolin:
It'd be chill if a recipe for whatever linux uniform becomes could be added to the workload recipes, given that building bbl and the disk image is a little more complicated than `./build.sh` now, and would be more robust against future sw-manager changes.

PR #142, opened 2018-10-30
Update docs to reflect changes to firesim-software

NathanTP:
Also renames linux-uniform to br-uniform since fedora is now a first-class citizen.


sagark:
I would prefer to leave the linux-uniform naming as is, rather than renaming to br-uniform. What do you think @davidbiancolin

docs/Running-Simulations-Tutorial/Running-a-Single-Node-Simulation.rst
```diff
@@ -108,9 +129,10 @@ have a NIC. This hardware configuration models a Quad-core Rocket Chip with 4
 MB of L2 cache and 16 GB of DDR3, and **no** network interface card.
 
 We will leave the last section (``[workload]``) unchanged here, since we do
-want to run Linux on our simulated system. The ``terminateoncompletion``
-feature is an advanced feature that you can learn more about in the
-:ref:`manager-configuration-files` section.
+want to run the buildroot-based Linux on our simulated system (you could also
+change this to fedora-uniform.json if you'd like to boot fedora). The
```
sagark: I don't want to point to fedora here because it's good to have the intro tutorials be as one-track as possible.

docs/Running-Simulations-Tutorial/Running-a-Single-Node-Simulation.rst
```diff
@@ -354,8 +376,9 @@ You can ignore the messages about the network -- that is expected because we
 are simulating a design without a NIC.
 
 Now, you can login to the system! The username is ``root`` and the password is
-``firesim``. At this point, you should be presented with a regular console,
-where you can type commands into the simulation and run programs. For example:
+``firesim`` (the password on fedora is ``riscv``). At this point, you should be
```
sagark: Same as above, I don't think we should point to fedora here.

docs/Advanced-Usage/Workloads/Booting-Fedora.rst
```diff
@@ -1,18 +0,0 @@
-Running Fedora on FireSim
```
sagark: Let's keep a separate page about booting Fedora

davidbiancolin:
It'd be chill if a recipe for whatever linux uniform becomes could be added to the workload makefile, given that building bbl and the disk image is a little more complicated than `./build.sh` now, and would be more robust against future sw-manager changes.

NathanTP:
I reverted the linux-uniform name change and updated the documentation with more modest changes. Let me know if that seems clear enough.

PR #143, opened 2018-11-01
Assertion Synthesis

davidbiancolin:
This only introduces assertion synthesis support as there are some hacks i'd like to clean up in the print stuff before bringing it online (though most of the code is present). 

This also fixes bugs in our use of Config, by adding Configs that make it less error prone to set a tile count without blowing away existing modifications to TileParams. 

See ucb-bar/midas#95. 

TODO:
- [x] Deliver print and assertion definition files to run farm instances. Or bake into the driver?
  - Deliver asserts until midas c-struct generation stuff is online 

RFC:
- Should we enable assertion synthesis by default?
  - Disabled

Testing:
- [x] Add midas examples for Assertions
- [x] Add midas examples for Prints
  - Push to later commit
- [x] Add dummy assertions in Rocket, catch in ML sim
- [x] Add dummy assertions in Rocket, catch in FPGA sim
  - Caught a real assertion in BOOM


sim/src/main/scala/firesim/TargetConfigs.scala
```diff
@@ -30,16 +30,10 @@ class WithNICKey extends Config((site, here, up) => {
   case NICKey => NICConfig(inBufPackets = 10)
 })
 
-class WithLargeTLBs extends Config((site, here, up) => {
+class WithRocketL2TLBs(entries: Int) extends Config((site, here, up) => {
   case RocketTilesKey => up(RocketTilesKey) map (tile => tile.copy(
-    icache = tile.icache map (_.copy(
-      nTLBEntries = 32 // TLB reach = 32 * 4KB = 128KB
```
davidbiancolin: These were the defaults in Rocket-Chip anyways.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -254,6 +254,7 @@ def get_required_files_local_paths(self):
 
         all_paths.append(self.server_hardware_config.get_local_driver_path())
         all_paths.append(self.server_hardware_config.get_local_runtime_conf_path())
+        all_paths.append(self.server_hardware_config.get_local_assert_def_path())
```
sagark: does this work properly if assertions aren't on?

sagark: nvm, saw the midas PR

PR #144, opened 2018-11-04
Removing Macros + Some endpoints support multiple instantiations of themselves

sagark:
- Replace macros with structs (https://github.com/ucb-bar/midas/pull/96)
    - Adds a macro used in firesim_top.cc to create a struct
    - Struct then gets passed into the constructor, provides access to mmio reg addrs
    - endpoint destructor responsible for free()ing struct
- Added TODO list for some missing things that will be in a separate PR
- Start replacing arg parsing for certain endpoints to support multiple instantiations of that endpoint (tracer, NIC, blkdev) in prep for supernode
- Some endpoints need more work to support multiple instantiations (Serial (arg parsing), UART (pty support))

### DIFF
--- a/sim/src/main/cc/endpoints/blockdev.cc
+++ b/sim/src/main/cc/endpoints/blockdev.cc
@@ -18,8 +18,9 @@
  * Setup software driver state:
  * Check if we have been given a file to use as a disk, record size and
  * number of sectors to pass to widget */
-blockdev_t::blockdev_t(simif_t* sim, const std::vector<std::string>& args): endpoint_t(sim) {
-    _ntags = BLOCKDEVWIDGET_0(num_trackers);
+blockdev_t::blockdev_t(simif_t* sim, const std::vector<std::string>& args, uint32_t num_trackers, uint32_t latency_bits, BLOCKDEVWIDGET_struct * mmio_addrs): endpoint_t(sim) {
+    this->mmio_addrs = mmio_addrs;
+    _ntags = num_trackers;
     long size;
     long mem_filesize = 0;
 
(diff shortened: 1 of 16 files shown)


sim/src/main/cc/endpoints/blockdev.cc
```diff
@@ -92,13 +94,13 @@ blockdev_t::~blockdev_t() {
  * Here, we set control regs e.g. for # sectors, allowed request length
  * at boot */
 void blockdev_t::init() {
-#ifdef BLOCKDEVWIDGET_0
+#ifdef BLOCKDEVWIDGET_struct_guard
```
davidbiancolin: Can't we just remove all this and just not instantiate the endpoint now?

davidbiancolin: By all this, i mean ifdefs around function bodies.

sagark:
@davidbiancolin this is ready for review now. This doesn't solve all problems (namely integrating other constants the widget chisel writer might want to emit), but I'd like to get it merged in since it's a lot of stuff that easily causes merge conflicts.

sim/src/main/cc/endpoints/simplenic.cc
```diff
@@ -79,34 +89,34 @@ simplenic_t::simplenic_t(
...
                 }
             } else {
-                fprintf(stderr, "INVALID MAC ADDRESS SUPPLIED WITH +macaddr=\n");
+                fprintf(stderr, "INVALID MAC ADDRESS SUPPLIED WITH +macaddrN=\n");
             }
         }
-        if (arg.find("+netbw=") == 0) {
-            char *str = const_cast<char*>(arg.c_str()) + 7;
+        if (arg.find(netbw_arg) == 0) {
+            char *str = const_cast<char*>(arg.c_str()) + netbw_arg.length();
```
davidbiancolin: Tight. That old pattern had to die.

sagark:
yeah, that's right

PR #146, opened 2018-11-07
UART prep for supernode on master

sagark:
- UARTs now use PTY if they are not the main UART. main UART (UART0) still uses stdout/stdin so existing non-supernode users are not affected
    - As a result, supernode users no longer have to pipe into a file to interact with console on simulated nodes.
    - To prevent having to launch a ton more screens for now, UARTs != UART0 will write their own uartlog (uartlog1, uartlog2, etc.)

### DIFF
--- a/sim/src/main/cc/endpoints/uart.cc
+++ b/sim/src/main/cc/endpoints/uart.cc
@@ -36,6 +36,7 @@ void sighand(int s) {
 uart_t::uart_t(simif_t* sim, UARTWIDGET_struct * mmio_addrs, int uartno): endpoint_t(sim)
 {
     this->mmio_addrs = mmio_addrs;
+    this->loggingfd = 0; // unused
 
     if (uartno == 0) {
         // signal handler so ctrl-c doesn't kill simulation when UART is attached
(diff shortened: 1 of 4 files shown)


sagark:
@alonamid @davidbiancolin ready to review.

sagark:
Tests seem to run fine with the default UART switched to UART1

sim/src/main/cc/endpoints/uart.cc
```diff
@@ -24,23 +33,53 @@ void sighand(int s) {
...
+        sigemptyset(&sigIntHandler.sa_mask);
+        sigIntHandler.sa_flags = 0;
+        sigaction(SIGINT, &sigIntHandler, NULL);
+        printf("UART0 is here (stdin/stdout).\n");
+        inputfd = STDIN_FILENO;
+        outputfd = STDOUT_FILENO;
+    } else {
+        // for UARTs that are not UART0, use a PTY
+        char slavename[SLAVENAMELEN];
+        int ptyfd = open("/dev/ptmx", O_RDWR);
```
a0u: `posix_openpt(O_RDWR | O_NOCTTY)` is a portable way to obtain the pty master file descriptor, as not all platforms have `/dev/ptmx` (although not too relevant given that firesim is tied to CentOS).

PR #155, opened 2018-11-09
Add back error device TLError

hqjenny:


sagark:
we're going to ship a release for CCC. Let's hold off on merging this until we have a new dev branch post CCC

PR #180, opened 2018-11-23
Add new gapbs workload

cookiestuf:
Add new gapbs workload to firesim. Only supports the kron synthetic input graph for trivial input size and graph500 input size. Multithreading enabled by default and number of threads default to number of cores on machine. Can verify runs by add `--verify` to `common_args` in json.

changes to Scott's `gapbs/`:
- `*cmd`: command files to generate run scripts
- `gen_run_scripts.sh`: can generate run scripts, binaries and graphs, and json
- `gapbs.sh`: starts benchmark and optionally verifies output
- `.gitignore`: don't complain about new `overlay` directory


docs/Advanced-Usage/Workloads/GAP-Benchmark-Suite.rst
```diff
@@ -0,0 +1,46 @@
...
+.. code-block:: bash
+
+    firesim launchrunfarm    -c workloads/gapbs.ini
+    firesim infrasetup       -c workloads/gapbs.ini
+    firesim runworkload      -c workloads/gapbs.ini
+    firesim terminaterunfarm -c workloads/gapbs.ini
+
+
+Simulation times are host and target dependent. For reference, on a
+four-core rocket-based SoC with a DDR3 + 1 MiB LLC model, with a 160
```
davidbiancolin: FYI, you probably ran this on a 90 MHz target as that's the default now.

deploy/workloads/runscripts/gapbs-scripts/ref.cmd
```diff
@@ -0,0 +1,6 @@
+./bfs -f benchmark/graphs/kron.sg -n64 > benchmark/out/bfs-kron.out
```
davidbiancolin: This doesn't look like the full benchmark; but a copy of graph500.cmd.

PR #182, opened 2018-11-27
Fix Fedora Networking

zhemao:
Fix the switch so that it properly handles multicast requests. Bump firesim-software for fedora network init script fixes.

### DIFF
--- a/target-design/switch/flit.h
+++ b/target-design/switch/flit.h
@@ -69,9 +69,11 @@ uint16_t get_port_from_flit(uint64_t flit, int current_port) {
 
     sendport = sendport & 0xFFFF;
     //printf("mac: %04x\n", sendport);
-    if (sendport != 0xffff) {
-        sendport = mac2port[sendport];
-    }
+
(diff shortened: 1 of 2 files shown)


target-design/switch/flit.h
```diff
@@ -58,8 +60,13 @@ int write_last_flit(uint8_t * send_buf, int tokenid, int is_last) {
...
+    uint16_t is_multicast = (flit >> 16) & 0x1;
     uint16_t flit_low = (flit >> 48) & 0xFFFF; // indicates dest
     uint16_t sendport = (__builtin_bswap16(flit_low));
+
+    if (is_multicast)
+	return BROADCAST_ADJUSTED;
+
     sendport = sendport & 0xFFFF;
     //printf("mac: %04x\n", sendport);
     if (sendport != 0xffff) {
```
sagark: this if() is no longer necessary. can you remove it and add a comment with something like "all mac lookup at this point is non-broadcast" ?

PR #186, opened 2018-11-29
Rework build.sbt to provide a separate midas-targetUtils package

davidbiancolin:
Associated PR: ucb-bar/midas#106
Here's my attempt to tease apart the annotations in MIDAS that you'd want to use in Rocket Chip RTL. Despite my earnest attempts to do things properly, SBT wasn't having it. 

I'm open to better names for the midas package. 

@alonamid

### DIFF
--- a/sim/build.sbt
+++ b/sim/build.sbt
@@ -7,13 +7,9 @@ lazy val commonSettings = Seq(
   traceLevel   := 15,
   scalacOptions ++= Seq("-deprecation","-unchecked","-Xsource:2.11"),
   libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % "test",
-  libraryDependencies += "org.json4s" %% "json4s-native" % "3.5.3",
+  libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.5.3",
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-  addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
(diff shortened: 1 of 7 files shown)


alonamid:
The name makes sense to me (although that as a "utils" package, it might end up being a home to more things that are not only annotations?)

davidbiancolin:
> The name makes sense to me (although that as a "utils" package, it might end up being a home to more things that are not only annotations?)

We're planning on added some of the MIDAS 2.0 transformations here too, specifically those that are target-rtl preserving.

davidbiancolin:
This is ready for review, but i'll merge it in after super node goes in.

alonamid:
Probably need to update the AutoILA docs with the correct package name

davidbiancolin:
@alonamid @sagark  This is ready for merge. I squashed all the intermediate commits since they were busted for various reasons (rebased submodules, SBT bugs).

sim/build.sbt
```diff
@@ -25,14 +21,62 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
...
+  .settings(
+    commonSettings,
+    crossScalaVersions := Seq("2.11.12", "2.12.4"))
+  .dependsOn(chisel, midasTargetUtils)
+lazy val macros     = (project in rocketChipDir / "macros")
+  .settings(commonSettings)
+
+// HACK: I'm strugging to override settings in rocket-chip's build.sbt (i want
+// the subproject to register a new library dependendency on midas's targetutils library)
+// So instead, avoid the existing build.sbt altogether and specify the project's root at src/
```
sagark: classic

PR #195, opened 2018-12-08
DRY out DMA-related transport

davidbiancolin:
Related: ucb-bar/midas#110. 

Uses the PR above to DRY-out DMA-based endpoints.

### DIFF
--- a/sim/src/main/scala/endpoints/SimpleNICWidget.scala
+++ b/sim/src/main/scala/endpoints/SimpleNICWidget.scala
@@ -8,7 +8,7 @@ import freechips.rocketchip.config.{Parameters, Field}
 import freechips.rocketchip.diplomacy.AddressSet
 import freechips.rocketchip.util._
 
-import midas.core._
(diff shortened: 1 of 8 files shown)


davidbiancolin:
This is ready for review. @zhemao @sagark

sim/src/main/scala/endpoints/SimpleNICWidget.scala
```diff
@@ -24,42 +24,6 @@ import TokenQueueConsts._
 
 case object LoopbackNIC extends Field[Boolean]
 
-class SplitSeqQueue(implicit p: Parameters) extends Module {
```
zhemao: So why do we no longer need SplitSeqQueue? Did we resolve the issue with wide BRAM queues?

davidbiancolin: A while back I changed them to use only one split; so they were basically just BRAMQueues under the hood. The firrtl bug has long since been resolved.

PR #197, opened 2018-12-12
Enable Endpointed Print synthesis

davidbiancolin:
This PR enables print synthesis. 

It's enabled by adding "WithPrintSynthesis" to your SimConfig. 
You have to annotate the arguments using the SynthesizePrintf annotator in midas.targetutils like so:

`printf (SynthesizePrintf("x%d p%d 0x%x\n", rf_waddr, rf_waddr, rf_wdata)) `

NB: It currently uses the format string to identify the printf, since you can't actually annotate the node. 

You can synthesize arbitrarily mainly printfs however, the current implementation aggregates all of the arguments into a single token, and can only compress out a token when *all* prints are disabled for that cycle given cycle, this means there's a bunch of wasted bandwidth in cases like: https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/rocket/RocketCore.scala#L844

Arguments:
+print-human-readable -> writes out the printf-formatted data (instead of a binary)
+printfile -> output file name
+print-start -> the cycle at which to start pulling print tokens from the target (saves DMA bandwidth to get you quickly to the ROI)
+print-end -> stops pulling tokens 

Testing:
- I added two midas examples (one packs and one serializes tokens)
  - The synthesized print log is diffed against the prints collected stderr as part of the scalatest

- Synthesizing the printf on the coreMonitorBundle in RC runs at:
  - 5 MHz when writing a binary
  - 250 KHz when formatting (not optimized for performance)
  - A lot of performance can be had simply by disabling the print when coreMonitorBundle.valid is not set, or using a more parsimonious print.


alonamid:
@a0u has identified a slight bug/conflict with the `ILATopWiring` transform due to the joint use of the `TopWiring` annotation. 
Seems like in https://github.com/firesim/firesim/blob/c1e2781150c714f7b3949c085e6f78e6acf3f86e/sim/src/main/scala/passes/ILATopWiring.scala#L135 ,
`state.annotations` includes the TopWiring annotations from the print-synthesis feature, which results in the printed signals being wired to an ILA in addition to the print endpoint.
One potential solution would be to remove `state.annotations` from that line in `ILATopWiring`, and add them back to the state of the new circuit, but I'm not sure if these annotations may be required for this part of the TopWiring transform: https://github.com/freechipsproject/firrtl/blob/4253791132c5c550e1bc4a8070cb54e558f17809/src/main/scala/firrtl/transforms/TopWiring.scala#L230

Thoughts?

davidbiancolin:
That is a dumb oversight on my part -- i think the easiest thing to do is to remove annotations I add in print synthesis after top wiring is complete.

davidbiancolin:
@sagark @alonamid This is ready for review.

PR #199, opened 2018-12-29
Driver kill

NathanTP:
If the hw image name is long, the driver process name might exceed 15 characters. pkill/pgrep can only match the first 15 characters (a limitation from /proc/PID/status). Anyway, truncating the name allows it to match, whether or not the name was long (I think it's safe to assume there won't be anything with the same prefix running).

Testing:
workload: two bare-metal binaries, one prints and exits, the other infinite loops
agfis:
  * memblade (driver name = FireSimMemBlade-f1)
  * firesim-quadcore-nic-ddr3-llc4mb (driver name = FireSim-f1)


NathanTP:
@zhemao FYI

PR #204, opened 2019-01-15
Heterofs

NathanTP:
This branch adds two changes to how rootfs are handled:

Heterogenous rootfs:
For example, you can now run fedora on one node, and buildroot on the other. Before this, you could customize each rootfs with, e.g., custom input files, but they still had to be based off some common_rootfs option. This is no longer required.

Optional Rootfs:
Workloads no longer require a rootfs to work. This is useful for bare-metal tests and avoids the need for dummy rootfs's. As a result, this removes the old default behavior of inferring a rootfs path from the workload name (I'm not aware of any existing workload that relies on this behavior).

Testing:
Tested on single node and 4-node configurations. Heterogenous workloads including a bare-metal, fedora, and several buildroot-based images ran together (the bare-metal was tested both with and without a dummy rootfs). These tests are not included to avoid cluttering the workloads directory. The new workload management system will solve that issue when it comes out and we can re-add unit tests then.


deploy/runtools/runtime_config.py
```diff
@@ -100,12 +100,19 @@ def get_boot_simulation_command(self, macaddr, blkdev, slotid,
         # the sed is in there to get rid of newlines in runtime confs
         driver = self.get_local_driver_binaryname()
         runtimeconf = self.get_local_runtimeconf_binaryname()
-        basecommand = """screen -S fsim{slotid} -d -m bash -c "script -f -c 'stty intr ^] && sudo ./{driver} +permissive $(sed \':a;N;$!ba;s/\\n/ /g\' {runtimeconf}) +macaddr0={macaddr} +blkdev0={blkdev} +slotid={slotid} +niclog0=niclog {tracefile} +trace-start0={trace_start} +trace-end0={trace_end} +linklatency0={linklatency} +netbw0={netbw} +profile-interval={profile_interval} +zero-out-dram +shmemportname0={shmemportname} +permissive-off +prog0={bootbin} && stty intr ^c' uartlog"; sleep 1""".format(
-                slotid=slotid, driver=driver, runtimeconf=runtimeconf,
-                macaddr=macaddr, blkdev=blkdev, linklatency=linklatency,
-                netbw=netbw, profile_interval=profile_interval,
-                shmemportname=shmemportname, bootbin=bootbin, tracefile=tracefile,
-                trace_start=trace_start, trace_end=trace_end)
+
```
sagark: is there supposed to be any functionality change here, or is it just reformatted?

NathanTP: factoring out the driver args makes it possible to add optional args that get added later (in this case the rootfs option). Line 111-112 would be awkward if the entire command was a single string.

PR #207, opened 2019-01-18
Expand functional model provisions for FRFCFS models

davidbiancolin:
Default FRFCFS functional models were under provisioned.


alonamid:
Wasn't there also something about `stallEventCounters = true` ?

alonamid:
And what about `FRFCFS16GBQuadRankLLC4MB`?

davidbiancolin:
> Wasn't there also something about `stallEventCounters = true` ?
 
Forgot; but not a bug fix. Add it next time. 

> And what about `FRFCFS16GBQuadRankLLC4MB`?

All configs that use the FRFCFS MAS extend the one i changed.

Issue #208, opened 2019-01-18
Print synthesis emits spurious data for large batch sizes 


davidbiancolin:
DMA-ing more than 16 beats seems to return faulty data. 

Notes:
- Not reproducable in ML simulation

PR #210, opened 2019-01-19
Print synthesis improvements & bug fixes

davidbiancolin:
This bumps MIDAS for a number of print-synthesis bug fixes and QoL improvements.  It also adds an annotator object for FpgaDebug because i was too lazy to use the existing style. 

Associated PR: ucb-bar/midas#111

One disturbing discovery i had was that DMA reads could drop beats when the destination buffer spanned page boundaries. I'm unsure why this hasn't been a problem for NIC / TracerV. I opened an issue (firesim/firesim#208).  I tried aligning the buffer to an 8 byte boundary and that was still insufficient. Maybe we should consider adding an alignment check in simif_t::pull. @sagark


davidbiancolin:
This is ready for review.

PR #218, opened 2019-01-26
Bump to upstream aws-fpga 1.4.6

sagark:
Notes:
- Users on 1.4.0 AMI appear to be able to still use AGFIs generated with 1.5.0 AMI + 1.4.6 aws-fpga shell (aka this branch still works on the 1.4.0 ami)
- Users on 1.5.0 AMI appear to be able to still use AGFIs generated with older AMI/shell, but this PR includes fresh images
- Updated docs
- ISCA workloads produce correct results
- FPGA flashing bug is dealt with as far as I can tell right now. It looks like there's some bug in the way the AMI re-loads the XOCL driver (which is included by default now in the AMI, and conflicts with the XDMA driver we use). So the FPGA flashing process is now: remove XOCL/XDMA/EDMA, insert XDMA, flash FPGAs, remove XOCL/XDMA/EDMA, insert XDMA, which works reliably. I was able to do 43 iterations of the fpga flash stress unit test with this and also was able to successfully reproduce all isca paper results. Also able to do 21 iterations of follow up fpga flash stress that also actually runs the simulation (boot linux and poweroff).
- See https://github.com/firesim/firesim-performance-result-regression/tree/9cb379a591679b281afba6c686b66849cdc2e001 for performance results (you must be owner on firesim to see)
- XSim shows instructions executing


deploy/runtools/run_farm.py
```diff
@@ -307,41 +312,58 @@ def get_and_install_aws_fpga_sdk(self):
...
                 run('make')
 
-    def unload_edma(self):
-        self.instance_logger("Unloading EDMA Driver Kernel Module.")
+    def unload_xdma(self):
+        self.instance_logger("Unloading XDMA/EDMA/XOCL Driver Kernel Module.")
+
         with warn_only(), StreamLogger('stdout'), StreamLogger('stderr'):
+            # fpga mgmt tools seem to force load xocl after a flash now...
+            # so we just remove everything for good measure:
```
davidbiancolin: lol.

sagark:
all the outputs in that repo are just the stuff generated by run-all.sh in firesim/deploy/workloads/, there's no additional postprocessing

PR #219, opened 2019-01-27
Fix AMI selection instructions

sagark:
Amazon removed the ability to launch 1.4.0 AMIs the regular way for some reason. This updates the docs to show how to still launch manager instances with the 1.4.0 AMI. This is supposed to go onto  master so new users can launch instances.


sagark:
on master, it just won't work

PR #222, opened 2019-01-30
Additional docs example for RTL simulation

dnshp:
Added an example for running all tests with waveform dumping (need to use run-<asm|bmark>-tests-debug, which is not mentioned in original docs)


docs/Advanced-Usage/Debugging/RTL-Simulation.rst
```diff
@@ -102,7 +102,14 @@ Run all RISCV-tools assembly and benchmark tests on a verilated simulator.
     make DESIGN=FireSimNoNIC
     make DESIGN=FireSimNoNIC -j run-asm-tests
     make DESIGN=FireSimNoNIC -j run-bmark-tests
+    
+Run all RISCV-tools assembly and benchmark tests on a verilated simulator with waveform dumping.
 
+::
+
+    make DESIGN=FireSimNoNIC verilator-debug
```
davidbiancolin: you should be able to just `make DESIGN=FireSimNoNIC`, and the the latter recipes will make sure the simulator is built (verilator by default).

PR #225, opened 2019-01-31
MIDAS Examples Build Support in Manager

davidbiancolin:
This PR updates the midas examples generator and scalatests to more closely resemble Firesim. This makes it easier to extend the manager to deploy midas examples, and DRYs out a bunch of generator stuff. 

I put target-agnostic utilities in the package `firesim.util`. 

This change is smaller than it appears as it is mostly shuffling scala around to reuse utilities in FireSim's generator in the midas examples.


davidbiancolin:
@sagark @alonamid This is ready for review.

PR #231, opened 2019-02-03
Add TARGET_PROJECT for testing FASED memory timing-models

davidbiancolin:
Associated MIDAS PR ucb-bar/midas#114.  

In preparation for presenting FASED at FPGA2019, i want FireSim to subsume all of the testing from my external repo. There's a bunch of code duplication which we can sort out in the future, specifically in the Driver, and Makefrags. 

Summary of changes: 
- Add a `fasedtests` target_project that uses the TL fuzzer as a target-design to fuzz the memory model
- Adds support for passing the diplomatic edge that drives the AXI4 slave interface to FASED
  - IdReuse, MaxRequestLength (and in some cases, Max{Reads, Writes}) are now automatically set
- Fixes a regression introduced by #207, where the LLC would not be instantiated correctly


davidbiancolin:
@sagark This is ready to review. I'm breaking up the larger PR. Easier to review + more regression points with working AGFIs.

sim/src/main/scala/firesim/Generator.scala
```diff
@@ -44,7 +44,7 @@ trait HasFireSimGeneratorUtilities extends HasTargetAgnosticUtilites with HasTes
 
     val portList = target.getPorts flatMap {
       case Port(id: DebugIO, _) => None
-      case Port(id: AutoBundle, _) => None // What the hell is AutoBundle?
```
sagark: also lol

davidbiancolin: @donggyukim :rofl:

davidbiancolin:
> lgtm, fresh AGFIs?

Done!

PR #232, opened 2019-02-05
CXXOPTS & Updates synthesized printf driver plusargs and defaults

davidbiancolin:
See ucb-bar/midas#115
- Based off feedback from Jerry and Albert, changes +args to synthesized-prints to be more sensible
  - Makes human-readable and cycle-prefix on by default
  - +print-human-readable -> +print-binary
  - +print-cycle-prefix -> +print-no-cycle-prefix

- Adds separate CXX optimization flags for driver (-O2), verilator(-O0), VCS(-O2)
  - updates runtime charts in docs, adds compile time table


davidbiancolin:
@sagark This is ready for review.

PR #245, opened 2019-02-24
need to install packages with pip2

sagark:


sagark:
this is just to fix the matplotlib install for generating memcached plots

i snuck the paramiko fix into nathan's PR, it just installs a particular version of the crypto library

PR #250, opened 2019-03-01
Flow Control via Ethernet Pause Frames

zhemao:
Implements Ethernet flow control using pause frames. 

https://en.wikipedia.org/wiki/Ethernet_flow_control

Related PRs:

* https://github.com/firesim/firechip/pull/23
* https://github.com/firesim/icenet/pull/9
* https://github.com/firesim/network-benchmarks/pull/3

### DIFF
--- a/target-design/switch/baseport.h
+++ b/target-design/switch/baseport.h
@@ -58,8 +58,10 @@ int BasePort::push_input(switchpacket *sp)
     int ethtype, ctrl, quanta;
 
     // Packets smaller than three flits are too small to be valid
-    if (sp->amtwritten < 3)
+    if (sp->amtwritten < 3) {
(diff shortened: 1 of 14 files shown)


target-design/switch/baseport.h
```diff
@@ -45,6 +53,28 @@ BasePort::BasePort(int portNo, bool throttle)
 {
 }
 
+int BasePort::push_input(switchpacket *sp)
+{
+    int ethtype, ctrl, quanta;
+
+    // Packets smaller than three flits are too small to be valid
+    if (sp->amtwritten < 3)
+        return 0;
```
sagark: seems like this is conflating errors with logging? should we just abort in this case?

PR #264, opened 2019-03-24
[WIP] Rebar refactoring

davidbiancolin:
Here's an initial pass at the rebar refactoring. It moves the FireSim-class targets into firechip entirely, much as it would be in a general chip project. 

TODO: 
- [ ] build-setup.sh updates. 
- [ ] more extensive build system testing

You now need to `export FIRESIM_STANDALONE` to indicate firesim is being used as top. This is kinda hacky, i'm open to better ways of dealing with this, but i figure since you need to sourceme to do anything else interesting, it's not _marginally_ more hacky.


sim/Makefile
```diff
@@ -35,20 +41,31 @@ $(FIRRTL_JAR): $(shell find $(FIRRTL_SUBMODULE_DIR)/src/main/scala -iname "*.sca
...
+endif
 
 # Include target-specific sources and input generation recipes
 include $(TARGET_PROJECT_MAKEFRAG)
 
 verilog: $(VERILOG)
 compile: $(VERILOG)
 
+# Phony targets for launching the sbt shell and running scalatests
+sbt: $(FIRRTL_JAR)
```
grebe: Do you ever mark these as phony with `.PHONY`?

sim/build.sbt
```diff
@@ -21,62 +21,36 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
 
 testGrouping in Test := isolateAllTests( (definedTests in Test).value )
 
-val rocketChipDir = file("target-rtl/firechip/rocket-chip")
-val fireChipDir  = file("target-rtl/firechip")
+lazy val firesimAsLibrary = sys.env.get("FIRESIM_STANDALONE") == None
```
grebe: I think this is fine. I've used `sys.props` and `-D` options when invoking `sbt` b/c I thought that was a little more explicit, but environment vars are fine too.

sim/Makefile
```diff
@@ -13,16 +17,18 @@ TARGET_PROJECT ?= firesim
...
+TARGET_PROJECT_MAKEFRAG ?= $(firesim_base_dir)/src/main/makefrag/$(TARGET_PROJECT)/Makefrag
 
-default: compile
+ifdef FIRESIM_STANDALONE
+
+$(info Firesim standalone)
+
+base_dir := $(firesim_base_dir)
 
 SBT ?= sbt
```
grebe: Is it intentional that `SBT` and related vars are not defined in the `else`?

sim/Makefile
davidbiancolin: Not this Makefile, will fix.

sim/build.sbt
davidbiancolin: Ah, sounds good.

sim/Makefile
davidbiancolin: Yeah, they are defined in rebar's makefrag.

PR #267, opened 2019-03-28
RDP docs update

dnshp:
Updated documentation on connecting to manager via Remote Desktop to be consistent with "Miscellaneous Tips" procedure.


sagark:
@dnshp can you re-target this PR onto master rather than dev? We can push it out straight away since it's just docs.

PR #271, opened 2019-04-02
Adds quad-core BOOM target config

farzadfch:
This PR adds a quad-core BOOM config. I've built the AGFI and booted Linux.
The maximum number of hart ID bits must be set according to the number of BOOM tiles, otherwise, we get an error at elaboration phase.

### DIFF
--- a/sim/src/main/scala/firesim/TargetConfigs.scala
+++ b/sim/src/main/scala/firesim/TargetConfigs.scala
@@ -1,5 +1,6 @@
 package firesim.firesim
 
+import chisel3.util.{log2Ceil}
 import freechips.rocketchip.config.{Parameters, Config}
 import freechips.rocketchip.tile._
 import freechips.rocketchip.tilelink._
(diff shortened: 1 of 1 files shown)


farzadfch:
`FireSimBoomConfig` does not instantiate an object of `WithNDuplicatedBoomCores`. So this code does not have to work with n=1.
Even if we had a config with `WithNDuplicatedBoomCores(1)`, it would work as well becuase `log2Ceil(1)` returns 1. I can compile with such a config and test it, if you want.

jerryz123:
I think `log2Ceil(1)` returns 0. 

https://github.com/freechipsproject/chisel3/blob/9f77bf54543724594711f0455ef91f51edfbd22e/src/main/scala/chisel3/util/Math.scala#L34-L45

jerryz123:
Also I am going to change FireSim to use a normal BoomConfig, instead of the SmallBoomConfig it is currently building (#272). BoomConfig is a more reasonable representation of a high performance core, but it does use more FPGA resources. You will probably be unable to synthesize a quad-core BOOM with the larger configuration.

davidbiancolin:
I've synthesized quad core, default boom, it fits.

davidbiancolin:
+1 for log2up.

farzadfch:
@jerryz123 SmallBoomConfig better suits my purpose as I may eventually integrate NVDLA with BOOM and default BOOM most probably won't fit. If you can please keep the option for SmallBoomConfig alongside default BOOM (if you can't that's also fine).

davidbiancolin:
No worries, we will keep the small config around because it's good for pipe cleaning and debugging other parts of the flow.

PR #272, opened 2019-04-03
Update FireSimBoomConfig to use normal sized BOOM

### LINKED ISSUE #266: Use larger BoomConfig instead of current SmallBoomConfig
Since wider systembuswidths are supported now.

jerryz123:
Resolves #266

### DIFF
--- a/deploy/sample-backup-configs/sample_config_build_recipes.ini
+++ b/deploy/sample-backup-configs/sample_config_build_recipes.ini
@@ -57,14 +57,14 @@ deploytriplet=None
 # BOOM-based targets
 #[fireboom-singlecore-no-nic-lbp]
 #DESIGN=FireBoomNoNIC
-#TARGET_CONFIG=FireSimBoomConfig
+#TARGET_CONFIG=FireBoomConfig
 #PLATFORM_CONFIG=FireSimConfig
 #instancetype=c4.4xlarge
 #deploytriplet=None
 
 [fireboom-singlecore-no-nic-ddr3-llc4mb]
 DESIGN=FireBoomNoNIC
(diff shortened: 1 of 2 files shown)


davidbiancolin:
This will require AGFI regeneration, at least for the BOOM designs.

farzadfch:
Quad-core normal size BOOM fails at placement. Memory model: FR-FCFS, 4MB LLC
```
[192.168.5.220] out: ERROR: [Place 30-487] The packing of instances into a set of CLBs defined by a pblock constraint could not be obeyed. There are a total of 111900 CLBs in the pblock, o
f which 29331 CLBs are available, however, the unplaced instances require 31140 CLBs. The unavailable CLBs are either taken by placed instances or are blocked due to exclude placement cons
traints. Please analyze your design to determine if the pblock can be resized or the number of LUTs, FFs, and/or control sets can be reduced.
[192.168.5.220] out: 
[192.168.5.220] out: Number of control sets and instances constrained to the pBlock
[192.168.5.220] out:    Control sets: 15198
[192.168.5.220] out:    Luts: 938332 (combined) 940131 (total), available capacity: 895200 
[192.168.5.220] out:    Flip flops: 492514, available capacity: 1790400
[192.168.5.220] out:    NOTE: each CLB can only accommodate up to 4 unique control sets so FFs cannot be packed to fully fill every CLB
```

sagark:
IMO we shouldn't silently change existing configs --- we should add a new config. If the issue is that we don't want people defaulting to the old config (since the name sounds like the "default"), then we should change the old one's name too, so that people using it notice that it has changed (because builds will fail).

jerryz123:
This is now done in Chipyard. Closing

PR #278, opened 2019-04-10
Fix python3 install, centos bumped to 3.6, installing 3.4 no longer creates python3/pip3 links

sagark:
This fixes currently broken master build-setup process.


sagark:
@NathanTP @davidbiancolin this is ready for review

Issue #217, opened 2019-01-26
Add C back to toolchain


sagark:

davidbiancolin:
this should include fixing all relevant workloads, including spec17

PR #290, opened 2019-05-15
Bump Rocket Chip 

### LINKED ISSUE #133: Update riscv-gcc to fix SPEC2017 regressions
Recent riscv-tools bumps to track upstream gnu-toolchain causes regressions in SPEC2017 gcc and perlbench. Previously, we were tracking a version of riscv-gcc which resolved these issues. (see riscv/riscv-gcc#100)

On the next riscv-tools bump, check to make sure the lra-contraints bugfix is present (https://github.com/riscv/riscv-gcc/commit/36e932c683490ddfa5225ad8d842b002a8a74ed4) in the version of riscv-gcc we are using, or revert to using our own submodule pointer for gcc (in a fork of the gnu-toolchain).  Any version greater than 7.2.0 should suffice. 

@donggyukim @cookiestuf @zhemao

davidbiancolin:
This bumps RC to the almost-present-day. The two major changes are:

- Bumps riscv-tools removing a few lingering FireSim specific modifications
  - Will point at the riscv-hosted repo if testing goes well
  - Resolves #133, #217 
- Reworks trace port handling, since it's now diplomatic and **enables TracerV by default**
  - Vec[TracedInsn] are now plumbed out as a HeterogeneousBag

Related PRs:

MIDAS: ucb-bar/midas#123
FireChip: firesim/firechip#26
SW: firesim/firesim-software#25
Prebuilt: firesim/firesim-riscv-tools-prebuilt#3

Todo:
- [x] Ship RISCV tools images
- [x] Share AGFIs
- [x] Remove QEMU submodule in install script

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -83,26 +83,23 @@ else
     cd target-design/firechip/riscv-tools/riscv-gnu-toolchain/build
     make -j16 linux
     cd $RDIR
-    cd sw
-    ./install-qemu.sh
-    cd $RDIR
 fi
 
 echo "export RISCV=$RISCV" > env.sh
 echo "export PATH=$RISCV/bin:$RDIR/$DTCversion:\$PATH" >> env.sh
 echo "export LD_LIBRARY_PATH=$RISCV/lib" >> env.sh
 
-cd "$RDIR/platforms/f1/aws-fpga/sdk/linux_kernel_drivers/xdma"
-make
-
-# Set up firesim-software
-cd $RDIR
-sudo pip3 install -r sw/firesim-software/python-requirements.txt
-
 # commands to run only on EC2
 # see if the instance info page exists. if not, we are not on ec2.
 # this is one of the few methods that works without sudo
 if wget -T 1 -t 3 -O /dev/null http://169.254.169.254/; then
+    cd "$RDIR/platforms/f1/aws-fpga/sdk/linux_kernel_drivers/xdma"
+    make
+
+    # Install firesim-software python libraries
+    cd $RDIR
+    sudo pip3 install -r sw/firesim-software/python-requirements.txt
+
     # run sourceme-f1-full.sh once on this machine to build aws libraries and
     # pull down some IP, so we don't have to waste time doing it each time on
     # worker instances
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -43,7 +43,7 @@ elif [ "$1" = "submodules-only" ]; then
 elif [ "$1" = "fast" ]; then
     git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
     cd firesim-riscv-tools-prebuilt
-    git checkout db68e61662d1ad1ba622165f7be50e5606f5ec58
+    git checkout 5fee18421a32058ab339572128201f4904354aaa
     PREBUILTHASH="$(cat HASH)"
     cd ../target-design/firechip
     git submodule update --init riscv-tools
@@ -82,6 +82,9 @@ else
     cd $RDIR
     cd target-design/firechip/riscv-tools/riscv-gnu-toolchain/build
     make -j16 linux
+    cd $RDIR
+    cd sw
+    ./install-qemu.sh
     cd $RDIR
 fi
 
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -43,7 +43,7 @@ elif [ "$1" = "submodules-only" ]; then
 elif [ "$1" = "fast" ]; then
     git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
     cd firesim-riscv-tools-prebuilt
-    git checkout 499867f3f2da4648ddaecf4dd7b1a59554c9d477
+    git checkout db68e61662d1ad1ba622165f7be50e5606f5ec58
     PREBUILTHASH="$(cat HASH)"
     cd ../target-design/firechip
     git submodule update --init riscv-tools
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -43,7 +43,7 @@ elif [ "$1" = "submodules-only" ]; then
 elif [ "$1" = "fast" ]; then
     git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
     cd firesim-riscv-tools-prebuilt
-    git checkout 4a2f79d5c5a8a93f3e6a83bd32a0926cdb8983c5
+    git checkout 85eb89490a730915a28eca3c25d4496193951565
     PREBUILTHASH="$(cat HASH)"
     cd ../target-design/firechip
     git submodule update --init riscv-tools
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -43,7 +43,7 @@ elif [ "$1" = "submodules-only" ]; then
 elif [ "$1" = "fast" ]; then
     git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
     cd firesim-riscv-tools-prebuilt
-    git checkout 85eb89490a730915a28eca3c25d4496193951565
+    git checkout 499867f3f2da4648ddaecf4dd7b1a59554c9d477
     PREBUILTHASH="$(cat HASH)"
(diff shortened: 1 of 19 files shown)


jerryz123:
Can we bump to May https://github.com/freechipsproject/rocket-chip/commit/b8baef6f26fa36c9b4e0b2a5eec12cd76ae5daf8

davidbiancolin:
I'm not averse to that; this was the commit that @abejgonzalez gave to me.

colinschmidt:
Yeah that basically changes nothing in the rocket-chip API, just a chisel bug fix.

abejgonzalez:
Fine with me. I agree on what @jerryz123 agrees with.

alonamid:
So now the entire TracerV test comparison is done is scala? Were you able to test it?

davidbiancolin:
> So now the entire TracerV test comparison is done is scala? Were you able to test it?

Yes & Yes. I also fixed a bug in the flush. I don't want to change anything more because that'll make the rest of your merge more hellish.

davidbiancolin:
This is ready for review. @sagark @a0u @alonamid @zhemao

sim/src/main/cc/endpoints/tracerv.cc
```diff
@@ -5,6 +5,9 @@
 #include <stdio.h>
 #include <string.h>
 #include <limits.h>
+#include <sstream>
+#include <iostream>
+#include <fstream>
```
alonamid: Do we need this now that you moved the test comparison to be external using scala?

sim/src/main/cc/endpoints/tracerv.h
```diff
@@ -16,13 +16,20 @@ class tracerv_t: public endpoint_t
...
     private:
         TRACERVWIDGET_struct * mmio_addrs;
         simif_t* sim;
         FILE * tracefile;
         uint64_t start_cycle, end_cycle, cur_cycle;
-	long dma_addr;
+        bool human_readable = false;
+        // Used in unit testing to check TracerV is correctly pulling instuctions off the target
+        bool test_output = false;
+	    long dma_addr;
```
alonamid: nit: indentation?

sim/src/main/scala/endpoints/TracerVWidget.scala
```diff
@@ -18,43 +18,79 @@ import icenet.IceNIC._
...
+  // Generates a Chisel type from that returned by a Diplomatic node's in() or .out() methods
+  def fromNode(ports: Seq[(Vec[TracedInstruction], Any)]): Seq[Vec[DeclockedTracedInstruction]] = ports.map({
+    case (bundle, _) => Vec(bundle.length, new DeclockedTracedInstruction(bundle.head.cloneType))
+  })
+}
+
+// The IO matched on by the TracerV endpoint: a wrapper around a heterogenous
+// bag of vectors. Each entry is Vec of committed instructions
+class TraceOutputTop(private val traceProto: Seq[Vec[DeclockedTracedInstruction]]) extends Bundle {
+  val traces = HeterogeneousBag(traceProto.map(_.cloneType))
```
alonamid: Why is this a HeterogeneousBag?

sim/src/main/scala/firesim/TargetMixins.scala
```diff
@@ -7,79 +7,118 @@ import freechips.rocketchip.tilelink._
...
+    sbus.coupleTo("coherence_manager") { in :*= _ }
+    mbus.coupleFrom("coherence_manager") { _ :=* BankBinder(mbus.blockBytes * (nBanks-1)) :*= out }
+  }
+}
+
+
+/** Copied from RC and modified to change the IO type of the Imp to include the Diplomatic edges
+  *  associated with each port. This drives FASED functional model sizing
+  */
+trait CanHaveFASEDOptimizedMasterAXI4MemPort { this: BaseSubsystem =>
```
alonamid: Is this specific to FASED now?

sim/src/main/scala/endpoints/TracerVWidget.scala
davidbiancolin: Different cores might have different traced instructions.

sim/src/main/scala/firesim/TargetMixins.scala
davidbiancolin: Essentially. I renamed this to make it more clear why we need this, since the traits in RC are sufficient to support misaligned memory regions.

sim/src/main/cc/endpoints/tracerv.cc
davidbiancolin: Good point.

sim/src/main/cc/endpoints/tracerv.h
davidbiancolin: Fixed. There was a stray tab.

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -442,5 +450,8 @@ void firesim_top_t::run() {
 #ifdef PRINTWIDGET_0_PRESENT
     print_endpoint->finish();
 #endif
+#ifdef TRACERVWIDGET_0_PRESENT
+    tracerv_endpoint_0->finish();
+#endif
```
zhemao: Don't you need to call finish for the other 7 widgets as well?

sim/src/main/scala/firesim/TargetConfigs.scala
```diff
@@ -54,11 +55,16 @@ class WithBoomL2TLBs(entries: Int) extends Config((site, here, up) => {
...
 
 class WithTraceRocket extends Config((site, here, up) => {
-   case RocketTilesKey => up(RocketTilesKey, site) map { r => r.copy(trace = true) }
+   case RocketTilesKey => up(RocketTilesKey, site)
 })
 
 class WithTraceBoom extends Config((site, here, up) => {
-   case BoomTilesKey => up(BoomTilesKey, site) map { r => r.copy(trace = true) }
+   case BoomTilesKey => up(BoomTilesKey, site)
+})
```
zhemao: Delete these two configs

davidbiancolin:
This is blocked on firesim-software.

build-setup-nolog.sh
```diff
@@ -43,7 +43,7 @@ elif [ "$1" = "submodules-only" ]; then
 elif [ "$1" = "fast" ]; then
     git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
     cd firesim-riscv-tools-prebuilt
-    git checkout 4a2f79d5c5a8a93f3e6a83bd32a0926cdb8983c5
+    git checkout 499867f3f2da4648ddaecf4dd7b1a59554c9d477
```
NathanTP: I don't see this commit in riscv-tools-prebuilt. It looks like the correct hash (going by the commit comments) is:
db68e61662d1ad1ba622165f7be50e5606f5ec58

is there a more recent update that hasn't been pushed?

NathanTP: Also, db68e6 does not include qemu while 4a2f79d does.

davidbiancolin: The db68e6 has should be the correct one. I'm not sure why QEMU wasn't built, i'm trying a fresh full build now to see why.

davidbiancolin: Yeah, so it seems the install-qemu script is responsible for building QEMU, it never happens in riscv-tools build.sh. I somehow broke install-qemu.sh when i manually compiled and installed a bunch of updated packages, not provided by rpm, to build openOCD. 

I think the right call is to provide a second build.sh script in firesim that simply skips building openOCD. That way we don't need a fork of riscv-tools, and we won't need to upgrade to package versions not provided by rpm.

Once i fix that, i'll add the ./install-qemu.sh script back to build-setup.sh. Alternatively, i could remove the additional qemu submodule and add the qemu installation to our firesim build.sh script.

NathanTP: SGTM. I would advocate for adding it to the build-setup.sh script. The model I've been going on for firesim-software is that it needs some version of riscv-tools and qemu, but they aren't submodules. With this stuff all getting upstreamed, there is theoretically less of a need to submodule specific commits etc.

This isn't actually documented though, I should probably add that to the README.

davidbiancolin: Updated.

PR #294, opened 2019-05-29
Major reorganization to enable FireSim-as-a-library

davidbiancolin:
This allows FireSim to be included as a submodule in a project-template-based project.

Code that can be reused across project-template and other target designs have been moved into `firesim-lib/`, this includes the following packages:

- configs: target-agnostic configurations, e.g. different FASED configurations, pulled from SimConfigs.scala
- utils: traits for mixing into a Generator App 

To get this repository to function correctly as a top level and a library i had to resort to some nastiness Now, to run any of the commands out of sim, you'll need to source sourceme-f1-manager.sh, or env.sh. This will tell the make system where to look for REBAR. Build-setup.sh has been modified to support initialization as a library vs as a top. Unfortunately, changing both FireSim and REBAR will require three commits due to a circular submodule dependency:

1) Commit FireSim changes without a REBAR bump to produce a valid ref
2) Commit REBAR changes, bumping FireSim to the the ref in (1)
3) Bump REBAR in FireSim, pointing to the ref in (2)

I think there's a strong case to be made for deprecating FireSim-as-a-top.


.gitmodules
```diff
@@ -4,18 +4,12 @@
...
-[submodule "sim/firrtl"]
-	path = sim/firrtl
-	url = https://github.com/ucb-bar/firrtl
 [submodule "sim/midas"]
 	path = sim/midas
 	url = https://github.com/ucb-bar/midas
-[submodule "sim/barstools"]
-	path = sim/barstools
-	url = https://github.com/ucb-bar/barstools.git
 [submodule "sim/firechip"]
```
abejgonzalez: This is a nitpick now... but if we stick with the FSim as top approach, we should rename this to match the new name of REBAR

davidbiancolin: You're right, i was unsure how much to commit to the name.

sim/firesim-lib/src/main/scala/configs/SimConfigs.scala
```diff
@@ -164,45 +163,45 @@ class FRFCFS16GBQuadRankLLC4MB3Div extends Config(
...
-
-class FireSimDDR3FRFCFSLLC4MBConfig extends Config(
-  new FRFCFS16GBQuadRankLLC4MB ++
-  new FireSimConfig)
-
-class FireSimDDR3FRFCFSLLC4MB3ClockDivConfig extends Config(
-  new FRFCFS16GBQuadRankLLC4MB3Div ++
-  new FireSimConfig)
-
+///*******************************************************************************
```
abejgonzalez: Shouldn't this be deleted? or what is the reasoning to keeping this commented out here?

abejgonzalez:
LGTM. SmallBoomConfig works well.

PR #309, opened 2019-07-10
Module-Based Endpoint System

davidbiancolin:
The real meat in this change lies in the porting of existing FireSim endpoints to use the black boxes. Associated PR ucb-bar/midas#130. 

Changes:
- Replaces legacy EndpointMatchers extending `Endpoint` with:
  1) A target-land endpoint interface
  2) A host-land interface that breaks the target-land interface above into decoupled channels
  3) A BaseModule that extends `IsEndpoint`, to be used in the target RTL
    -> the `io` field is the target land interface (1)
    -> the `endpointIO` field is the host-land interface prototype (2)
    -> `widget()` is a lambda that will elaborate the host-land widget. 
  4) A companion object to helper instantiate the endpoint

- hostPort broken out of io into separate IO field to simplify bundle class hierarchies
  - hence the io.hPort -> hPort changes

- Sim configs have been removed or moved to target configs where appropriate

Other changes:
- a modification to endpoint instance naming convention required some modification of the macros used in firesim_top.cc
- all midas examples broken into a "DUT", the old module, and a wrapper that instantiates the DUT and endpoints.


davidbiancolin:
Closing to change base branch.

PR #328, opened 2019-07-31
Add hostParam option to configure ILA depth

jerryz123:
Lets us easily capture more signals in ILA.

### DIFF
--- a/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
+++ b/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
@@ -40,6 +40,19 @@ themselves (eg. in Chisel3.util.Queue).
 Note: In case the module with the annotated signal is instantiated multiple times,
 all instatiations of the annotated signal will be wired to the ILA.
 
+Setting a ILA Depth
+-------------------
+
+The ILA depth parameter specifies the duration in cycles to capture annotated signals
+after a trigger. Increasing this parameter may ease debugging, but will also increase
+FPGA resource utilization. The default depth is 1024 cycles. The desired depth can be
+configured much like the desired HostFrequency by appending a mixin to the 
+`PLATFORM_CONFIG`.
+
+::
+
+   PLATFORM_CONFIG=FireSimConfig_ILA8192
+
 
 
 Using the ILA at Runtime
--- a/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
+++ b/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
@@ -44,13 +44,15 @@ Setting a ILA Depth
 -------------------
 
 The ILA depth parameter specifies the duration in cycles to capture annotated signals
-after a trigger. Increasing this parameter may ease debugging, but will also increase
+around a trigger. Increasing this parameter may ease debugging, but will also increase
 FPGA resource utilization. The default depth is 1024 cycles. The desired depth can be
 configured much like the desired HostFrequency by appending a mixin to the 
-`PLATFORM_CONFIG`.
+`PLATFORM_CONFIG`. See :ref:`Generating-Different-Targets` for details on `PLATFORM_CONFIG`.
+
+Below is an example `PLATFORM_CONFIG` that can be used in the `build_recipes` config file.
 
 ::
-
+   
    PLATFORM_CONFIG=FireSimConfig_ILA8192
 
 
--- a/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
+++ b/docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
@@ -53,7 +53,7 @@ Below is an example `PLATFORM_CONFIG` that can be used in the `build_recipes` co
 
 ::
    
-   PLATFORM_CONFIG=FireSimConfig_ILA8192
+   PLATFORM_CONFIG=FireSimConfig_ILADepth8192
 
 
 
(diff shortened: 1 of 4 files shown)


docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
```diff
@@ -40,6 +40,19 @@ themselves (eg. in Chisel3.util.Queue).
...
+
+The ILA depth parameter specifies the duration in cycles to capture annotated signals
+after a trigger. Increasing this parameter may ease debugging, but will also increase
+FPGA resource utilization. The default depth is 1024 cycles. The desired depth can be
+configured much like the desired HostFrequency by appending a mixin to the 
+`PLATFORM_CONFIG`.
+
+::
+
+   PLATFORM_CONFIG=FireSimConfig_ILA8192
```
alonamid: Maybe give the context that this is part of the build recipe?

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
```diff
@@ -19,7 +19,7 @@ import firrtl.transforms.TopWiring._
     This also has an option to pass a function as a parmeter to generate custom output files as a result of the additional ports
   * @note This *does* work for deduped modules
   */
-class ILATopWiringTransform(dir: File = new File("/tmp/")) extends Transform {
+class ILATopWiringTransform(dir: File = new File("/tmp/"), depth: Int = 1024) extends Transform {
```
alonamid: Another nit: reading this, depth seems fairly generic. Naming it ila_window_depth might be more explicit?

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
```diff
@@ -63,18 +63,19 @@ class ILATopWiringTransform(dir: File = new File("/tmp/")) extends Transform {
     }
 
     //body
+    val probetriggers = 2
```
alonamid: Do you have a particular reason for changing the default to 2 (other than resource utilization?)

sim/firesim-lib/src/main/scala/util/Configs.scala
```diff
@@ -53,5 +58,17 @@ class  F65MHz extends WithDesiredHostFrequency(65)
 class  F60MHz extends WithDesiredHostFrequency(60)
 class  F55MHz extends WithDesiredHostFrequency(50)
 class  F50MHz extends WithDesiredHostFrequency(50)
+class  F45MHz extends WithDesiredHostFrequency(45)
+class  F40MHz extends WithDesiredHostFrequency(40)
+class  F35MHz extends WithDesiredHostFrequency(35)
+class  F30MHz extends WithDesiredHostFrequency(30)
```
alonamid: lol, sneaking in some lower Host Frequencies in this PR?

docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
```diff
@@ -40,6 +40,19 @@ themselves (eg. in Chisel3.util.Queue).
 Note: In case the module with the annotated signal is instantiated multiple times,
 all instatiations of the annotated signal will be wired to the ILA.
 
+Setting a ILA Depth
+-------------------
+
+The ILA depth parameter specifies the duration in cycles to capture annotated signals
+after a trigger. Increasing this parameter may ease debugging, but will also increase
+FPGA resource utilization. The default depth is 1024 cycles. The desired depth can be
+configured much like the desired HostFrequency by appending a mixin to the
```
alonamid: I would like @davidbiancolin 's opinion here if the platform config suffix is the correct way to go here

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
jerryz123: Resource utilization, I've never needed to write a trigger of the form 'probe === a OR probe === b'. I think we should optimize for the common case.

sim/firesim-lib/src/main/scala/util/Configs.scala
jerryz123: Increasing window size necessitates setting a lower target frequency, I've had to go below 50 MHz for large boom configs with L2, asserts, and deep ILA.

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
jerryz123: I can name it data_depth, to match what the actual variable is called

alonamid: sure

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
alonamid: what about trying to trigger on combinations of ready && valid ? (without an explicit fire signal)

sim/firesim-lib/src/main/scala/util/Configs.scala
alonamid: ok

sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
jerryz123: Thats fine when `probetriggers = 2`.

What doesn't work is a trigger like
`count === 3'h1 || count === 3'h2`

docs/Advanced-Usage/Debugging/Debugging-Hardware-Using-ILA.rst
davidbiancolin: Yeah that's perfectly fine, though maybe we'd want a more descriptive name, even just ILADepth<NUM> in case there are more of these in the future. Such as the number of probe triggers.

PR #331, opened 2019-08-20
Preliminary Golden Gate (MIDAS II) Support

davidbiancolin:
See ucb-bar/midas#134 for a longer description. 

This PR introduces support for the next version of the MIDAS compiler, called _Golden Gate_.  Currently, Golden Gate generates token channels and initializes them in the same fashion that MIDAS does in order to give FMR = 1 assuming no other host delays. The only change this PR introduces to FireSim users is support for doing multi-ported RAM optimizations (e.g., BOOM's reg file), which provides a means to fit a substantially larger target machine one on FPGA (in our ICCAD paper we went from fitting 4 to 6 large BOOM cores on a single VU9P).


deploy/post-build-scripts/publish-to-s3.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#Pushes a results-build subdirectory to an S3 bucket
+
+result_dir=$1
+dest_dir=$(basename ${result_dir})
+bucket_name=iccad2019-midas2
```
alonamid: Do you still want to leave these here?

deploy/sample-backup-configs/sample_config_build_recipes.ini
```diff
@@ -5,12 +5,12 @@
 # edit config_build.ini to actually "turn on" a config to be built when you run
 # buildafi
 
-#[firesim-singlecore-nic-lbp]
-#DESIGN=FireSim
-#TARGET_CONFIG=FireSimRocketChipSingleCoreConfig
-#PLATFORM_CONFIG=FireSimConfig
-#instancetype=c4.4xlarge
-#deploytriplet=None
+[firesim-singlecore-nic-lbp]
```
alonamid: Are we changing our default provided AFIs?

sim/firesim-lib/src/main/scala/endpoints/BlockDevWidget.scala
```diff
@@ -77,8 +77,9 @@ class BlockDevWidget(implicit p: Parameters) extends EndpointWidget()(p) {
   rRespBuf.reset  := reset.toBool || targetReset
   wAckBuf.reset  := reset.toBool || targetReset
 
-  io.hPort.toHost.hReady := tFireHelper.fire(io.hPort.toHost.hValid)
-  io.hPort.fromHost.hValid := tFireHelper.fire(io.hPort.fromHost.hReady)
+  // Hack
```
alonamid: Is it possible to specify in the comment the reason for the hack? (the `fixMeOnNextRocketBump is some indication, but still...)

sim/firesim-lib/src/main/scala/endpoints/SimpleNICWidget.scala
```diff
@@ -190,9 +190,13 @@ class SimpleNICWidget(implicit p: Parameters) extends EndpointWidget()(p)
   val NICtokenToBig = Module(new NICTokenToBigTokenAdapter)
 
   val target = io.hPort.hBits
-  val tFire = io.hPort.toHost.hValid && io.hPort.fromHost.hReady && io.tReset.valid
-  val targetReset = tFire & io.tReset.bits
-  io.tReset.ready := tFire
+  val fixMeOnNextRocketBump = true.B // Need a dummy predicate to exclude for now
```
alonamid: Why would a Rocket bump resolve this?

sim/firesim-lib/src/main/scala/endpoints/TracerVWidget.scala
```diff
@@ -97,7 +97,7 @@ class TracerVWidget(traceProto: Seq[Vec[DeclockedTracedInstruction]])(implicit p
 
   io.tReset.ready := tFireHelper.fire(io.tReset.valid)
   io.hPort.fromHost.hValid := tFireHelper.fire(io.hPort.fromHost.hReady)
-  io.hPort.toHost.hReady := tFireHelper.fire(io.hPort.toHost.hValid)
+  io.hPort.toHost.hReady := tFireHelper.fire
```
alonamid: Why do you not need to `hValid` anymore?

sim/src/main/makefrag/midasexamples/Makefrag
```diff
@@ -65,6 +65,8 @@ TARGET_LD_FLAGS :=
 # PointerChaser requires a custom memory initialization
 ifeq ($(DESIGN),PointerChaser)
 LOADMEM ?= $(GENERATED_DIR)/mem_init.hex
+#LOADMEM ?=
```
alonamid: Do you need this line?

deploy/sample-backup-configs/sample_config_build_recipes.ini
davidbiancolin: I uncommented this to do some testing; i can put it back in

sim/firesim-lib/src/main/scala/endpoints/SimpleNICWidget.scala
davidbiancolin: DecoupledHelper.fire() can take no arguments to take the conjunction of all booleans provided to it.

sim/firesim-lib/src/main/scala/endpoints/TracerVWidget.scala
davidbiancolin: This is actually related to the uncommented //Hack above. Here i've made ready depend on valid, since ready fans out to all of the token channels in that feed tokens into this endpoint. If one of those token channels has it's "sub-token" valid before the rest, it will be falsely dequeued. Really i should fuse the token properly by making the ready for each sub-token depend on AND {valids of all other sub-tokens} instead of doing it this way.

deploy/post-build-scripts/publish-to-s3.sh
davidbiancolin: Good point.

sim/src/main/makefrag/midasexamples/Makefrag
davidbiancolin: Good catch.

davidbiancolin:
> If I understand correctly, the PR is not only preliminary Golden Gate support, but also an update to a testing workload but just booting Linux and immediately shutting it off
Yup; i use it to quickly check all no-nic targets can boot BR linux

> as well as and update to the Rocket debug port?
That is removing a hack in the Generator to not neglect the port (it doesn't show up anymore)

zhemao:
I would prefer that you fix the token backpressure issue properly before we merge this.

davidbiancolin:
> I would prefer that you fix the token backpressure issue properly before we merge this.

It's fixed in the endpoint branch and is tied to those changes.

PR #333, opened 2019-08-22
Reorder build process and fix sw requirements

tmagik:
This is mostly to PR to generate some discussion on better ways to simplify the build.

If you like it, please merge, otherwise some things I'd like commentary on are:

- possible use of freedom-u-sdk and/or sifive binaries for RiscV toolchain
- how replace 'firesim.pem' auth with something more likely to survive a security audit
- non-AWS FPGA sim platforms, such as x86 or RiscV hosts with Xilinx U200 ( see https://github.com/tmagik/XRT/tree/riscv-hacks )

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
 # FireSim initial setup script. This script will:
-# 1) Initalize submodules (only the required ones, minimizing duplicates
-# 2) Install RISC-V tools, including linux tools
-# 3) Installs python requirements for firesim manager
+# 1) Installs python requirements for firesim manager
+# 2) Initalize submodules (only the required ones, minimizing duplicates
+# 3) Install RISC-V tools, including linux tools
 
 # TODO: build FireSim linux distro here?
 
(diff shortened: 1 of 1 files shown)


davidbiancolin:
Hey Troy,

Sorry for the tardy response, and thanks for your first PR. You touch on a number of different things so let me try to delegate a bit. 

Can you explain the rationale for the reordering? I think the change to FireSim software is sensible but i'll punt to @NathanTP on that. 

> * possible use of freedom-u-sdk and/or sifive binaries for RiscV toolchain
Firstly, FireSim is undergoing a bunch of churn as we move to make it more of a library under Chipyard. We've changed how we build target toolchains as part of that work, in that it's being punted to the enclosing project. Development on FireSim is done against the `dev` branch; we merge into `master` for releases. Have a look at the build-setup.sh on `dev`, and we can talk about how to make it more useful in other contexts where FireSim is being submoduled.

> * how replace 'firesim.pem' auth with something more likely to survive a security audit
@sagark 
Admittedly, this smells pretty bad but but we've had little need to change it because we use these keys only on relatively isolated EC2 instances to facilitate communication between manager -> {runfarm, buildfarm}. You could well use a different key to log onto your manager instance. To do local dev, where the firesim manager serves no purpose, we provide no key and just ignore the build-setup.sh message. 

> * non-AWS FPGA sim platforms, such as x86 or RiscV hosts with Xilinx U200 ( see https://github.com/tmagik/XRT/tree/riscv-hacks )

FireSim has long been tied to EC2 F1, but we supported other 7-series Zynq boards in the past (ZC706, Zedboard) in the past. There are a few hurdles to bringing this up, I've described them previously [here](https://groups.google.com/forum/#!searchin/firesim/ZYnq%7Csort:date/firesim/IrIVNMeRCuw/8Jl24YwZBAAJ). XRT looks like it could simply this effort substantially; i'll have to read into that further. How critical is this to you?

NathanTP:
The PR to firesim-software is fine (I approved it there). Although it should point to dev instead of master.

As for the re-ordering of requirements installation. I have no opinion there. As long as they get installed before running any marshal commands.

PR #339, opened 2019-09-09
[docs] Provide more detail about tracerv +args

davidbiancolin:
This is useful for people who are using the trace during MIDAS-level simulation.

### DIFF
--- a/docs/Advanced-Usage/Debugging/TracerV.rst
+++ b/docs/Advanced-Usage/Debugging/TracerV.rst
@@ -31,12 +31,12 @@ the "tracing" section.
     startcycle=XXXX
     endcycle=YYYY
 
-TracerV Runtime Arguments (+args)
----------------------------------
+MIDAS- & FPGA-Level Simulation Arguments (+args)
+------------------------------------------------
 
 Under the hood, TracerV settings in your config_runtime.ini are being translated to plusargs 
 that are passed the simulator during ``firesim runworkload``. You can pass these same plusargs to MIDAS- and FPGA-level
-simulators if you invoke them directly or by setting ``EXTRA_SIM_ARGS=<plusarg string>`` if you invoke them through ``make``.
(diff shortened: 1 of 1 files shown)


alonamid:
I think this is actually a broader topic. Do we generally describe +args in the docs for other widgets? I thought the decision was to abstract stuff using the manager and the config files.
If we are starting to describe +args, I think it need to be appropriately distinguished/separated, and explain that this is not the recommended method of use? Maybe centralized under a separate +args docs page?

davidbiancolin:
The fundamental problem is (1) inis only work for deployment to the run farm and not for use with a midas-level simulator and (2) the inis only provide sugar for a very limited subset of plusargs. They don't, for instance, support synthesized printfs, the dram model, or block device arguments. 

I agree: ideally we'd only have one way of doing things.

alonamid:
I think the main question is how prevalent is midas-level simulation (i.e. - non-runfarm simulation). Under the assumption that it is used only by firesim devs, or new device developers, perhaps the best place for plus-args docs is under the midas-level sim page of the docs?

davidbiancolin:
Realistically, until we bring up more accurate target-land simulation of the same simulated system, pretty much anyone changing RTL will need to use MIDAS-level simulation.

docs/Advanced-Usage/Debugging/TracerV.rst
```diff
@@ -31,5 +31,25 @@ the "tracing" section.
     startcycle=XXXX
     endcycle=YYYY
 
+TracerV Runtime Arguments (+args)
```
sagark: Maybe add "midas-level, fpga-level sw simulation only" or something similar to the title here?

sagark:
I think the current location @davidbiancolin added this is better, esp since how to use the tracer the "ini way" is also described on this page.

docs/Advanced-Usage/Debugging/TracerV.rst
davidbiancolin: That's sounds good.

PR #341, opened 2019-09-19
Module-Based Endpoint System

### LINKED ISSUE #335: HostReady backpressure in endpoints depends on HostValid
[Applies only to dev; once the midas2 branch is merged]

Backpressure in endpoints on endpoint-sunk tokens now depends on valid to prevent prematurely dequeuing a sub-token. This wasn't required in MIDAS I because the target could always enqueue all output tokens simultaneously. 

Fix by correctly tokenizing interface (tell Golden Gate this endpoint consumes a single input token and a produces a single output token), or in the interim, by using a ready-valid helper when aggregating multiple tokens together.

davidbiancolin:
The real meat in this change lies in the porting of existing FireSim endpoints to use the black boxes. Associated PR ucb-bar/midas#130. 

Changes:
- Replaces legacy EndpointMatchers extending `Endpoint` with:
  1) A target-land endpoint interface
  2) A host-land interface that breaks the target-land interface above into decoupled channels
  3) A BaseModule that extends `IsEndpoint`, to be used in the target RTL
    -> the `io` field is the target land interface (1)
    -> the `endpointIO` field is the host-land interface prototype (2)
    -> `widget()` is a lambda that will elaborate the host-land widget. 
  4) A companion object to helper instantiate the endpoint

- hostPort broken out of io into separate IO field to simplify bundle class hierarchies
  - hence the io.hPort -> hPort changes

- Sim configs have been removed or moved to target configs where appropriate

Other changes:
- a modification to endpoint instance naming convention required some modification of the macros used in firesim_top.cc
- all midas examples broken into a "DUT", the old module, and a wrapper that instantiates the DUT and endpoints.

Resolves #335


sagark:
Are there docs that go with this?

sagark:
I re-ran ISCA workloads on this:
https://github.com/firesim/firesim-performance-result-regression/commit/79446cbd61d8a9289c094ee9dd8b4ba130288056

Everything looks good although we've lost some simulation performance along the way.

PR #344, opened 2019-09-27
[config] Prepend FASEDEndpoints to EndpointMap in configs

davidbiancolin:
Since the updated Endpoint matcher class was appended instead of prepended to the EndpointMap, channel generation would see the older Endpoint definition first, and use the old frequency division.

This only affects configs that include at least _two_ instances of DefaultMemModel, with at least two different requested frequency divisions. Additionally, this only affects channel generation, the memory model constructor still gets the expected parameterization.

None of the default AGFIs use this feature.


davidbiancolin:
EndpointMap gets a final laugh.

PR #345, opened 2019-09-27
Use GNU make 4.x for toolchain build

a0u:
This is needed for `scripts/firesim-setup.sh` with ucb-bar/chipyard#261.

make 3.82 is still the default for CentOS release 7.6 (FPGA Developer AMI 1.6.0), which is too old for cross-compiling glibc 2.28 and newer.

The second commit includes minor cleanup to avoid unnecessary pollution of the global `.gitconfig`.

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -61,20 +61,9 @@ do
     shift
 done
 
-# We're a submodule of rebar, so don't initalize the submodule
-# ignore riscv-tools for submodule init recursive
-# you must do this globally (otherwise riscv-tools deep
-# in the submodule tree will get pulled anyway
-git config submodule.toolchains/riscv-tools.update none
-git config --global submodule.experimental-blocks.update none
-git config --global submodule.sims/firesim.update none
 # Disable the REBAR submodule initially, and enable if we're not in library mode
 git config submodule.target-design/chipyard.update none
 git submodule update --init --recursive #--jobs 8
-# unignore riscv-tools,catapult-shell2 globally
-git config --global --unset submodule.sims/firesim.update
-git config --unset submodule.toolchains/riscv-tools.update
-git config --global --unset submodule.experimental-blocks.update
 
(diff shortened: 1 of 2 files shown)


build-setup-nolog.sh
```diff
@@ -61,20 +61,9 @@ do
     shift
 done
 
-# We're a submodule of rebar, so don't initalize the submodule
-# ignore riscv-tools for submodule init recursive
-# you must do this globally (otherwise riscv-tools deep
-# in the submodule tree will get pulled anyway
-git config submodule.toolchains/riscv-tools.update none
```
davidbiancolin: Yes, thank you.

scripts/machine-launch-script.sh
```diff
@@ -12,6 +12,10 @@ sudo yum install -y python36 patch diffstat texi2html texinfo subversion chrpath
 sudo yum install -y gtk3-devel
 # deps for firesim-software (note that rsync is installed but too old)
 sudo yum install -y python36-pip python36-devel rsync
+# Install GNU make 4.x (needed to compile riscv-glibc 2.28+)
```
davidbiancolin: We need to announce this very publicly in the appropriate channels.

a0u: Yeah.  On further thought, not everyone might be using the AMI, so it is probably better to enable devtoolset-7 only if it exists and the make version is too old.

build-setup-nolog.sh
```diff
@@ -102,6 +91,16 @@ else
     target_chipyard_dir=$RDIR/target-design/chipyard
 fi
 
+# Enable latest Developer Toolset for GNU make 4.x
+devtoolset=''
+for dir in /opt/rh/devtoolset-* ; do
```
a0u: `scripts/build-toolchains.sh` in Chipyard now explicitly checks for the make version, so sourcing devtoolset can be made optional (to accommodate non-CentOS hosts).  Having an obsolete make version will still be caught early in the build process.

PR #346, opened 2019-09-27
Serializable Endpoint/Bridge Annotations & Golden Gate Stage

davidbiancolin:


sagark:
this lgtm. will approve on agfis etc.

albert-magyar:
Also LGTM. Good idea moving things that produce artifacts into a single source file.

PR #352, opened 2019-10-04
Add SHA-3 demo recipe

a0u:
This adds a `firesim-singlecore-sha3-l2-nic-ddr3-llc4mb` recipe for the tutorial.  The LLC model might not be necessary with the L2 present, but it's used here to be consistent with other recipes.

The chipyard submodule points to ucb-bar/chipyard#265.


a0u:
The chipyard pointer needed to be changed to fix a CI failure.

a0u:
Passed CI.

PR #354, opened 2019-10-04
Move FASED Configs into TARGET_CONFIG strings

davidbiancolin:
[ I will generate AGFIs for this first. And then rebase the serializable endpoint stuff off these to regenerate a new set of AGFIs in parallel if these changes are acceptable.]

As part of the endpoint rework we've pushed much of the configuration of FASED into the generation of the target. This updates the build recipes to reflect that, so as to generate the correct memory timing model.

It also splits up what remains of SimConfigs into:
- CompilerConfigs.scala -> Configs that will be mixed into a parameters instance used by GG
- FASEDTargetConfigs.scala -> Target-land configs for instantiating FASED memory models

I'm not convinced these are the best names, but in a prior meeting we seemed to like using "Target" and "Compiler"  in the names to distinguish between the target's configs and Golden Gate's configs.

### DIFF
--- a/deploy/sample-backup-configs/sample_config_build_recipes.ini
+++ b/deploy/sample-backup-configs/sample_config_build_recipes.ini
@@ -8,49 +8,49 @@
 [firesim-singlecore-nic-lbp]
 DESIGN=FireSim
 TARGET_CONFIG=FireSimRocketChipSingleCoreConfig
-PLATFORM_CONFIG=FireSimConfig
+PLATFORM_CONFIG=BaseF1Config
 instancetype=c5.4xlarge
 deploytriplet=None
 
 [firesim-singlecore-no-nic-lbp]
 DESIGN=FireSimNoNIC
 TARGET_CONFIG=FireSimRocketChipSingleCoreConfig
-PLATFORM_CONFIG=FireSimConfig_F160MHz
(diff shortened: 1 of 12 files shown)


deploy/sample-backup-configs/sample_config_build_recipes.ini
```diff
@@ -121,26 +121,33 @@ instancetype=c5.4xlarge
...
-PLATFORM_CONFIG=FireSimDDR3FRFCFSLLC4MBConfig_F90MHz
+TARGET_CONFIG=DDR3FRFCFSLLC4MB_FireSimRocketChipSha3L2Config
+PLATFORM_CONFIG=BaseF1Config_F90MHz
+instancetype=c5.4xlarge
+deploytriplet=None
+
+[firesim-singlecore-sha3-l2-nic-ddr3-llc4mb-120mhz]
+DESIGN=FireSim
+TARGET_CONFIG=DDR3FRFCFSLLC4MB_FireSimRocketChipSha3L2Config
+PLATFORM_CONFIG=BaseF1Config_F120MHz
```
a0u: Should we simply target 120 MHz by default?  The AGFI build took slightly less than 4 hours for me on a z1d.xlarge instance.

PR #365, opened 2019-10-08
[docs] Fixup Config; Golden Gate naming in Debugging, Targets files

davidbiancolin:
Also expand on Config string concatenation and describe how to inspect configs with the Scala REPL.


docs/Advanced-Usage/Generating-Different-Targets.rst
```diff
@@ -101,13 +101,49 @@ compound Config instance. See the example below:
...
+    // Can specify undefined classes this way. ex: TARGET_CONFIG=Override2_Base
 
 With this scheme, you don't need to define a Config class for every instance you
 wish to generate. We use this scheme to specify FPGA frequencies (eg.
-BaseF1Config_F90MHz) in manager build recipes, but it's also very useful for doing
-sweeping over a parameterization space.
+"BaseF1Config_F90MHz") in manager build recipes, but it's also very useful for doing
+sweeping over a parameterization space. 
+
+**Note that the precedence of Configs decreases from left to right in a string**.  Appending a config to an existing one will only have an effect if it
```
alonamid:
```suggestion
**Note that the precedence of Configs decreases from left to right in a string**.  Appending (as opposed to prepending) a config to an existing one will only have an effect if it
```

docs/Advanced-Usage/Generating-Different-Targets.rst
```diff
@@ -101,13 +101,49 @@ compound Config instance. See the example below:
...
+sets a field not already set in higher precendence Configs.
+
+For example, "BaseF1Config_F90MHz" is equivalent to
+"BaseF1Config_F90MHz_F80MHz" as ``DesiredHostFrequency`` resolves to 90 MHz,
+but "F90MHz_BaseF1Config" is distinct "F80MHz_F90MHz_BaseF1Config" in
```
alonamid:
```suggestion
but "F90MHz_BaseF1Config" is distinct from "F80MHz_F90MHz_BaseF1Config" in
```

docs/Advanced-Usage/Generating-Different-Targets.rst
```diff
@@ -101,13 +101,49 @@ compound Config instance. See the example below:
...
+that ``DesiredHostFrequency`` resolves to 90 and 80 MHz respectively.
```
alonamid:
```suggestion
that ``DesiredHostFrequency`` resolves to 90 MHz in the former and 80 MHz in the latter.
```

docs/Advanced-Usage/Generating-Different-Targets.rst
```diff
@@ -67,31 +66,32 @@ Specifying A Target Instance
 To generate a specific instance of a target, the build system leverages four Make variables:
 
 1. ``TARGET_PROJECT``: this points the Makefile (`sim/Makefile`) at the right
-   target-specific Makefrag, which defines the generation and MIDAS-level
+   target-specific Makefrag, which defines the generation and Golden Gate-level
```
alonamid: In other places in the docs you left this as MIDAS-level software simulation. Can we leave this as MIDAS-level, since Golden[-]Gate-Level because really long? (and you've left the term Midas hanging around in places like Midas-examples...)

docs/Advanced-Usage/Generating-Different-Targets.rst
davidbiancolin: I think this is unambiguous enough, given that you've said it in other places, and there's an example below. The docs are too wordy as it is.

docs/Advanced-Usage/Generating-Different-Targets.rst
davidbiancolin: I intended to leave ML, ML for now. I agree GG-level isn't good. 

I didn't change anything that would require a code change. The midasexamples will remain the midasexamples for now.

davidbiancolin:
This is the first of many updates.

PR #371, opened 2019-10-09
[FASED] Change default runtime LLC line length to 64B

davidbiancolin:
[This is not RTL changing]

With the current parameterization (generation-time configuration) of FASED, to get the full 4MB capacity you need to use 128B lines at runtime. I'm seeing anomalous behavior in some benchmarks at this length, and so i'm reducing this to 64B which not only matches the line size of our target machines, but was more heavily validated and also matches the default size we used until recently.

I will generate a plot with CCBench and drop it in here.

To port existing HWDB entries, we just need to search-replace 4MB -> 2MB and everything will just work without needing to rebuild AGFIs.


davidbiancolin:
Wrong base branch.

PR #372, opened 2019-10-09
[Configs] Change FASED default line lengths to 64B

davidbiancolin:
[This is not RTL changing]

With the current parameterization (generation-time configuration) of FASED, to get the full 4MB capacity you need to use 128B lines at runtime. I'm seeing anomalous behavior in some benchmarks at this length, and so i'm reducing this to 64B which not only matches the line size of our target machines, but was more heavily validated and also matches the default size we used until recently.

I will generate a plot with CCBench and drop it in here.

To port existing HWDB entries, we just need to search-replace 4MB -> 2MB and everything will just work without needing to rebuild AGFIs.

### DIFF
--- a/sim/firesim-lib/src/main/scala/configs/FASEDTargetConfigs.scala
+++ b/sim/firesim-lib/src/main/scala/configs/FASEDTargetConfigs.scala
@@ -87,7 +87,7 @@ class LBP32R32W extends Config(
   new WithFuncModelLimits(32,32) ++
   new WithDefaultMemModel)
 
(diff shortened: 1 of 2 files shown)


alonamid:
If the goal is to minimize effort by not rebuilding AGFIs, it's worth considering the effort of updating slideware and documentation with updated HWDB names

davidbiancolin:
For the purposes of minimizing changes to the Demo, these can be merged in after we've generated the AMI. They need to go into the release.

PR #388, opened 2019-10-12
Add support for blackboxes in FAMETransform via clock gating

albert-magyar:
MIDAS PR: ucb-bar/midas#150

### DIFF
--- a/sim/firesim-lib/src/main/scala/passes/AsyncResetReg.scala
+++ b/sim/firesim-lib/src/main/scala/passes/AsyncResetReg.scala
@@ -7,6 +7,7 @@ package passes
 import firrtl._
 import firrtl.ir._
 import firrtl.Mappers._
+import firrtl.transforms._
 
 class AsyncResetReg extends chisel3.experimental.MultiIOModule {
   import chisel3.core._
@@ -18,7 +19,7 @@ class AsyncResetReg extends chisel3.experimental.MultiIOModule {
 }
 
 // Transform RocketChip's async reset to sync reset
-object AsyncResetRegPass extends firrtl.passes.Pass {
+object AsyncResetRegPass extends firrtl.Transform {
   override def name = "[FireSim] Replace AsyncReset Pass"
   override def inputForm = MidForm
   override def outputForm = MidForm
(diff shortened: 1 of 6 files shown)


sim/firesim-lib/src/main/scala/passes/AsyncResetReg.scala
```diff
@@ -54,7 +55,18 @@ object AsyncResetRegPass extends firrtl.passes.Pass {
       case _ => m
     } 
 
-  def run(c: Circuit): Circuit = {
-    c.copy(modules = c.modules map transform)
+  def execute(cs: CircuitState): CircuitState = {
+    val transformedMods = cs.circuit.modules.map(transform)
+
+    // Drop AsyncResetReg BlackBoxSourceHelper annotations
```
davidbiancolin: Is this just to be more complete?

sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -112,5 +112,7 @@ class NarrowPrintfModuleF1Test extends TutorialSuite("NarrowPrintfModule",
   simulationArgs = Seq("+print-no-cycle-prefix", "+print-file=synthprinttest.out")) {
   diffSynthesizedPrints("synthprinttest.out")
 }
+
 // MIDAS 2.0 compiler tests
 class WireInterconnectF1Test extends TutorialSuite("WireInterconnect")
+class VerilogAccumulatorF1Test extends TutorialSuite("VerilogAccumulator")
```
davidbiancolin: Did you intentionally neglect the chisel accumulator?

davidbiancolin:
You should also make sure the docs make it clear this feature exists.

sim/src/test/scala/midasexamples/TutorialSuite.scala
albert-magyar: Yes but I can add it if that makes sense.

sim/firesim-lib/src/main/scala/passes/AsyncResetReg.scala
albert-magyar: Otherwise the AsyncResetReg ends up in the list of blackbox resource files. Since the pass deletes the module, it should delete the annotation, or the blackbox Verilog definition can conflict with the pure-FIRRTL mock synchronous implementation.

davidbiancolin: Got it.

sim/src/test/scala/midasexamples/TutorialSuite.scala
davidbiancolin: You should either add it or remove it all together. I'm always down for less code.

albert-magyar: Removed

PR #394, opened 2019-10-13
Add ground test bridge and bridge driver

zhemao:
This bridge is needed to allow FireSimTraceGen to exit cleanly.

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/GroundTestBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/GroundTestBridge.scala
@@ -1,0 +1,44 @@
+//See LICENSE for license details
+package firesim.bridges
+
+import midas.widgets._
+
+import chisel3._
+import chisel3.util._
+import freechips.rocketchip.config.Parameters
+
+class GroundTestBridge extends BlackBox
+    with Bridge[HostPortIO[GroundTestBridgeTargetIO], GroundTestBridgeModule] {
+  val io = IO(new GroundTestBridgeTargetIO)
+  val bridgeIO = HostPort(io)
+  val constructorArg = None
+  generateAnnotations()
+}
+
+object GroundTestBridge {
+  def apply(success: Bool)(implicit p: Parameters): GroundTestBridge = {
+    val bridge = Module(new GroundTestBridge)
+    bridge.io.success := success
+    bridge
+  }
+}
+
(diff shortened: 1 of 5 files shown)


sim/firesim-lib/src/main/scala/bridges/GroundTestBridge.scala
```diff
@@ -0,0 +1,41 @@
...
+}
+
+class GroundTestBridgeModule(implicit p: Parameters)
+    extends BridgeModule[HostPortIO[GroundTestBridgeTargetIO]] {
+  val io = IO(new WidgetIO)
+  val hPort = IO(HostPort(new GroundTestBridgeTargetIO))
+
+  val tFire = hPort.toHost.hValid && hPort.fromHost.hReady
+  hPort.toHost.hReady := tFire
+  hPort.fromHost.hValid := tFire
```
davidbiancolin: Tie both of these off to true.B

sim/firesim-lib/src/main/scala/bridges/GroundTestBridge.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  genROReg(hPort.hBits.success, "success")
```
davidbiancolin: Add an intermediate reg here that's sticky on success (i.e., set when toHost.fire && hPort.hBits.sucess) and tie toHost.hReady := true.B per previous comment.

sim/firesim-lib/src/main/scala/bridges/GroundTestBridge.scala
davidbiancolin: You don't produce output tokens. We need avoid this pattern of making the channels dependent on one another whenever possible.

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -331,6 +332,49 @@ uint64_t host_mem_offset = -0x80000000LL;
     #endif
 #endif
 
+#ifdef GROUNDTESTBRIDGEMODULE_struct_guard
```
davidbiancolin: Technically you only need this you're going to super-node GroundTest but this is safe.

zhemao:
Okay I updated the token handling and success register. @davidbiancolin Can you take another look?

PR #400, opened 2019-10-18
Flatten the MIDAS submodule into FireSim

davidbiancolin:
Maintaining Golden Gate (MIDAS II) as a separate repo is a considerable development burden and given FireSim's migration towards becoming more a library, and the relative lack of users that just pull in MIDAS, there's little need to keep it in a separate repo. If some users want just the compiler we can look at publishing Golden Gate in the future.

This maintains the commit history for all of MIDAS, which predates the public firesim repo by at least three years. All MIDAS commits have their messages prefixed with [MIDAS]. Checking out one of those commits will take you to a snapshot of the MIDAS repository.


albert-magyar:
David's contributor commit count:

image

albert-magyar:
How does this compare with just leaving the existing commits in the old repo and just deprecating the repo?

davidbiancolin:
The only real cost is the loss of history. Once we start making changes on top of the now flattened submodule the cost of that will become more apparent as it'll be a little bit more annoying to track down issues that may have preceded the flattening. 

Squashing all the history into one commit keeps FireSim's history more sane.

albert-magyar:
Ah, I didn’t get that this was meant to be squashed.

davidbiancolin:
> Ah, I didn’t get that this was meant to be squashed.

It wasn't -- but that's the essentially the alternative right? Just dropping the code in.

PR #405, opened 2019-10-22
Make sure tracegen works in standalone mode

zhemao:


sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -169,6 +169,11 @@ $(OUTPUT_DIR)/%.vpd: $(OUTPUT_DIR)/% $(EMUL)-debug
 AXE_DIR=$(chipyard_dir)/tools/axe/src
 AXE=$(AXE_DIR)/axe
 
+ifdef FIRESIM_STANDALONE
```
davidbiancolin: Where does this recipe live in non-standalone?

PR #415, opened 2019-11-04
QCOW2 disk image support

sagark:
This results in massive savings for root disk sizes in sim. Also includes some misc cleanup for supernode handling. This does not include support for generating qcow2 images in marshal, I just generated them by hand for linux-uniform.

The default linux-uniform .img is 2.0G, the qcow2 version is 40M. Verified that as the simulation writes to disk, the qcow2 size grows.

infrasetup times for linux-uniform:

qcow, fresh inst: 1m45s
img, fresh inst: 2m6s

qcow, primed inst: 50s
img, primed inst: 1m1s

This is now ready for review @davidbiancolin @a0u . 

@NathanTP this is ready to test out with marshal qcow2 support.


build-setup-nolog.sh
```diff
@@ -143,6 +143,10 @@ if wget -T 1 -t 3 -O /dev/null http://169.254.169.254/; then
     sudo pip3 install -r sw/firesim-software/python-requirements.txt
     cat sw/firesim-software/centos-requirements.txt | sudo xargs yum install -y
 
+    # Setup for using qcow2 images
+    cd $RDIR
+    ./scripts/install-nbd-kmod.sh
```
davidbiancolin: can this be put in machine launch?

deploy/runtools/firesim_topology_elements.py
```diff
@@ -353,7 +370,7 @@ def get_sim_start_command(self, slotno):
         Currently hardcoded to 4 nodes.
```
davidbiancolin: This comment seems out-of-date.

deploy/runtools/run_farm.py
```diff
@@ -378,6 +378,34 @@ def fpga_node_xdma(self):
...
+        self.instance_logger("""Setting up remote node for qcow2 disk images.""")
+        with StreamLogger('stdout'), StreamLogger('stderr'):
+            # get qemu-nbd
+            run('sudo yum -y install qemu-img')
+            # copy over kernel module
+            put('../build/nbd.ko', '/home/centos/nbd.ko', mirror_local_mode=True)
+
+    def load_nbd_module(self):
+        """ load the nbd module. """
+        self.unload_nbd_module()
```
davidbiancolin: I think it might be good to capture in comments why we have to unload first>

build-setup-nolog.sh
sagark: I wanted to put this here for two reasons:
1. It could be brittle to start out, so it's easier to get a log for it this way/fix it
2. If something updates the kernel version, it needs to be rebuilt

deploy/runtools/run_farm.py
sagark: done

deploy/runtools/firesim_topology_elements.py
sagark: fixed

PR #418, opened 2019-11-05
Fix Target-Level Simulation When FireSim is Top

davidbiancolin:
[This requires a Chipyard bump or a branch of Chipyard. Will not merge until we have reason to bump.] 

Presently, if you're using FireSim-as-top, you can't cd into chipyard and run simulations out of sims/verilator or sims/vcs. This, alongside ucb-bar/chipyard#328, fixes that.


davidbiancolin:
This was cherry picked into dev.

PR #419, opened 2019-11-06
GG cleanup: Remove Platform Mapping, legacy Compiler object

davidbiancolin:
This is a code cleanup PR. It also fixes a bug in one of the midasexamples tests.

Key Changes: 

- Removes PlatformMapping. 
  - SimulationMapping just directly instantiates PlatformShim
  - Saves an extra firrtl linking step; and removes an extra chisel elaboration step
- Changes `Platform` config to be a lambda that elaborates the desired FPGAShim
   - Fewer LoC; lets users provide their own Shim not defined in golden gate
- Removes old Compiler object
  - Should always use Stage or Phase
- Removes a bunch of legacy Transform arguments and annotations obviated by new Bridge system 
  - e.g., target IO passed to transforms (obviated by Bridge changes)
  - e.g, addedTargetIOAnnotation


sim/midas/src/main/scala/midas/core/SimWrapper.scala
```diff
@@ -246,6 +241,7 @@ class SimWrapper(chAnnos: Seq[FAMEChannelConnectionAnnotation],
   val channelPorts = IO(new SimWrapperChannels(bridgeChAnnos, bridgeAnnos, leafTypeMap))
   val hostReset = IO(Input(Bool()))
   val target = Module(new TargetBox(chAnnos, leafTypeMap))
+  target.suggestName(targetInstName)
```
albert-magyar: Does anything break if this doesn't give it the right name?

davidbiancolin: Yes, but this is no more brittle than it was before.

PR #433, opened 2019-12-02
Diplomatic Bridges & Intelligent FPGA-DRAM Allocation

davidbiancolin:
This is the first of three phases to make FPGATop and the BridgeModules instantiated there diplomatic. Here, i just convert the FPGA-DRAM handling to be diplomatic (DMA and MMIO will follow later). This enables the following enhancements:

1) _any bridge_ can now request FPGA DRAM, not just FASED timing models
2) LoadMem only initializes DRAM that has been allocated.
3) Cacheline-striped target memory systems.


sim/midas/src/main/scala/midas/platform/F1Shim.scala
```diff
@@ -141,99 +133,85 @@ class F1Shim(implicit p: Parameters) extends PlatformShim {
...
         io.dma.r.bits.data,
         io.dma.r.bits.last,
-        io.dma.r.bits.id,
-        io.dma.r.bits.user)
+        io.dma.r.bits.id)
     }
 
-  when (io.slave(0).aw.fire()) {
-    printf("[slave,awfire,%x] addr %x, len %x, size %x, burst %x, lock %x, cache %x, prot %x, qos %x, region %x, id %x, user %x\n",
+  when (io_slave(0).aw.fire()) {
```
zhemao, line 148: Wait, does it actually work to reference io_slave before it is created? I usually get a null pointer exception if I do this. Check what happens if AXIDebugPrint is set to true.

sim/midas/src/main/scala/midas/platform/ZynqShim.scala
```diff
@@ -7,23 +7,26 @@ import chisel3._
...
 import freechips.rocketchip.config.{Parameters, Field}
+import freechips.rocketchip.diplomacy.{LazyModule}
 import freechips.rocketchip.util.ParameterizedBundle
 
-import midas.core.{HostMemChannelNastiKey, ChannelWidth}
+import midas.core.ChannelWidth
 
-abstract class PlatformShim(implicit p: Parameters) extends Module {
-  def top: midas.core.FPGATop
+abstract class PlatformShim(implicit p: Parameters) extends MultiIOModule {
```
zhemao, line 15: I feel like this class should be in a separate Scala file.

sim/midas/src/main/scala/midas/platform/ZynqShim.scala
```diff
@@ -7,23 +7,26 @@ import chisel3._
...
+  def top: midas.core.FPGATopImp
   def headerConsts: Seq[(String, Long)]
   def genHeader(sb: StringBuilder, target: String) {
     import widgets.CppGenerationUtils._
+    //We need to split out a C++ header from the C header or VCS croaks
+    //sb.append("#include <cstdint>\n")
```
zhemao, line 21: Does this comment need to be here anymore?

sim/midas/src/main/scala/midas/widgets/Widget.scala
```diff
@@ -31,27 +32,37 @@ object WidgetMMIO {
...
-  private def setWidgetName(n: String) {
-    wName = Some(n)
-  }
-  def getWName: String = {
-    wName.getOrElse(throw new  RuntimeException("Must build widgets with their companion object"))
-  }
+  def printCRs = module.crRegistry.printCRs
+}
+
+abstract class WidgetImp(wrapper: Widget)(implicit p: Parameters) extends LazyModuleImp(wrapper) {
```
zhemao, line 59: Is the implicit parameters argument necessary here? Usually module implementations don't declare their own implicit parameters object, since they will use the one from the parent lazy module.

sim/midas/src/main/scala/midas/widgets/CppGeneration.scala
```diff
@@ -24,9 +24,15 @@ case class UInt32(value: BigInt) extends IntLikeLiteral {
 }
 
 case class UInt64(value: BigInt) extends IntLikeLiteral {
-  def typeString = "uint64_t"
+  def typeString = "unsigned long long"
```
zhemao, line 27: What's the reason for changing this?

sim/midas/src/main/scala/junctions/Nasti2AXI4.scala
```diff
@@ -0,0 +1,36 @@
...
+package junctions
+
+import chisel3._
+import chisel3.MultiIOModule
+
+import freechips.rocketchip.amba.axi4._
+import freechips.rocketchip.config.{Parameters}
+
+class AXI42NastiIdentityModule(params: AXI4BundleParameters)(implicit p: Parameters) extends MultiIOModule {
+  val axi4 = IO(Flipped(new AXI4Bundle(params)))
```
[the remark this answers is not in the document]
davidbiancolin: Nothing in particular, i like not having to type `.io`. I can change it if your prefer.

sim/midas/src/main/scala/midas/core/FPGATop.scala
```diff
@@ -4,87 +4,162 @@ package midas
...
+  memAXI4Node :*= AXI4Buffer() :*= xbar
+  xbar := loadMem.toHostMemory
+  val targetMemoryRegions = sortedRegionTuples.zip(dramOffsets).map({ case ((bridgeSeq, addresses), hostBaseAddr) =>
+    val regionName = bridgeSeq.head.memoryRegionName
+    val virtualBaseAddr = addresses.map(_.base).min
+    val offset = hostBaseAddr - virtualBaseAddr
+    for (bridge <- bridgeSeq) {
+      (xbar := AXI4Buffer()
+            // Implements the base-and-bounds address translation
+            := AXI4AddressTranslation(offset, addresses, regionName)
```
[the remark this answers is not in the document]
davidbiancolin: Good point. I need more :=* in my life.

sim/midas/src/main/scala/midas/platform/F1Shim.scala
davidbiancolin: Ugh, probably not. i was just blindly porting this.

sim/midas/src/main/scala/midas/platform/ZynqShim.scala
davidbiancolin: Agreed.

sim/midas/src/main/scala/midas/platform/ZynqShim.scala
davidbiancolin: This is related to your question about the `unsigned long long` change. As it turns out, we were never emitting `uint64_t`, if we had been we would have run into compilation errors in VCS ML sims. Since we include this header in basically every compilation unit (= bug), VCS 
compiles it part of C and C++ files and complains about the library we use to include the fixed-width types.

Obviously, we should emit fixed-width types. The easiest way to do this IMO is split the header files, into a const.h that is included in all of the places the legacy midas code wants it, and a const.hpp which just includes all the bridge good stuff we use.

sim/midas/src/main/scala/midas/widgets/Widget.scala
davidbiancolin: Good point, will do.

sim/midas/src/main/scala/midas/widgets/CppGeneration.scala
davidbiancolin: See above.

sim/midas/src/main/scala/junctions/Nasti2AXI4.scala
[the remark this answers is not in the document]
zhemao: I think that would be best, if only to not have to use a MultiIOModule.

sim/midas/src/main/scala/midas/platform/ZynqShim.scala
zhemao: I'm pretty sure if you use `#include <stdint.h>` it should work in C or C++.

sim/midas/src/main/scala/midas/platform/F1Shim.scala
```diff
@@ -5,240 +5,38 @@ import chisel3._
...
-      )
-  }
+    top.module.io.ctrl <> io.master
+    top.module.io.dma  <> io.dma
+    io_slave.zip(top.module.mem).foreach({ case (io, bundle) => io <> bundle })
 
-  when (io.slave(0).r.fire()) {
-    printf("[slave(0),rfire,%x] resp %x, data %x, last %x, id %x, user %x\n",
-      cyclecount,
+    val (wCounterValue, wCounterWrap) = Counter(io.master.aw.fire(), 4097)
```
zhemao, line 36: I suppose this is old, but why exactly is it 4097 and not 4096?

davidbiancolin: @sagark

sagark: uhhhhh. probably just padding to avoid having to think about wraparound

sagark: wait nvm

PR #434, opened 2019-12-02
Adjust BlockDevBridge to match new Option-type parameters

jerryz123:
Companion to https://github.com/ucb-bar/testchipip/pull/58 and https://github.com/ucb-bar/chipyard/pull/347


davidbiancolin:
FYI, this is going to require a chipyard bump thus an AGFI regeneration to not break FireSim-as-top.

jerryz123:
> FYI, this is going to require a chipyard bump thus an AGFI regeneration to not break FireSim-as-top.

Can we just roll this into the upcoming FireSIm/Chipyard release bump next week?

davidbiancolin:
> Can we just roll this into the upcoming FireSIm/Chipyard release bump next week?

That's probably the best approach.

PR #435, opened 2019-12-03
Make block device log to file like niclog does

zhemao:


sim/firesim-lib/src/main/cc/bridges/blockdev.cc
```diff
@@ -17,6 +17,7 @@
 /* Uncomment to get DEBUG printing
```
davidbiancolin: Remove these comments.

deploy/runtools/runtime_config.py
```diff
@@ -122,7 +122,7 @@ def array_to_plusargs(valuesarr, plusarg):
         command_bootbinaries = array_to_plusargs(all_bootbinaries, "+prog")
         zero_out_dram = "+zero-out-dram"
 
-        basecommand = """screen -S fsim{slotid} -d -m bash -c "script -f -c 'stty intr ^] && sudo ./{driver} +permissive $(sed \':a;N;$!ba;s/\\n/ /g\' {runtimeconf}) +slotid={slotid} +profile-interval={profile_interval} {zero_out_dram} {command_macs} {command_rootfses} +niclog0=niclog {tracefile} +trace-start0={trace_start} +trace-end0={trace_end} {command_linklatencies} {command_netbws}  {command_shmemportnames} +permissive-off {command_bootbinaries} && stty intr ^c' uartlog"; sleep 1""".format(
+        basecommand = """screen -S fsim{slotid} -d -m bash -c "script -f -c 'stty intr ^] && sudo ./{driver} +permissive $(sed \':a;N;$!ba;s/\\n/ /g\' {runtimeconf}) +slotid={slotid} +profile-interval={profile_interval} {zero_out_dram} {command_macs} {command_rootfses} +niclog0=niclog +blkdev-log0=blkdev-log {tracefile} +trace-start0={trace_start} +trace-end0={trace_end} {command_linklatencies} {command_netbws}  {command_shmemportnames} +permissive-off {command_bootbinaries} && stty intr ^c' uartlog"; sleep 1""".format(
```
davidbiancolin: Does this not always generate the blockdev log?

sim/firesim-lib/src/main/cc/bridges/blockdev.cc
zhemao: We want this flag commented out by default so that it doesn't affect simulation performance. It basically functions like the TOKENVERIFY flag in the NIC bridge.

deploy/runtools/runtime_config.py
zhemao: See above comment. The logging code isn't compiled in by default.

davidbiancolin:
@zhemao clean up after yourself!

PR #444, opened 2019-12-11
make sure SCALA_SOURCES set properly in standalone mode

zhemao:
This redundancy with the chipyard makefrag is kind of annoying, but I don't really see a better way of doing this.

### DIFF
--- a/sim/src/main/makefrag/firesim/Makefrag
+++ b/sim/src/main/makefrag/firesim/Makefrag
@@ -34,6 +34,10 @@ common_chisel_args = $(patsubst $(base_dir)/%,%,$(GENERATED_DIR)) $(DESIGN_PACKA
 
 ifdef FIRESIM_STANDALONE
 	firesim_sbt_project := {file:${chipyard_dir}}firechip
+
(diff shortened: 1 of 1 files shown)


sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -34,6 +34,10 @@ common_chisel_args = $(patsubst $(base_dir)/%,%,$(GENERATED_DIR)) $(DESIGN_PACKA
 
 ifdef FIRESIM_STANDALONE
 	firesim_sbt_project := {file:${firesim_base_dir}/target-rtl/chipyard/}firechip
+
+	lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.scala" -print 2> /dev/null)
```
davidbiancolin: So there's no way to reuse this from Chipyard?

zhemao: Not easily. I guess we could make a separate makefrag in Chipyard that we can source without worrying about polluting environment variables.

Issue #410, opened 2019-11-01
fatal error at firesim launchrunfarm 


CristinaZN:
hi, here is a problem.please help me
i create instance c4 and f1 both in us-east-1a

[centos@ip-192-168-0-187 deploy]$ firesim launchrunfarm
FireSim Manager. Docs: http://docs.fires.im
Running: launchrunfarm

An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 8 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
This probably means there was no more capacity in this availability zone. Try the next one.
An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 8 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
This probably means there was no more capacity in this availability zone. Try the next one.
An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 8 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
This probably means there was no more capacity in this availability zone. Try the next one.
An error occurred (Unsupported) when calling the RunInstances operation: Your requested instance type (f1.2xlarge) is not supported in your requested Availability Zone (us-east-1f). Please retry your request by not specifying an Availability Zone or choosing us-east-1a, us-east-1b, us-east-1c, us-east-1d, us-east-1e.
This probably means there was no more capacity in this availability zone. Try the next one.
An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 8 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
This probably means there was no more capacity in this availability zone. Try the next one.
An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 8 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
This probably means there was no more capacity in this availability zone. Try the next one.
[...]

PR #448, opened 2019-12-23
Update first time limit increase instructions for new vCPU based limits on EC2

sagark:


sagark:
Waiting for confirmation that these are correct from #410

sagark:
These worked as expected on my own account, ready for review.

PR #456, opened 2020-01-16
Fix Copy Results Failure

### LINKED ISSUE #329: Copy results files fail
Occasionally, the runfarm instance will fail to copy files back to the manager. Manually re-running the commands works. See [here](https://groups.google.com/forum/#!topic/firesim/GI29vu-93es) for the initial report.

Here is a trace provided by hzhuo2@wisc.edu:
2019-08-01 21:23:10,886 [instance_log] [INFO ]  [192.168.1.43] Slot 0 completed! copying results.
2019-08-01 21:23:10,886 [flush       ] [DEBUG]  [localhost] local: mkdir -p /home/centos/firesim/deploy/results-workload/2019-08-01--21-04-56-cus-fed29//cus-fed29-test/
2019-08-01 21:23:10,893 [copy_back_jo] [DEBUG]  [localhost] 
2019-08-01 21:23:10,893 [copy_back_jo] [DEBUG]  [localhost] 
2019-08-01 21:23:10,894 [flush       ] [DEBUG]  [192.168.1.43] run: sudo mkdir -p /home/centos/sim_slot_0/mountpoint
2019-08-01 21:23:11,063 [flush       ] [DEBUG]  [192.168.1.43] run: sudo mount /home/centos/sim_slot_0/cus-fed29-test.img /home/centos/sim_slot_0/mountpoint
2019-08-01 21:23:11,600 [flush       ] [DEBUG]  [192.168.1.43] run: sudo chmod -Rf 777 /home/centos/sim_slot_0/mountpoint
2019-08-01 21:23:13,551 [flush       ] [DEBUG]  Fatal error: run() received nonzero return code 1 while executing!
2019-08-01 21:23:13,551 [flush       ] [DEBUG]  Requested: sudo chmod -Rf 777 /home/centos/sim_slot_0/mountpoint
2019-08-01 21:23:13,551 [flush       ] [DEBUG]  Executed: /bin/bash -l -c "sudo chmod -Rf 777 /home/centos/sim_slot_0/mountpoint"
2019-08-01 21:23:13,551 [flush       ] [DEBUG]  Aborting.
2019-08-01 21:23:13,573 [<module>    ] [ERROR]  Fatal error.

abejgonzalez:
Fixes #329 by removing the `immutable` attribute on the `nfs` file that was causing problems. This PR also changes the `chmod` to `chown` based on what @a0u recommends. 

This was tested in Chipyard with Fedora boot (`fedora-base.json`).

### DIFF
--- a/deploy/runtools/firesim_topology_elements.py
+++ b/deploy/runtools/firesim_topology_elements.py
@@ -241,7 +241,7 @@ class FireSimServerNode(FireSimNode):
         runcommand = self.server_hardware_config.get_boot_simulation_command(
             slotno, all_macs, all_rootfses, all_linklatencies, all_maxbws,
             self.server_profile_interval, all_bootbins, self.trace_enable,
-            self.trace_select, self.trace_start, self.trace_end, 
+            self.trace_select, self.trace_start, self.trace_end,
             self.autocounter_readrate, all_shmemportnames)
 
         run(runcommand)
@@ -279,7 +279,8 @@ class FireSimServerNode(FireSimNode):
                     rfsname = """/home/centos/sim_slot_{}/{}""".format(simserverindex, rfsname)
 
                 run("""sudo mount {blockfile} {mntpt}""".format(blockfile=rfsname, mntpt=mountpoint))
-                run("""sudo chmod -Rf 777 {}""".format(mountpoint))
+                run("""sudo chattr -i {}/etc/sysconfig/nfs""".format(mountpoint))
+                run("""sudo chown -R centos {}""".format(mountpoint))
 
             ## copy back files from inside the rootfs
             with warn_only(), StreamLogger('stdout'), StreamLogger('stderr'):
--- a/deploy/runtools/firesim_topology_elements.py
+++ b/deploy/runtools/firesim_topology_elements.py
@@ -279,7 +279,7 @@ class FireSimServerNode(FireSimNode):
                     rfsname = """/home/centos/sim_slot_{}/{}""".format(simserverindex, rfsname)
 
                 run("""sudo mount {blockfile} {mntpt}""".format(blockfile=rfsname, mntpt=mountpoint))
-                run("""sudo chattr -i {}/etc/sysconfig/nfs""".format(mountpoint))
+                run("""sudo chattr -R -i {}""".format(mountpoint))
                 run("""sudo chown -R centos {}""".format(mountpoint))
 
             ## copy back files from inside the rootfs


deploy/runtools/firesim_topology_elements.py
```diff
@@ -279,7 +279,8 @@ def copy_back_job_results_from_run(self, slotno):
                     rfsname = """/home/centos/sim_slot_{}/{}""".format(simserverindex, rfsname)
 
                 run("""sudo mount {blockfile} {mntpt}""".format(blockfile=rfsname, mntpt=mountpoint))
-                run("""sudo chmod -Rf 777 {}""".format(mountpoint))
+                run("""sudo chattr -i {}/etc/sysconfig/nfs""".format(mountpoint))
```
a0u: There may be other files with the immutable bit.

```suggestion
                run("""sudo chattr -R -i {}""".format(mountpoint))
```

deploy/runtools/firesim_topology_elements.py
```diff
@@ -279,7 +279,8 @@ def copy_back_job_results_from_run(self, slotno):
...
+                run("""sudo chown -R centos {}""".format(mountpoint))
```
a0u: Information about the remote user should be factored out into a variable instead of hardcoded as `centos`, although I see the assumptions about the username and home directory being endemic to this file (so the issue is broader than this PR).

abejgonzalez:
> One problem with this approach is that it modifies the whole root filesystem. This would probably make the image unbootable (since certain things need to be owned by root). I feel like people might try to copy the image out of the worker for debugging and find it all messed up.
> 
> An alternative is to copy the files to a tmp directory on the runfarm instance first, and chown it there. It might look something like this (totally untested):
> 
> ```
>             ## copy back files from inside the rootfs
>             outputsDir = """/home/centos/sim_slot_{}/outputsDir""".format(simserverindex)
>             run("""sudo mkdir -p {}""".format(outputsDir))
>             with warn_only(), StreamLogger('stdout'), StreamLogger('stderr'):
>                 for outputfile in jobinfo.outputs:
>                     run("""sudo cp -r {src} {dst}""".format(src=mountpoint + outputfile, dst=outputsDir + outputfile))
>                 run("""sudo chattr -R -i {}""".format(outputsDir))
>                 run("""sudo chown -R centos {}""".format(outputsDir))
>                 get(remote_path=outputsDir, local_path=job_dir)
> ```

What was the initial concern with `chmod`? Mainly that it was inelegant? I assume that `chmod` approach preserves the fs.

As for the copy approach, do we really want to have duplicate files on the runfarm... what happens if they are very large (and there is multiple of them)? Maybe I'm just thinking of the worst case.

a0u:
> What was the initial concern with chmod? Mainly that it was inelegant? I assume that chmod approach preserves the fs.

`chmod` and `chown` modify the metadata destructively on the filesystem itself.  Aside from postmortem debugging scenario raised by @NathanTP, usually this is not a concern since the rootfs image is presumably disposable since the next `infrasetup` should restore it to a pristine state.

> As for the copy approach, do we really want to have duplicate files on the runfarm... what happens if they are very large (and there is multiple of them)? Maybe I'm just thinking of the worst case.

Instead of copying, using an [overlayfs](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt) with a tmpfs upper layer would avoid duplication of files when using the "metadata only copy up" (`metacopy`) mount option.  But this entails a slightly more complicated procedure.

abejgonzalez:
I guess the question is... how often do we think people will copy the image and try to debug it? Has this come up a lot in the past? If not, I think we can just add a comment/todo somewhere and prioritize it for later when someone asks for it.

NathanTP:
I mean, it's not a huge deal. I've done it before when stuff isn't booting on FireSim, I try to bring the image back and boot it in Marshal. I'll approve the current approach as well, it just seemed a little cleaner this way and I doubt we're going to have such big files in our guests that the extra disk space is an issue.

deploy/runtools/firesim_topology_elements.py
abejgonzalez: Agreed. But as you said, probably requires a larger PR. Ill go ahead and file an issue.

abejgonzalez:
> I mean, it's not a huge deal. I've done it before when stuff isn't booting on FireSim, I try to bring the image back and boot it in Marshal. I'll approve the current approach as well, it just seemed a little cleaner this way and I doubt we're going to have such big files in our guests that the extra disk space is an issue.

Im not, opposed to reversing to `chmod` (assuming that still allows you to boot from the image later on - correct me if Im wrong @NathanTP).

NathanTP:
That wouldn't work because we don't know what the settings were before. It also doesn't solve the issue of some other weird file being created in the future that we didn't account for.  Another option would be to run the chmod/chown only on the files we're copying, right before we copy them (moving the chmod/chown into the for loop). That way the use is more in direct control of what's happening (only files they specify will be copied). I think that's probably a best of both worlds here.

abejgonzalez:
Offline conversation. I will try to put the `chmod`/`chown` to just the files that are getting copied off. (may not work since you might need to change the permissions recursively up the fs)

abejgonzalez:
@NathanTP @sagark Selectively `chattr`ing and `chmod/chown`ing the files does not work. You get the following error:

```
Warning: get() encountered an exception while downloading 'file'
Underlying exception:
    Permission denied
```

abejgonzalez:
For now, Im going to merge this PR. However, I'll create an issue stating that there needs to be a more elegant solution.

PR #459, opened 2020-01-21
Fix "Fix Copy Results Failure"

abejgonzalez:
Ok I trusted that the `-R` flag would work on `chattr` and that is not true. Even if you add the `-f` to it, it still gives an error code of `1` despite it working afterward (similar to `chmod`ing then it working afterwards). So #456 needs to be patched. 

Here are two options for a quick fix (which I think should go into the next Chipyard bump).

1. Just change the file attributes on the faulty file (`nfs`)
2. Ignore the error code 1 and continue with the copying.

IMO # 1 is better since we know that we are going to fix this later down the line. Thoughts?

### DIFF
--- a/deploy/runtools/firesim_topology_elements.py
+++ b/deploy/runtools/firesim_topology_elements.py
@@ -279,7 +279,7 @@ class FireSimServerNode(FireSimNode):
                     rfsname = """/home/centos/sim_slot_{}/{}""".format(simserverindex, rfsname)
 
                 run("""sudo mount {blockfile} {mntpt}""".format(blockfile=rfsname, mntpt=mountpoint))
-                run("""sudo chattr -R -i {}""".format(mountpoint))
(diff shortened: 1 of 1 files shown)


abejgonzalez:
For testing, I have booted this on BOOM with the `outputs.json` file modified to work on Fedora.

abejgonzalez:
This should go in before the `dev` -> `master` merge. Are there any specific tests that people want me to run?

NathanTP:
wow, fabric sucks. I just spent an hour goofing around with various solutions and none of them work. Why scp needs to CD into a directory to copy its contents is beyond me. The 'use_sudo' option of fabric also doesn't work because it uses the shell's sudo and not /usr/bin/sudo and on centos that seems to be lacking needed features. I'm out of ideas other than manually copying files to a tmp directory before copying back to the manager.

If people don't want to do that, then this approach will suffice for now to just get it working (but it's really fragile and will probably break eventually). I'll probably remove the NFS stuff from fedora eventually.

PR #461, opened 2020-01-22
[makefrags] Remove an errant info statement in firesim's makefrag

davidbiancolin:
Also be more forthright about DRAM initialization time.

### DIFF
--- a/sim/src/main/cc/firesim/firesim_top.cc
+++ b/sim/src/main/cc/firesim/firesim_top.cc
@@ -528,7 +528,7 @@ void firesim_top_t::run() {
     }
 
     if (do_zero_out_dram) {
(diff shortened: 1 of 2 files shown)


sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -528,7 +528,7 @@ void firesim_top_t::run() {
     }
 
     if (do_zero_out_dram) {
-        fprintf(stderr, "Zeroing out FPGA DRAM. This will take a few seconds...\n");
+        fprintf(stderr, "Zeroing out FPGA DRAM. This will take a few minutes...\n");
```
sagark: classic

PR #464, opened 2020-01-22
[manager] Change default build instances to z1d.2xlarge

davidbiancolin:
These are marginally more expensive (0.744 vs 0.68 / hour on-demand) over c5.4xlarges but run significantly faster and have more memory (64 vs 32 GiB). 

The memory increase removes build-failures due to Vivado OOM issues with large designs.


sagark:
Can you update the build docs references to instance type too? e.g. https://github.com/firesim/firesim/blob/7f36976cde51309834851ff4f246a2fe007ce324/docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst

davidbiancolin:
> Can you update the build docs references to instance type too? e.g. https://github.com/firesim/firesim/blob/7f36976cde51309834851ff4f246a2fe007ce324/docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst

Done

PR #475, opened 2020-01-25
[makefrag] Parallelize verilator-based ML-simulator compilation

davidbiancolin:
In general, this provides at least a 2x speedup when compiling Midas-level simulators. 
Re: @farzadfch comments here: https://groups.google.com/forum/#!topic/firesim/RiU_WKZlZ7E. 

I just copied what we were doing in chipyard. (h/t @jerryz123)

### DIFF
--- a/sim/midas/src/main/cc/rtlsim/Makefrag-verilator
+++ b/sim/midas/src/main/cc/rtlsim/Makefrag-verilator
@@ -16,6 +16,8 @@
 
 VERILATOR ?= verilator --cc --exe
 override VERILATOR_FLAGS := --assert -Wno-STMTDLY -O3 \
+	--output-split 10000 \
+	--output-split-cfuncs 100 \
 	-CFLAGS "$(CXXFLAGS) $(CFLAGS)" \
 	-LDFLAGS "$(LDFLAGS) " \
 	$(VERILATOR_FLAGS)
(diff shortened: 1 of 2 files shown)


farzadfch:
Just wanted to point out that the speedup is dependant on the number of c files generated. The number of files should be roughly equal to the number of parallel jobs. Otherwise, if number of files is much larger, we may actually get slowdown.

For my simulations, I manually optimize `--output-split` and `--output-split-cfuncs` based the size of my design and the type of instance that I am running on. With a quad-core small BOOM running on a 48-thread instance (z1d.12xlarge), `--output-split 50000 ` and `--output-split-cfuncs 500 ` delivers the best speedup (about 5x).

davidbiancolin:
I copied this from chipyard without doing a parameter sweep.  Given that the hosts we end up running this on vary pretty wildly, i think airing on the side of more files is probably the right call -- does penalty grow rapidly for # files > # threads? 

I'd consider is optimizing for the default manager instance type we recommend.

farzadfch:
For a quad-core small BOOM, it is about 3x slower with `--output-split 10000` compared to w/o spliting. Here are the times that measured on c4.4xlarge:

  | Quad | Single
-- | -- | --
w/o split | 9m31s | 2m19s
output-split 10000 output-split-cfuncs 100 | 30m18s | 1m53s
output-split 200000 output-split-cfuncs 2000 | 5m20s | 1m17

The command that I run: `time make verilator-debug -j16`. These do not include the Chisel compilation and elaboration (only verilator build and gcc compilation).

Of course, the smartest thing would be to change the split parameters based on the design size, but if that's not possible, I think this should be optimized for larger designs as they take longer to build.

PR #477, opened 2020-01-25
[TracerV] Fix the reported target cycle

davidbiancolin:
Since the trace pulled off the BridgeModule can begin at any target cycle, each token includes a target-cycle field. The counter being used to fill this field was being incremented when the Bridge was back-pressuring the simulator leading to erroneously large target cycles being reported in the trace and inaccurate cycle-based triggering. 

This PR fixes that and cleans up the token handling. 

Ideally this would go in the release, but that would require an AGFI regeneration. We could also ship it so that any user-built AGFIs would have the correct behavior.


sagark:
i'll regen agfis

PR #485, opened 2020-01-30
[synthPrintf] Use topWiringPrefix to prevent name collisions

davidbiancolin:
This works around a bug in TopWiring to unbreak synthesizedPrintfs in the top-level of target module hierarchy.

In synthesizePrintf, annotated printfs are translated to an aggregate wire before calling topWiring on that wire. Since it wasn't being given a prefix, topWiring was trying to create ports using the same name as the aggregate: these ports instead are mapped to <name>_0 to disambiguate them. TopWiring doesn't properly report this in it's mappings causing print synthesis to break when it attempts to look up the generated port.

It also cleans up prefix handling in the BridgeModule.


davidbiancolin:
I thought you might say that. :P

PR #498, opened 2020-02-11
Update build-setup script for Chipyard as top

abejgonzalez:
When Chipyard is top, this script breaks because the script is not able to enter the Chipyard directory from `target-design`. This makes it so it uses the prior relative path specified.

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -156,7 +156,7 @@ if wget -T 1 -t 3 -O /dev/null http://169.254.169.254/; then
     make
 
     # Install firesim-software dependencies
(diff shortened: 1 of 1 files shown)


davidbiancolin:
Have we bumped chipyard in firesim to a point such that FireMarshal has been promoted / does this work with FireSim as top?

abejgonzalez:
I tried this on the most recent `dev` version of Chipyard that included both the bumped FireSim and FireMarshal changes. I have not tried doing this on FireSim-as-top. When I get the chance I will try that and update here.

sagark:
the best kind of regression

sagark:
after discussion, we're going to address this differently. closing this.

PR #499, opened 2020-02-12
Deterministically Initialize Unreset Target-State

davidbiancolin:
This PR fixes a non-determinism bug in FPGA simulations where target registers that are not reset in the target design would appear to take on different values between simulation runs. To get deterministic simulation results, FireSim relies upon target-registers and memories (perhaps implemented as BRAMs) taking on a fixed value after FPGA programming (in most cases, 0). 

While this is generally the case, there are circumstances under which this assumption fails to hold. This is detailed in this [Xilinx AR](https://www.xilinx.com/support/answers/44174.html). 

The solution is to keep the target clock-gated until FPGA initialization is complete, which i've implemented using the direct-clock-gating scheme we use in the F1 transform to support verilog black boxes.

As part of this PR, i've also:
- simplified the reset sequence of the simulation to remove the notion of a SimReset
- fixed a bug where ram optimizations failed to work under a direct clock gating scheme
- cherry-picked a bugfix from the multiclock branch to fix verilator simulation of our clock gating

### DIFF
--- a/sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala
+++ b/sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala
@@ -226,10 +226,13 @@ class EmitAndWrapRAMModels extends Transform {
       val readWritePorts = annos.collect({ case anno: ModelReadWritePort  => new ReadWritePort(anno, mod.ports)})
       require(readWritePorts.isEmpty)
 
-      val clocks = mod.ports.filter(_.tpe == ClockType)
-      // TODO: turn this back to == 1
-      assert(clocks.length >= 1)
-      val hostClock = clocks.find(_.name == "clock").getOrElse(clocks.head) // TODO: naming convention for host clock
+      val (hostClock, updatedPortList) = mod.ports.find(p => p.tpe == ClockType && p.name == "clock") match {
+        case Some(port) => (port, mod.ports)
+        case None =>
+          val cPort = Port(NoInfo, "clock", Input, ClockType)
+          (cPort, cPort +: mod.ports)
+      }
+
       val hostReset = mod.ports.find(_.name == "hostReset").get
 
       val name = ns.newName("RamModel")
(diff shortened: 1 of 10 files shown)


sim/midas/src/main/verilog/BUFGCE.v
```diff
@@ -3,10 +3,22 @@ module BUFGCE(
...
-  /* verilator lint_on COMBDLY */
+   reg       enable_latch;
+   always @(posedge I)
+     enable_latch <= CE;
+`ifdef VERILATOR
+   // Note: Verilator doesn't like procedural clock gates
+   // They cause combinational loop errors and UNOPT_FLAT
+   assign O = (I & enable_latch);
+`else
+   // VCS doesn't like the Verilator clock gate
```
albert-magyar:
```suggestion
   // Note: VCS doesn't like the Verilator clock gate
```

sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala
```diff
@@ -226,18 +226,21 @@ class EmitAndWrapRAMModels extends Transform {
       val readWritePorts = annos.collect({ case anno: ModelReadWritePort  => new ReadWritePort(anno, mod.ports)})
       require(readWritePorts.isEmpty)
 
-      val clocks = mod.ports.filter(_.tpe == ClockType)
-      // TODO: turn this back to == 1
-      assert(clocks.length >= 1)
-      val hostClock = clocks.find(_.name == "clock").getOrElse(clocks.head) // TODO: naming convention for host clock
+      val (hostClock, updatedPortList) = mod.ports.find(p => p.tpe == ClockType && p.name == "clock") match {
+        case Some(port) => (port, mod.ports)
+        case None =>
```
albert-magyar: So this is just checking if the extracted model (not the generated RAM model) has a port named `clock`? And doesn't `hostClock` contain the same port in either case?

PR #503, opened 2020-02-13
Warn when System Commit Bandwidth Exceeds TracerV Capacity

davidbiancolin:
I'm not sure this the right approach, but this lets larger systems pass through compilation without needing to remove the TracerV bridge.

Also add a colored logging layer around FIRRTL's logger for use in Golden Gate.

TODO:
- [ ] Remove LazyLogging mixin to Widget

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -216,11 +216,18 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters) extends Bridg
   lazy val toHostCPUQueueDepth  = TOKEN_QUEUE_DEPTH
   lazy val dmaSize = BigInt((BIG_TOKEN_WIDTH / 8) * TOKEN_QUEUE_DEPTH)
 
-  val uint_traces = traces map (trace => Cat(trace.valid, trace.iaddr).pad(64))
-  outgoingPCISdat.io.enq.bits := Cat(Cat(trace_cycle_counter, 
-                                         0.U((outgoingPCISdat.io.enq.bits.getWidth - Cat(uint_traces).getWidth - trace_cycle_counter.getWidth).W)),
-                                     Cat(uint_traces))
-
+  // This is a temporary workaround for large multi-boom systems. Multiclock
+  // instantiates one tracerV instance per core
+  val perInsnBits = 64
+  val uint_traces = traces.map(trace => Cat(trace.valid, trace.iaddr).pad(perInsnBits))
+  val availableTokenBits = outgoingPCISdat.io.enq.bits.getWidth - trace_cycle_counter.getWidth
+  val insnsPerPayload = availableTokenBits / perInsnBits
+  if (insnsPerPayload != uint_traces.size) {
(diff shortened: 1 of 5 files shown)


sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -216,11 +216,18 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters) extends Bridg
...
-                                         0.U((outgoingPCISdat.io.enq.bits.getWidth - Cat(uint_traces).getWidth - trace_cycle_counter.getWidth).W)),
-                                     Cat(uint_traces))
-
+  // This is a temporary workaround for large multi-boom systems. Multiclock
+  // instantiates one tracerV instance per core
+  val perInsnBits = 64
+  val uint_traces = traces.map(trace => Cat(trace.valid, trace.iaddr).pad(perInsnBits))
+  val availableTokenBits = outgoingPCISdat.io.enq.bits.getWidth - trace_cycle_counter.getWidth
+  val insnsPerPayload = availableTokenBits / perInsnBits
+  if (insnsPerPayload != uint_traces.size) {
```
alonamid: but what if you have less?
For example: 
uint_traces.size = 1 (one core, in order, one commit every cycle)
insnsPerPayload = (512 - 64) / 64 = 7
This should still be fine, but with the `!=` you will get the warning.
Should this just be a less-or-equal-than?

davidbiancolin: Yeah this wrong.

davidbiancolin: That aside, do you think this should be an error?

alonamid: I don't think it should be an error (if people don't care about the trace, this shouldn't break their build). 
However, I do acknowledge the fact the I tend to ignore all the warnings that pop up during firrtl compilation. I think a potential solution he would be to have it as a warning, but also output a text warning as the first line of the output TRACEFILE written by the TracerV bridge driver. This way, when someone tries to ingest the binary trace, they will get a processing error (because there will be a bad first line), and might open the text file to see the text warning. And if someone tries to read the human-readable version, you will be able to see the warning in the first line. That way the warning reaches it's intended audience (trace users), but doesn't break a build for other people.

davidbiancolin: Yeah i don't feel like doing that much work considering that in multiclock we're going to have one of these per core. You'd need to have an 8-wide machine to break it in that case. So there's two options, as far as i see it.

1) We can leave this as is. I fix the logic.
2) I promote this to an error. At least then the user knows to remove the TracerV bridge instead of a getting an inscrutable chisel error (It was @farzadfch that initially reported this.)

At this point, if we build a target without TracerV do things even work?

sagark: imo 1. is fine

davidbiancolin:
Multiclock makes it much harder to exercise this, I'll revisit it if people start having issues with it in the future.

PR #510, opened 2020-02-19
document suffixtags and fix handling of empty suffixtag

sagark:


sagark:
@davidbiancolin this also fixes a minor issue in the zero out dram docs which i noticed while resolving the merge conflict

PR #511, opened 2020-02-20
Fix firemarshal move

sagark:
works in firesim as top


sagark:
works in chipyard as top also.

https://github.com/ucb-bar/chipyard/pull/442

PR #512, opened 2020-02-22
Extra VCS/Verilator flags | "sed" out more verilog | Fix MIDAS level sim dependencies

abejgonzalez:
This PR does a small amount of QoL changes in addition to adding recipes for Ariane integration:

- Adds `*.swp` files to the `.gitignore`
- Removes random extra spaces at the EOL (curtesy of my Vim setup)
- Adds VCS flags for SV
- Adds Verilator flags to ignore and error on certain types of SV constructs (used in many types of projects)
- Adds `$VERILOG` to the VCS/Verilator GoldenGate simulation targets. This makes is so that the simulator is rebuilt on Verilog changes.
- Extra "sed"ding out potential Verilog statements that can occur in blackboxes. 
- Reverse how TracerV puts data into packets sent to driver (+ cleanup docs and rename code)
- Add recipes for Ariane


jerryz123:
Can you summarize what these changes do in the pr description? To preserve for posterity.

abejgonzalez:
This aligns with the following Chipyard PR: https://github.com/ucb-bar/chipyard/pull/448

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/TracerV.rst
```diff
@@ -25,7 +25,7 @@ Building a Design with TracerV
 
 In all FireChip designs, TracerV is included by default. Other targets can
 enable it by attaching a TracerV Bridge to the RISC-V trace port of one-or-more
-cores. By default, only the cycle number, instruction address, and valid bit
+cores or instruction streams (istream). By default, only the cycle number, instruction address, and valid bit
```
sagark: can you just call it instruction streams everywhere, instead of istream

deploy/sample-backup-configs/sample_config_build_recipes.ini
```diff
@@ -42,7 +42,23 @@ instancetype=z1d.2xlarge
...
 
-# Single-core, Rocket-based recipes
+# Single-core, Ariane-based recipes
+[firesim-ariane-singlecore-nic-l2-llc4mb-ddr3]
+DESIGN=FireSim
+TARGET_CONFIG=WithNIC_DDR3FRFCFSLLC4MB_FireSimArianeConfig
+PLATFORM_CONFIG=F90MHz_BaseF1Config
+instancetype=z1d.2xlarge
+deploytriplet=None
+
```
jerryz123: I would prefer to include only the no-nic option here, to reduce sample recipe bloat.

sim/midas/src/main/cc/rtlsim/Makefrag-verilator
```diff
@@ -15,7 +15,24 @@
 #   (optional) verilator_conf: An verilator configuration file
 
 VERILATOR ?= verilator --cc --exe
-override VERILATOR_FLAGS := --assert -Wno-STMTDLY -O3 \
+ifeq ($(shell grep -i "module.*ariane" $(emul_v)),)
+    verilator_assert = --assert
+endif
+override VERILATOR_FLAGS := \
+	-Wno-STMTDLY \
```
abejgonzalez: These flags are added to make Ariane work. See the following (https://github.com/pulp-platform/ariane/blob/382865dbdcb135c7e3e4305d6e0cd3a996fda3d8/Makefile#L378-L387). Specifically, they do the following:

```
PINMISSING
Warns that a module has a pin which is not mentioned in a cell instantiation. If a pin is not missing it should still be specified on the cell declaration with a empty connection, using "(.pin_name())".

IMPLICIT
Warns that a wire is being implicitly declared (it is a single bit wide output from a sub-module.) While legal in Verilog, implicit declarations only work for single bit wide signals (not buses), do not allow using a signal before it is implicitly declared by a cell, and can lead to dangling nets.

fatal
Don't error on warnings.

PINCONNECTEMPTY
Warns that a cell instantiation has a pin which is connected to .pin_name(), e.g. not another signal, but with an explicit mention of the pin. It may be desirable to disable PINCONNECTEMPTY, as this indicates intention to have a no-connect.

ASSIGNDLY
Warns that you have an assignment statement with a delayed time in front of it.

DECLFILENAME
Warns that a module or other declaration's name doesn't match the filename with path and extension stripped that it is declared in. 

UNUSED
Warns that the specified signal is never used/consumed

UNOPTFLAT
Warns that due to some construct, optimization of the specified signal is disabled.

BLKANDNBLK
BLKANDNBLK is an error that a variable comes from a mix of blocked and non-blocking assignments.

-Wno-style
Disable all code style related warning messages
```

abejgonzalez: I think we can either choose `Wno-style` or `Wall` (enable all style warnings). It depends on which you all prefer.

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/TracerV.rst
```diff
@@ -25,7 +25,7 @@ Building a Design with TracerV
 
 In all FireChip designs, TracerV is included by default. Other targets can
 enable it by attaching a TracerV Bridge to the RISC-V trace port of one-or-more
-cores. By default, only the cycle number, instruction address, and valid bit
+cores or instruction streams. By default, only the cycle number, instruction address, and valid bit
```
davidbiancolin: Can we not just call the disjunction here a hart, or am i, too, misunderstanding the term.

sim/midas/src/main/cc/rtlsim/Makefrag-verilator
davidbiancolin: I'd prefer this be exposed as a hook and the additional flags provided specially in ariane case.

However, for now could you add a comment explaining which flags you added and why (for ariane support) directly in the source.

davidbiancolin: As per Wno-style, i personally don't care. Whatever was more consistent with the old behavior.

sim/target-agnostic.mk
```diff
@@ -172,7 +172,8 @@ $(fpga_v): $(VERILOG) $(fpga_work_dir)/stamp
 	$(firesim_base_dir)/../scripts/repo_state_summary.sh > $(repo_state)
 	cp -f $< $@
 	sed -i "s/\$$random/64'b0/g" $@
-	sed -i 's/fatal/fatal(0, "")/g' $@
+	sed -i "s/void'(randomize(val));/val = '0;/g" $@
```
davidbiancolin: Can you explain this?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -154,8 +154,8 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters) extends Bridg
   lazy val toHostCPUQueueDepth  = TOKEN_QUEUE_DEPTH
   lazy val dmaSize = BigInt((BIG_TOKEN_WIDTH / 8) * TOKEN_QUEUE_DEPTH)
 
-  val uint_traces = traces map (trace => Cat(trace.valid, trace.iaddr).pad(64))
-  outgoingPCISdat.io.enq.bits := Cat(Cat(trace_cycle_counter, 
+  val uint_traces = (traces map (trace => Cat(trace.valid, trace.iaddr).pad(64))).reverse
```
davidbiancolin: So the higher number harts were appearing first in the token?

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/TracerV.rst
abejgonzalez: It is not a "hart". It is the stream of committed instructions. Thus for something like BOOM that commits N>=1 instructions, that value has to be specified here. This gets even tricker when there are multi-core designs where each core has a different amount of commit ports.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
abejgonzalez: The higher number instruction commit ports were appearing first in the token. For ex:

Token={CycleCount...Zeros...Core0_CommitPort0, Core0_CommitPort1}.

is now

Token={CycleCount...Zeros...Core0_CommitPort1, Core0_CommitPort0}.

I personally think that the main solution to this would be to make the Chisel module understand how many cores + streams per core there are and nicely emit the proper files in the driver. However, this is a temporary solution to the ordering issue.

sim/target-agnostic.mk
abejgonzalez: Ariane had a randomize statement that was getting through to synth, so I went ahead and just made it zero.

I actually don't think this is necessary anymore. I'll double check.

sim/midas/src/main/cc/rtlsim/Makefrag-verilator
abejgonzalez: Can you clarify what you mean by "hook"? Do you want an `ifdef` like statement to add the flags if Ariane is found? Or do you want something else?

abejgonzalez: Discussed offline: Just have the minimal set of new flags with explanations.

abejgonzalez:
@alonamid @sagark @jerryz123 @davidbiancolin. This PR is now ready for a full review. I was able to boot no-nic configs with all major cores (ariane, boom, rocket)

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/TracerV.rst
```diff
@@ -270,11 +270,11 @@ when using TracerV under certain conditions:
 * TracerV by default outputs only instruction address and a valid bit and assumes
   that the combination of these fits within 64 bits. Changing this requires
   modifying ``sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala``.
-* The number of cores/instruction streams is currently not automatically detected.
-  To collect data for multiple cores/instruction streams, you must change the
-  ``NUM_CORES`` macro at the top of ``sim/firesim-lib/src/main/cc/bridges/tracerv.h``.
+* The number of cores or instruction streams is currently not automatically detected.
+  To collect data for multiple cores or instruction streams, you must change the
+  ``NUM_INSTR_STREAMS`` macro at the top of ``sim/firesim-lib/src/main/cc/bridges/tracerv.h``.
```
sagark: this needs to be changed back per discussion in firesim meeting

sim/midas/src/main/cc/rtlsim/Makefrag-verilator
```diff
@@ -15,7 +15,30 @@
...
+	-Wno-UNOPTFLAT \
+	-Wno-BLKANDNBLK \
+	-Wno-style \
+	-Wall
+
+# normal flags used for midas builds (that are incompatible with ariane)
+MIDAS_VERILATOR_FLAGS = \
+	--assert
+
+override VERILATOR_FLAGS := \
```
davidbiancolin: If we're going to go as far as detecting what target is coming in from the top we should just emit this from the generator now.  We can put it in the generate makefrag which is already being sourced my the target-specific makefrag in sim/

abejgonzalez: Discussed offline: As a middle ground, I moved this to the FireSim makefrag located in `sim/src/main/makefrag/firesim/Makefrag`

sim/midas/src/main/cc/Makefile
```diff
@@ -58,7 +59,7 @@ $(OUT_DIR)/$(DESIGN).chain:
 
 override CXXFLAGS += -I$(midas_dir) -I$(util_dir)
 # The trailing whitespace is important for some reason...
-override LDFLAGS := $(LDFLAGS) -L$(GEN_DIR) -lstdc++ -lpthread -lgmp -lmidas 
+override LDFLAGS := $(LDFLAGS) -L$(GEN_DIR) -lstdc++ -lpthread -lgmp -lmidas
```
abejgonzalez:
```suggestion
override LDFLAGS := $(LDFLAGS) -L$(GEN_DIR) -lstdc++ -lpthread -lgmp -lmidas 
```

PR #513, opened 2020-02-22
Bump verilator to v4.028

abejgonzalez:
Chipyard uses v4.016 but I am bumping to v4.028 in other locations (including Chipyard). I think it would be good to follow.


abejgonzalez:
Closing since #512 addresses this.

PR #514, opened 2020-02-26
Add comment with no_net_config to sample_config_runtime

jerryz123:
I can never remember the name of this config correctly.

### DIFF
--- a/deploy/sample-backup-configs/sample_config_runtime.ini
+++ b/deploy/sample-backup-configs/sample_config_runtime.ini
@@ -15,7 +15,7 @@ spotinterruptionbehavior=terminate
 spotmaxprice=ondemand
 
 [targetconfig]
-#topology=no_net_config
+#Set topology=no_net_config to run without a network simulation
 topology=example_8config
 no_net_num_nodes=2
 linklatency=6405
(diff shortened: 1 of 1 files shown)


deploy/sample-backup-configs/sample_config_runtime.ini
```diff
@@ -15,6 +15,7 @@ spotinterruptionbehavior=terminate
 spotmaxprice=ondemand
 
 [targetconfig]
+#topology=no_net_config
```
sagark:
```suggestion
#Set topology=no_net_config to run without a network simulation
```

davidbiancolin:
Probably the biggest QoL improvement of the last year :P 

Though you probably should have put the setting on a different line so users could just delete the # and be on their way

jerryz123:
> Though you probably should have put the setting on a different line so users could just delete the # and be on their way
That's what I had originally, blame @sagark

davidbiancolin:
> > Though you probably should have put the setting on a different line so users could just delete the # and be on their way
> 
> That's what I had originally, blame @sagark

Rofl, but what i mean is:
# Text explaining how to enable no network simulation
# <Line to uncomment>

PR #515, opened 2020-02-27
Move dramsim files to testchipip

zhemao:
We are adding mm_dramsim2.cc to testchipip. This change switches midas to use the sources there instead of duplicating them. 

https://github.com/ucb-bar/testchipip/pull/66


sim/midas/src/main/cc/Makefile
```diff
@@ -39,11 +39,15 @@ $(info clock period: $(CLOCK_PERIOD))
 
 shim := FPGATop
 
-override CXXFLAGS := $(CXXFLAGS) -std=c++11 -Wall -I$(midas_dir)/dramsim2
+testchip_dir=$(TOP_DIR)/generators/testchipip/src/main/resources
```
a0u: Maybe add a check for `TOP_DIR` being defined, like `DESIGN`/`GEN_DIR`/`DRIVER` above?
```Makefile
ifeq ($(strip $(TOP_DIR)),)
$(error Define TOP_DIR, the path to the root of Chipyard)
endif
```

zhemao:
@davidbiancolin I have tested both.

davidbiancolin:
@zhemao deeeeeeeeeleeeeeeeeeteeeeeeeee meeeeeeeee

PR #526, opened 2020-03-28
Trigger System Documentation

davidbiancolin:
This adds a page describing how to add triggers to a chisel design and the structure of the generated target circuit.


docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
+Simulation Triggers
+===================
+
+It is often useful to globally coordinate debug and instrumentation features using specific target-events that may be distributed across the target design.
+For instance, you may wish to enable collection of synthesized prints and sampling of
+AutoCounters simulataenously when a specific instruction is committed on any
+core, or alternatively if the memory system sees a ready or write to a
```
alonamid:
```suggestion
core, or alternatively if the memory system sees a ``ready`` or writes to a
```

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+particular memory address. Golden Gate's trigger system enables this by aggregating annotated ``TriggerSources``
+distributed throughout the design using a centralized credit-based system
+which then drives a single-bit level-sensitive enable to all ``TriggerSinks`` sinks distributed throughout the design.
+This enable is asserted while the design remains in the region-of-interested
```
alonamid:
```suggestion
This enable signal is asserted while the design remains in the region-of-interest
```

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+(ROI).  Sources signal the start of the ROI by granting a
+credit and signal the end of the ROI by asserting a debit. Since there can be
+multiple sources, each of which might grant credits, the trigger is only
+disabled when the system has been debited as exactly as many times as it has
+been credited (it has a balance of 0).
```
alonamid: Does ordering matters here?

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+***************
+In order to permit distributing trigger sources across the whole design, you
+must annotate distinct boolean signals as credits and debits using methods
+provided by the ``TriggerSource`` object. We provide an example below.
+
+.. literalinclude:: ../../sim/src/main/scala/midasexamples/TriggerWiringModule.scala
+    :language: scala
+    :start-after: DOC include start: TriggerSource Usage
+    :end-before: DOC include end: TriggerSource Usage
+
```
alonamid: This example is fine for very advanced users, but I think it might also be worth also having a simpler example that doesn't use the credit and debit terminology.
To be frank, if I wanted to re-implement the TracerV trigger based on reading these docs and this example, I don't think I would be able to understand how to do it.
What would help me is having an example that just uses a single trigger boolean signal (with no credit and debit terminology), that I essentially want to "magically and deterministically" (this is where the GG magic happens using this entire credit/debit system) wire to the Sink.

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+the design is under reset are not counted (the reset used is implicit
+reset of the Chisel Module in which you invoked the method).
+If the module provides no implicit reset or if you wish to credit or debit the trigger system while the
+local module's implicit reset is asserted, use ``TriggerSource.{creditEvenUnderReset, debitEvenUnderReset}`` instead.
+
+Trigger Sinks
+*************
+Like sources, trigger sinks are boolean signals that have been annotated
+alongside their associated clock. These signals will be driven by a Boolean
+value created by the trigger system. If trigger sources exist your design, the
```
alonamid:
```suggestion
value created by the trigger system. If trigger sources exist in your design, the
```

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+generated trigger will override all assignments made in the chisel to the same
```
alonamid: hmm.... I think this is important to emphasize, because this is not trivial (that you don't need to do any special handling in case a trigger exists or doesn't exist. That it will just override using FIRRTL last-connect semantics).
Maybe highlight in bold?

docs/Golden-Gate/Triggers.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Bridges that use the default ``HostPort`` interface add an additional cycle of
+latency in the bridge's local domain since their token channels model a single
+register stage to improve simulation FMR. Thus, without using a different
+``HostPort`` implementation, trigger sources generated by a Bridge and trigger
+sinks that feed into a Bridge will each see one additional bridge-local cycle
+of latency. In constrast, synthesized printfs and assertions, and AutoCounters
+all use wire channels (since they are unidirectional interfaces, the extra
+register stage is not required to improve FMR) and will see no additional sink
+latency.
```
alonamid: So the bottom line is that a trigger will be offset by one base-clock cycle compared to all other current debugging instrumentation?

alonamid:
These docs are very nuanced and detailed, and they should be since they are in the Golden Gate section. 
However, I do think that the usage instructions get a little lost within the nuances. It might be worth having a brief "quick usage" section, which just describes the simple case of adding the simplest trigger?

It is important to document the debit/credit system, but I think that given the current limitation of the system (single trigger), from a user API perspective it's easier to think of it as a "global AND" between trigger sources. I understand the nuances of the debits and credits between bridges and across clock domains, but I currently couldn't think of a scenario where a would like to add a trigger and think about the various credits and debits rather than a "global AND". Alternatively - maybe give an example of such a scenario and contrast it against a simple single trigger might also be a good way to explain the system.

docs/Golden-Gate/Triggers.rst
davidbiancolin: If i understand you correctly, yes, at least. 

in other words if you wrote:
```
val creditSig = <some bool>
val sink = <some bool>
TriggerSource.credit(creditSig)
TriggerSink(sink)
printf(synthesizePrintf(credit)))
printf(synthesizePrintf(sink)))
```
In the captured printf trace, sink would go high 2 cycles after credit went high. 

If you look at the provided waveform, you can definitely imagine situations where the trigger
reaches the sinks in certain clock domains in < 1 base clock cycle (e.g., when both the source and sink domain are faster than the base domain).

docs/Golden-Gate/Triggers.rst
davidbiancolin: Not within a cycle.

PR #527, opened 2020-03-29
Multiclock Documentation Part 1

davidbiancolin:
I added a subsection in Targets which provides i think most of they key information a user needs to know if they want to simulate a system with multiple clocks.

A longer-form section in the Golden Gate documentation (a la Triggers) will follow. There I'll provide something more specification-like, and describe the current implementation.


davidbiancolin:
Opened against master, lol.

PR #532, opened 2020-04-15
Chipyard Rocket Chip Bump

davidbiancolin:
@albert-magyar  Can you review the firrtl change
@zhemao Can you review the reset please. 

The big changes here are:
1) Changing how we handle AsyncReset -> introduces a CoerceAsyncToSyncReset
  -> We can delete the old black-box based pass in the future
2) Copies all the old generatorUtils sources into a midas.rocketchip package
  -> I will wean us off of that before the next release

See ucb-bar/chipyard#503.


sim/midas/src/main/scala/midas/passes/CoerceAsyncToSyncReset.scala
```diff
@@ -0,0 +1,34 @@
...
+  def inputForm = LowForm
+  def outputForm = LowForm
+
+  private def onType(t: Type): Type = t match {
+    case AsyncResetType => BoolType
+    case ResetType => BoolType
+    case o => o
+  }
+  private def onExpr(e: Expression): Expression = e.map(onExpr).map(onType) match {
+    case p@DoPrim(PrimOps.AsAsyncReset, Seq(arg), _, _) => arg
```
albert-magyar, line 27:
```suggestion
    case p @ DoPrim(PrimOps.AsAsyncReset, Seq(arg), _, _) => arg
```

albert-magyar: Technically the map(onType) would have already made `p.tpe` be `BoolType`, but it looks a lot less weird with the extra copy arg.

sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
```diff
@@ -45,6 +45,11 @@ private[midas] class MidasTransforms(implicit p: Parameters) extends Transform {
       firrtl.passes.SplitExpressions,
       firrtl.passes.CommonSubexpressionElimination,
       new firrtl.transforms.DeadCodeElimination,
+      new firrtl.transforms.InferResets,
+      firrtl.passes.CheckTypes,
+      CoerceAsyncToSyncReset,
+      new ResolveAndCheckNoReset,
```
davidbiancolin, line 51: Oops, forgot to remove the dumping.

PR #542, opened 2020-05-04
Fix Host AXI4 ID Truncation

davidbiancolin:
I restricted the memory ID space without actually adding a ID Indexer. This fixes that by using the full ID width available on the host. If the required ID space of all bridges exceeds the available width, users can inject the IDIndexer using `HostMemIdSpaceKey`.

This also bumps chipyard to work around out issue with the SiFive UART driver.

### DIFF
--- a/deploy/sample-backup-configs/sample_config_hwdb.ini
+++ b/deploy/sample-backup-configs/sample_config_hwdb.ini
@@ -10,32 +10,37 @@
 # own images.
 
 [firesim-boom-singlecore-nic-l2-llc4mb-ddr3]
-agfi=agfi-0750d8ac4485b4e76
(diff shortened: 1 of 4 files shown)


davidbiancolin:
> Is there any require statement that checks whether the diplomatically decided number of ID bits exceeds the Host memory ID bits? Would be helpful to add that.

Good catch. Forgot to cherry-pick that out of my feature branch.

zhemao:
We should add a similar check for other diplomatically elaborated fields, specifically the number of address bits.

davidbiancolin:
The address bits are effectively checked when the total DRAM allocated is summed up, but i can do that.

PR #548, opened 2020-05-07
Remove Platform-Specific Handling of MIDAS-Level Simulation

davidbiancolin:
One of the things users had to do if they wanted to bring up their own host platform was supply a bunch of harness code for doing ML simulation, and that code had to be hosted in Golden Gate.

This removes that need by simulating the submodule hierarchy below FPGATop. Thus the user need only define a `Config` for their host the defines the widths of the AXI4 interfaces. I Think this is safe since the RTL emitted in the shim has historically been static, and relatively thin / glue logic and so can be tested as part of FPGA-level simulation. 

I added two tests that run an AXI4Fuzzer targeting the ZC706 and Zedboard to demonstrate this.


sim/midas/src/main/scala/midas/core/FPGATop.scala
```diff
@@ -62,13 +62,17 @@ class FPGATopIO(implicit val p: Parameters) extends WidgetIO {
...
 case class HostMemChannelParams(
-  size: BigInt,
-  beatBytes: Int,
-  idBits: Int,
-  maxXferBytes: Int = 256)
+    size: BigInt,
+    beatBytes: Int,
+    idBits: Int,
+    maxXferBytes: Int = 256) {
+  def axi4BundleParams = AXI4BundleParameters(log2Ceil(size), 8 * beatBytes, idBits)
```
zhemao, line 69: Probably best to use keyword arguments here.

davidbiancolin: Sounds good.

PR #551, opened 2020-05-13
Separate plusargs per memory channel and synthesized print settings in manager

zhemao:
Three unrelated changes here

1. Make dramsim a standalone makefrag variable so that it's easier to disable for midas-level sim
2. Allow the DRAM timing model for each channel to have a different set of parameters
3. Expose options for synthesized prints in the config_runtime.ini

### DIFF
--- a/deploy/sample-backup-configs/sample_config_runtime.ini
+++ b/deploy/sample-backup-configs/sample_config_runtime.ini
@@ -57,3 +57,8 @@ suffixtag=
 # In general, this is not required to produce deterministic simulations on
 # target machines running linux. Enable if you observe simulation non-determinism.
 zerooutdram=no
+
+[synthprint]
+start=0
+end=-1
+cycleprefix=yes
--- a/deploy/sample-backup-configs/sample_config_runtime.ini
+++ b/deploy/sample-backup-configs/sample_config_runtime.ini
@@ -59,6 +59,10 @@ suffixtag=
 zerooutdram=no
 
 [synthprint]
+# Start and end cycles for outputting synthesized prints.
+# They are given in terms of the base clock and will be converted
+# for each clock domain.
 start=0
 end=-1
+# When enabled (=yes), prefix print output with the target cycle at which the print was triggered
 cycleprefix=yes
--- a/sim/midas/src/main/scala/midas/models/dram/FASEDMemoryTimingModel.scala
+++ b/sim/midas/src/main/scala/midas/models/dram/FASEDMemoryTimingModel.scala
@@ -22,6 +22,10 @@ import Console.{UNDERLINED, RESET}
 
 import java.io.{File, FileWriter}
 
+object ModelID {
+  var id = 0
+}
+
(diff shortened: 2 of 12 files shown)


sim/midas/src/main/scala/midas/models/dram/FASEDMemoryTimingModel.scala
```diff
@@ -22,6 +22,10 @@ import Console.{UNDERLINED, RESET}
 
 import java.io.{File, FileWriter}
 
+object ModelID {
```
davidbiancolin, line 25: I mentioned this earlier but we should reuse the ID in Widget.scala. We could add a method, `getInstanceID` there.

sim/midas/src/main/scala/midas/models/dram/FASEDMemoryTimingModel.scala
```diff
@@ -552,9 +552,9 @@ class FASEDMemoryTimingModel(completeConfig: CompleteConfig, hostParams: Paramet
     // Accepts an elaborated memory model and generates a runtime configuration for it
     private def emitSettings(fileName: String, settings: Seq[(String, String)])(implicit p: Parameters): Unit = {
       val file = new File(p(OutputDir), fileName)
-      val writer = new FileWriter(file)
+      val writer = new FileWriter(file, wId != 0)
```
davidbiancolin, line 555: i'm asusming this is to denote appending to an existing file --> can you include the kwarg if it has a good name

deploy/sample-backup-configs/sample_config_runtime.ini
```diff
@@ -57,3 +57,8 @@ suffixtag=
 # In general, this is not required to produce deterministic simulations on
 # target machines running linux. Enable if you observe simulation non-determinism.
 zerooutdram=no
+
+[synthprint]
```
davidbiancolin, line 61: This probably needs a comment explaining what a base cycle is, or a link to the docs.

sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -95,6 +95,7 @@ EXTRA_VERILATOR_FLAGS = \
 # SW RTL Simulation Args -- for MIDAS- & FPGA-level Simulation #
 ################################################################
 TIMEOUT_CYCLES = 100000000
+DRAMSIM = +dramsim
```
davidbiancolin, line 98: how about HOST_MEM_ARGS

sim/midas/src/main/scala/midas/models/dram/FASEDMemoryTimingModel.scala
zhemao: This is a java class, so  I don't think keyword arguments work.

davidbiancolin: Huh, the more you know.

PR #553, opened 2020-05-14
Reduce Verilator make to -j4 to reduce CI memory footprint

albert-magyar:
@davidbiancolin let's see if this passes CI


albert-magyar:
This fixed the crash in the 4GB CI context. Also, building Verilator is very fast, even under `-j4`, so this adds negligible overhead to the `machine-launch-script.sh` runtime.

PR #554, opened 2020-05-15
Remove Strober Sources

davidbiancolin:
I hope that one day we'll bring Strober up in FireSim, but now that MIDAS has been inlined there's less reason to keep its code around. This removes all [dead] strober-related sources, and cleans up the available configs a little bit.

One question: should driver code that depends on the width of the control interface be parameterized by header generation? The CTRL-is-32bits assumption is baked into some of the driver code but parameterized in other places. AXI4-lite is defined to be 32-bits wide so making this assumption is not entirely unreasonable.


sim/midas/src/main/cc/simif_emul.cc
```diff
@@ -142,13 +142,15 @@ void simif_emul_t::wait_read(std::unique_ptr<mmio_t>& mmio, void *data) {
 
 void simif_emul_t::write(size_t addr, data_t data) {
   size_t strb = (1 << CTRL_STRB_BITS) - 1;
-  master->write_req(addr << CHANNEL_SIZE, CHANNEL_SIZE, 0, &data, &strb);
+  // LS by 2 because the driver presents word addresses, but AXI4-lite expects
+  // byte addresses
+  master->write_req(addr << 2, 2, 0, &data, &strb);
```
albert-magyar, line 147: Is there a reason to drop the use of CHANNEL_SIZE as a stand-in for the literal instead of just moving the definition?

davidbiancolin: It used to refer to the maximum AXI4 size of the control interface. I should probably just add it back with a better name since there exists no global notion of a "Channel's width" now.

sim/midas/src/main/cc/simif_emul.cc
```diff
@@ -142,15 +142,16 @@ void simif_emul_t::wait_read(std::unique_ptr<mmio_t>& mmio, void *data) {
...
-  master->write_req(addr << 2, 2, 0, &data, &strb);
+  static_assert(CTRL_AXI4_SIZE == 2, "AXI4-lite control interface has unexpected size");
+  master->write_req(addr << CTRL_AXI4_SIZE, CTRL_AXI4_SIZE, 0, &data, &strb);
   wait_write(master);
 }
 
 data_t simif_emul_t::read(size_t addr) {
   data_t data;
-  master->read_req(addr << 2, 2, 0);
+  master->read_req(addr << CTRL_AXI4_SIZE, CTRL_AXI4_SIZE, 0);
```
albert-magyar, line 154: 👍

sim/midas/src/main/scala/midas/passes/StroberUtils.scala
```diff
@@ -1,17 +1,19 @@
...
 import firrtl.Mappers._
 import firrtl.annotations._
-import core.ChainType
-import mdf.macrolib.SRAMMacro
-import mdf.macrolib.Utils.readMDFFromString
 import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet, LinkedHashSet}
 
+/**
+  * This file contains legacy FIRRTL features that did not exist in upstream
+  * FIRRTL at the time of Strober's development. These continue to be used in:
```
albert-magyar, line 13: Seems like a reasonable compromise, moving to `midas.passes` but keeping the `Strober` name in `StroberUtils`.

davidbiancolin:
CI is failing because it's missing Abe's fix.

timsnyder-siv:
@davidbiancolin is there an issue filed somewhere for getting the at least the simulation-replay portions of the strober flow working again?  I don't see one, just wanted to make sure I'm not missing it.

PR #556, opened 2020-05-16
Use wdata option for dromajo

abejgonzalez:
This follows the TraceIO changes in Chipyard to connect to the `wdata` Option type.

Related PRs:
https://github.com/ucb-bar/chipyard/pull/553
https://github.com/ucb-bar/testchipip/pull/85


abejgonzalez:
> There's a CY bump that's coming with this right?

I don't know what the policy is for bumping CY with FireSim but my impression is that I first get this merged in then separately bump CY in FireSim? Is there a better way to do it?

PR #557, opened 2020-05-17
Generator Unification Phase 1

davidbiancolin:
This PR removes all FireSim-hosted generation and uses the Chipyard generator for elaborating the targets instead.  Also add a test to check for key FCCA annotation serialization/deserialization.

I stopped short of trying to reuse the recipes in `common.mk` as that is going to be a little more difficult than i had initially hoped. 

Longer term TODOs:
- [x] Decide where to set cover / propertyLib handing @alonamid 
  -  `cover.setPropLib(new midas.passes.FireSimPropertyLibrary())`
  - open an issue; discuss in Fsim meeting
- [x] Fix makefrag generation for device tests
  - open an issue; discuss in CY meeting


sim/midas/src/test/scala/midas/FAMEAnnotationSerializationSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  val baseFCCA = FAMEChannelConnectionAnnotation(
+    globalName = "test",
+    channelInfo = PipeChannel(1),
+    clock = None,
+    sources = None,
+    sinks = None)
+
+  "PipeChannel FCCAs" should "serialize and deserialize correctly" in {
+    val deserAnno = serializeAndDeserialize(baseFCCA)
+    assert(baseFCCA == deserAnno)
```
albert-magyar, line 27: Seems like a good test to add.

PR #558, opened 2020-05-17
machine-launch-script log + fix git install bug + chipyard bump

abejgonzalez:
Have the `machine-launch-script` redirect output to a log file to find issues and use `git224` instead of `git2u` (it looks like `git2u` was removed). Also fixes FireSim's CY pointer.


abejgonzalez:
I also think that this should be potentially changed so that the FireSim CI is able to catch this and error.

davidbiancolin:
This otherwise LGTM but can you `tee` both stderr and standard out so that they show up in the CI output. 

We'll make CI catch machine setup errors in a separate PR.

davidbiancolin:
The git version change might be problematic, but hopefully not. :|

PR #559, opened 2020-05-18
Add Chipyard env by default to FireSim env

abejgonzalez:
For Chipyard-as-top systems, the `firesim-setup.sh` script is normally used with `--library`. This will disable building the toolchain, but as a result, the Chipyard env script is not sourced when running `source sourceme-f1-manager.sh`. This changes it so that the proper source is added by default (whether `SKIP_TOOLCHAIN` is disabled or not).

Note: This requires that `build-toolchains` is run before `source sourceme-f1-manager.sh` (which is implied in the Chipyard documentation)

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -164,8 +164,9 @@ if [ "$SKIP_TOOLCHAIN" != true ]; then
         fi
     )
     source "$target_chipyard_dir/env.sh"
-    env_append "source $target_chipyard_dir/env.sh"
 fi
+
+env_append "source $target_chipyard_dir/env.sh"
 
 cd $RDIR
 


abejgonzalez:
This is probably also another good thing to check with the CI!

davidbiancolin:
I asked in the Chipyard slack about this and it seemed like people wanted to supply their own $RISCV when using FireSim under Chipyard, which is to say doing this would clobber whatever existed in the users environment (which i understand was how it was before). 

@jerryz123 @colinschmidt Do you have thoughts on this?

davidbiancolin:
Basically the model is you should have already sourced Chipyard's env since you're using Chipyard as top.

abejgonzalez:
IIRC this breaks `buildafi` with Chipyard-as-top... But maybe its because I didn't source the Chipyard `env.sh` first. I'll double-check this quickly.

davidbiancolin:
Alternatively, we could conditionally source the env.sh in CY if the CY_ENV_SOURCED thing isn't already set. Ultimately we should do w.e CY users think is most intuitive.

abejgonzalez:
I think either approach is fine. However, if people want to provide their own $RISCV without CY then it might be best to leave things alone. I'll once again check that this isn't an issue.

colinschmidt:
I want chipyard and all subcomponents(including FireSim) to respect my currently set `$RISCV` .

davidbiancolin:
I like being responsible for less shit.

abejgonzalez:
Huh. So it looks like the `buildafi` fails on the `f1` make target with the Chipyard `env.sh` sourced before the FireSim `sourceme`. I believe its because `buildafi` re-sources the `sourceme-f1-manager.sh` in a clean environment? See `run` when building the fpga driver https://github.com/firesim/firesim/blob/110872157bd36a7b36afb3fbd7a095ee2abd1f07/deploy/buildtools/buildafi.py#L44-L45. Ill try to see if there is a way to have the current environment be used... or maybe just pass the $RISCV variable(s).

davidbiancolin:
> . Ill try to see if there is a way to have the current environment be used... or maybe just pass the $RISCV variable(s).

Ugh sorry, yeah i think if you can use the current environment that will lead to fewer surprises. Funny enough i don't think we need to build the F1 driver there, but you'll likely run into the same problem if you try running `infrasetup`.

abejgonzalez:
Closing since this will be addressed in a different PR.

PR #560, opened 2020-05-18
Chipyard-as-Top RISC-V variables used in buildafi/infrasetup

abejgonzalez:
This fixes the problem mentioned in #559 where Chipyard-as-Top FireSim would not use the RISC-V variables during `buildafi` and `infrasetup`. This makes it so that the default `RISCV/LD_LIB_PATH/PATH` are added before FireSim's `sourceme` so that it can be present and overridden by FireSim's `sourceme`.

### DIFF
--- a/deploy/buildtools/buildafi.py
+++ b/deploy/buildtools/buildafi.py
@@ -4,6 +4,7 @@ import time
 import random
 import string
 import logging
+import os
 
 from fabric.api import *
 from fabric.contrib.console import confirm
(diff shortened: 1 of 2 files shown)


davidbiancolin:
Do you know how hard it would be to just pull in the entire environment instead of specific variables?

abejgonzalez:
From what I can tell from the `1.14` Fabric documentation, neither `local` or `run` use the local env. variables. However, if we bump to the latest Fabric, we should be able to do `run.replace_env` to just copy the fully environment (but that seems to be a much larger PR/solution to implement).

PR #562, opened 2020-05-19
add expect to machine-launch-script.sh

abejgonzalez:
I was hoping to add `expect` to the launch script so that BOOM's FireSim CI doesn't have to install it on the manager itself (`expect` is used to talk to interactive programs so that they can be scripted). Currently, I just append the install to the end of the launch script in my own CI every time I launch a manager instance. Depending on how FireSim wants to do its CI (and Chipyard's FireSim CI), we may want to add this globally.

### DIFF
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -49,6 +49,9 @@ sudo yum -y install bash-completion
 # graphviz for manager
 sudo yum -y install graphviz python-devel
 
+# used for CI
+sudo yum -y expect
+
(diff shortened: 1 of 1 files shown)


scripts/machine-launch-script.sh
```diff
@@ -49,6 +49,9 @@ sudo yum -y install bash-completion
 # graphviz for manager
 sudo yum -y install graphviz python-devel
 
+# used for CI
+sudo yum -y expect
```
abejgonzalez, line 53:
```suggestion
sudo yum -y install expect
```

PR #563, opened 2020-05-20
Make machine-launch fail on error. 

davidbiancolin:
CI will correctly pick up problems and the log will end at the problematic command. 

h/t @abejgonzalez

### DIFF
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -36,7 +36,7 @@ rm -rf $DTCversion
 # get a proper version of git
 sudo yum -y remove git
 sudo yum -y install epel-release
-sudo yum -y install https://centos7.iuscommunity.org/ius-release.rpm
+sudo yum -y install https://repo.ius.io/ius-release-el7.rpm
 sudo yum -y install git224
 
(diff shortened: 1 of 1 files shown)


davidbiancolin:
And because of on going weirdness with ius they've tested this for me.

scripts/machine-launch-script.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e
```
abejgonzalez, line 3: Would this be better as a `set -ex` so that you also get to what command runs and its args?

abejgonzalez:
Wow. This failure where the `https://centos7.iuscommunity.org/ius-release.rpm` can't be opened is now just happening (this didn't error out 8hrs ago). Going to the link points to this issue: https://github.com/iusrepo/announce/issues/18

abejgonzalez:
You should change the `https://centos7.iuscommunity.org/ius-release.rpm` link to `https://repo.ius.io/ius-release-el7.rpm`

davidbiancolin:
Cool, easy enough.

PR #570, opened 2020-05-22
Update nbd install

abejgonzalez:
Occasionally, the NBD module is built incorrectly causing `infrasetup` to fail inserting the module. This is a solution done by Sam Steffl that seems to resolve the issue.

### DIFF
--- a/scripts/install-nbd-kmod.sh
+++ b/scripts/install-nbd-kmod.sh
@@ -13,37 +13,43 @@ KURL="http://vault.centos.org/7.6.1810/updates/Source/SPackages/$KRPM"
 
 GENERICBUILDDIR=$(pwd)/build
 NBDBUILDDIR=$GENERICBUILDDIR/nbdbuild
+KBUILDDIR=$NBDBUILDDIR/BUILD/kernel-$KSRC/linux-$KSRC.x86_64
+OUTPUTFILE=$GENERICBUILDDIR/nbd.ko
 
 mkdir -p $GENERICBUILDDIR
 cd $GENERICBUILDDIR
 echo $(pwd)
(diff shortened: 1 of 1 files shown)


abejgonzalez:
This was tested in the FireSim CI work here (https://github.com/riscv-boom/riscv-boom/pull/409)

scripts/install-nbd-kmod.sh
```diff
@@ -13,37 +13,43 @@ KURL="http://vault.centos.org/7.6.1810/updates/Source/SPackages/$KRPM"
...
+cd $KBUILDDIR
 sed -i 's/REQ_TYPE_SPECIAL/REQ_TYPE_DRV_PRIV/g' $KBUILDDIR/drivers/block/nbd.c
 
-# now build
-make -j32 prepare
-make -j32 modules_prepare
+# now clean/setup .config
+cd $KBUILDDIR
+make -j32 mrproper
+make -j32 oldconfig
```
a0u, line 43: Just to be clear, this is a fairly significant functional change to the flow:
* `mrproper` removes the kernel config initialized by the RPM `%prep` stage.  This is the config that ships with the source RPM, which I expect to match the binary package deployed in the AMI.
* `oldconfig` replaces it with a config taken from the kernel currently running on the build host, usually `/boot/config-$(uname -r)`.

On a fresh manager instance, the two configs should be identical (but need to confirm that).  However, they can diverge if the user ran `yum upgrade` on the manager before the build.

a0u: Also, `oldconfig` is potentially interactive.  If an option is undefined, it prompts the user to select its value.  Instead use `olddefconfig` to automatically choose the default.

PR #573, opened 2020-05-26
Fix Synthesizable Unittests

davidbiancolin:
I'd like to include these in CI but they've bit rotted a bit through the last couple of RC bumps.

### DIFF
--- a/sim/midas/src/main/scala/midas/SynthUnitTests.scala
+++ b/sim/midas/src/main/scala/midas/SynthUnitTests.scala
@@ -23,7 +23,8 @@ class WithAllUnitTests extends Config((site, here, up) => {
       Module(new ReadyValidChannelUnitTest(timeout = timeout)),
       Module(new CounterTableUnitTest),
       Module(new LatencyHistogramUnitTest),
-      Module(new AddressRangeCounterUnitTest))
+      Module(new AddressRangeCounterUnitTest)
+    )
   }
 })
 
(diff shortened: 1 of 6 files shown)


sim/midas/src/main/scala/midas/SynthUnitTests.scala
```diff
@@ -23,7 +23,8 @@ class WithAllUnitTests extends Config((site, here, up) => {
       Module(new ReadyValidChannelUnitTest(timeout = timeout)),
       Module(new CounterTableUnitTest),
       Module(new LatencyHistogramUnitTest),
-      Module(new AddressRangeCounterUnitTest))
```
zhemao, line 26: Is there any reason to change this beyond aesthetics?

sim/midas/src/main/scala/midas/core/Channel.scala
```diff
@@ -241,6 +241,8 @@ class ReadyValidChannelUnitTest(
 
   val dut = Module(new ReadyValidChannel(payloadType))
   val reference = Module(new ShiftQueue(payloadType, queueDepth))
+  val referenceReset = Wire(Bool())
+  reference.reset := referenceReset
```
zhemao, line 245: Why does this need to be a wire?

sim/midas/src/main/scala/midas/SynthUnitTests.scala
davidbiancolin: i was commenting lines out to figure out which test was causing FIRRTL to croak, this made that more brainless.

sim/midas/src/main/scala/midas/core/Channel.scala
davidbiancolin: reference.reset is an abstract Reset. For the LIBDN tester hardware to generate correctly it needs to be a concrete reset -- specifically a Bool in this case.  Using the wire here both gives us a reference to drive from the LIBDN tester the and forces a particular concrete type on to the port.

PR #582, opened 2020-06-17
Fix Index Handling in Assertion Synthesis 

### LINKED ISSUE #581: WithSynthAsserts identity incorrect (at least in RTL simulation)
I was curious whether `WithSynthAsserts` would work when running [MIDAS-level RTL simulation in Verilator](https://docs.fires.im/en/latest/Advanced-Usage/Debugging-in-Software/RTL-Simulation.html#midas-level-simulation)

To try it, I added an assertion that I knew would fire, built my verilator and ran `make run-asm-tests`, an assertion fired but the printed message did not correspond to the assertion that I added.

To reproduce this with FireSim 1.10:
- make a clean clone of FireSim 1.10 && `build-setup.sh fast`
- apply the following patch to `target-design/chipyard/generators/rocket-chip`
```
diff --git a/src/main/scala/rocket/Frontend.scala b/src/main/scala/rocket/Frontend.scala
index 9d9adb739..e911ac3cf 100644
--- a/src/main/scala/rocket/Frontend.scala
+++ b/src/main/scala/rocket/Frontend.scala
@@ -88,6 +88,7 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
   val clock_en = clock_en_reg || io.cpu.might_request
   io.cpu.clock_enabled := clock_en
   assert(!(io.cpu.req.valid || io.cpu.sfence.valid || io.cpu.flush_icache || io.cpu.bht_update.valid || io.cpu.btb_update.valid) || io.cpu.might_request)
+  assert(!io.cpu.req.valid)
   val gated_clock =
     if (!rocketParams.clockGate) clock
     else ClockGate(clock, clock_en, "icache_clock_gate")
```
   - the new assertion will fire when we get the first valid TL transaction into the Frontend
- `make -C sim PLATFORM_CONFIG=WithSynthAsserts_BaseF1Config && make -C sim run-asm-tests`

When I ran, I got a message about an assertion in `pbus.coupler_to_blkdevctrl.fragmenter`:
```
-bash-4.2$ cat output/f1/FireSim-FireSimRocketConfig-WithSynthAsserts_BaseF1Config/rv64si-p-csr.out 
random min: 0x0, random max: 0xffffffffffffffff
Commencing simulation.
C0:         19 [1] pc=[0000000000010040] W[r10=0000000000010040][1] R[r 0=0000000000000000] R[r 0=0000000000000000] inst=[00000517] auipc   a0, 0x0
C0:         20 [1] pc=[0000000000010044] W[r10=0000000000010000][1] R[r10=0000000000010040] R[r 0=0000000000000000] inst=[fc050513] addi    a0, a0, -64
C0:         21 [1] pc=[0000000000010048] W[r 0=0000000000000000][1] R[r10=0000000000010000] R[r 0=0000000000000000] inst=[30551073] csrw    mtvec, a0
module: TLFragmenter, path: FireSim.TestHarness.subsystem_pbus.coupler_to_slave_named_blkdevcontroller.fragmenter]
Assertion failed
    at Fragmenter.scala:300 assert (!repeater.io.full || !aHasData)
 at cycle: 71
```

davidbiancolin:
Changes to assertion synthesis for multiclock introduced a regression where synthesized assertions might not be caught, or linked to the wrong message. This resolves #581. 

Also, adds an AssertTorture test that randomly generates a module hierarchy with a large number of assertions attached to one of multiple available clocks. One assertion fires per clock-domain, per-cycle. The assertion order is verified by comparing the message to a printf that fires under the inverse of same condition.


sim/midas/src/main/scala/midas/widgets/Assert.scala
```diff
@@ -29,8 +29,8 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
     val stallN = (!assertFire || resume)
     val dummyPredicate = true.B
```
zhemao, line 30: This isn't used anymore is it?

davidbiancolin: No i'll remove it.

PR #583, opened 2020-06-21
[WIP] Zynq Support

davidbiancolin:
Here i've done the initial task of getting bitstream builds working for the zedboard and ZC706. I'm opening this PR in case someone is ambitious enough to finish it off themselves while i work on my dissertation. I'm happy to provide guidance. 

This is mostly a code dump from midas-zynq, though i've scrubbed out some unnecessary abstraction from that repo and removed zybo. Sources live under:

`platforms/zynq/common` -> shared tcl and verilog
`plaforms/zynq/zedboard/base` -> the base project for zedboard
`plaforms/zynq/zc706/base` -> the base project for zc706

Like F1 you create a skeleton FPGA project with `make replace-rtl`, which like the F1 flow,  copies a base project  (e.g., `platforms/zynq/zc706/base`) into a working directory based on the design tuple (e.g., `platforms/zynq/zc706/cl_<TUPLE>`),  and dumps the generated RTL in there.

Typing `make <ARGS> fpga` will kick off a local vivado build. For zc706:
`make PLATFORM=zynq TARGET_CONFIG=FireSimSmallSystemConfig PLATFORM_CONFIG=ZC706Config fpga` 
For zedboard:
`make PLATFORM=zynq TARGET_CONFIG=FireSimSmallSystemConfig PLATFORM_CONFIG=ZedboardConfig BOARD=Zedboard fpga` 

A brain dump -- please pardon my grammar -- of the TODOs to get this merged in to dev consist of:
- [ ] Pipe clean driver builds 
  - I've not checked that this even works. With the xilinx-provided arm cross-compiler on your path things _should_ just work but there will likely be issues to work through.
  - firesim_top.{cc/h} might need to be reworked to suppose multiple hosts more easily. It's probably just easiest to put the main class definition and instantiation on a macro switch for now. 
  - it should be possible to run a simulation at this point so long as none of the bridges need DMA or they are disabled at runtime (these bridges are printf; NIC; tracerv/fireperf/fromajo) 
- [ ] DMA support. 
  - Expose one of the AXI4 master interfaces from the ARM subsystem. 
     - Generate the system.bd; reconfigure the ARM IP; export the bd back to tcl and check in
     - Probably want to drop a AXI4 width adapter here for now to match the DMA width of F1 (512 bits). I'm pretty confident the bridges are _not_ all robust against changes in DMA width atm.  
  - Connect DMA interface through midas_wrapper.v to the generated verilog. 
  - Update the ARM dts (and associated tasks) and extend the simif_zynq to use the new device
- [ ] Host frequency configuration
  - the zyng projects have a clocking.vh that could be generated by a custom FIRRTL pass.
    - something akin to [this](https://github.com/firesim/firesim/blob/master/sim/firesim-lib/src/main/scala/passes/EC2F1Artefacts.scala) but for Zynq
  - I would like to see the same PLL generation scripts reused across F1/Zynq but that is secondary to supporting the feature through the same platform configs

Other outstanding issues:
- [ ] Tooling for running simulations?
  - We use the manager to invoke simulators with their myriad plusargs. It might be worthwhile to have make commands that does this until we get manager support up. It could possibly reuse the plusArgs defined in makefrags for midas-level simulation. 
- [ ] Fireperf shared libraries
  - These are natively compiled and need to be cross-compiled instead. Maybe they shouldn't be built in build-setup.sh or build-setup needs to be more aware of available hosts.
  - I think it's fine for Fireperf to be disabled in the interim.
- [ ] AutoILA 
  - I don't think this is a critical feature to get right initially i think for Zynq we can just emit either TCL or `mark_debug` inline attributes with `mark_debug` once there is support for that in FIRRTL.
- [ ] FPGA-level simulation support
  - It would be nice if we weren't flying blind before going to FPGA.

Sagar has ideas for manager support that can follow later.


davidmetz:
Thanks a lot!
I won't get to it next week but will work on it the week after

sequencer:
If possible, I wanna have a try to take this PR over. Since I do got a ZYNQ board today, and really wanna to understand the internal of firesim.

sequencer:
Or is possible to use Firesim without ZYNQ, just communicate to external with PCI or Ethernet?

davidbiancolin:
> If possible, I wanna have a try to take this PR over. Since I do got a ZYNQ board today, and really wanna to understand the internal of firesim.

Hey @sequencer, sorry i never replied (i think i lost the email around xmas)  Are you still interested?

sequencer:
Hi @davidbiancolin, Willing to contribute!
I have some question and send via email, so just copy-paste it here.
I can purchase a US+ board, but is Arm SoC imperative? I have VU9P and KU040 and other big FPGA boards, but no US+ boards, is it possible to use FireSim with these boards? If arm core is compulsory, can you tell me why? and is there any workaround from this?

davidbiancolin:
> I can purchase a US+ board, but is Arm SoC imperative? I have VU9P and KU040 and other big FPGA boards, but no US+ boards, is it possible to use FireSim with these boards? If arm core is compulsory, can you tell me why? and is there any workaround from this?

An embedded arm core is not required, and in fact, i'd prefer not to support platforms that use the embedded arm. FireSim builds considerably more host-software than early forms of MIDAS did, and so cleanly managing builds for both x86 and ARM is a royal PITA that i'd rather not deal with right now. 

The VU9P is an ultrascale+ part (if that's what you mean by US+). I'm assuming you have a VCU118 if you have that FPGA and so that would be the board i'd suggest you use. To produce the most AWS-like environment you should plug that into a fairly beefy desktop over PCIE -- you'll use XDMA to interact with the board.

sequencer:
Yes, I have VCU118.
Sounds like the only host-fpga connection needed is XDMA?

davidbiancolin:
XDMA is the critical one. JTAG is almost certainly required too for programming and debug. Other than that i think you should be good.

davidbiancolin:
Were you planning on using FPGA shells?

sequencer:
I’m trying to remove dependency of rocketchip, by setting up a small dependency chain(diplomacy and tilelink/axi only), so I think i planning, at beginning, I won’t use FFGA shells.

sequencer:
Maybe porting a small design, something like GCD sound good ;p
Than I'll working on the stand-alone Diplomacy/TileLink project.

davidbiancolin:
The rocketchip dependency is pretty baked into the FireSim code base, so i wouldn't be wary of depending on it. 

Our shells to date use Vivado block diagrams, so that would be the best way to introduce a VCU118 shell without any new scala dependencies. Frankly, it's probably just the easiest thing to do in the short run.

sequencer:
Got it. I'll use fpga-shell + VCU118 to start my first run.
How to deal with the IOs? like UART and SPI memory?
And I think JTAG is required too. using Xilinx VIO to make them work?

sequencer:
Seems 
https://github.com/firesim/firesim/blob/7e32128bea5390ca528052dcdffb13656ee41162/sim/midas/src/main/scala/midas/platform/F1Shim.scala#L21
Is where I should start from?

davidbiancolin:
> Seems
> https://github.com/firesim/firesim/blob/7e32128bea5390ca528052dcdffb13656ee41162/sim/midas/src/main/scala/midas/platform/F1Shim.scala#L21
> 
> 
> Is where I should start from?

You have two choices:

1) Write a Shim (like hte one you linked) in which you use the FPGA shells stuff and instantiate IP directly from there. The Shim would then be the root of the module hierarchy

2) Follow the zynq example in this branch, and instantiate the generated verilog in a Vivado block diagram / or some top-level verilog etc... 

If you want to use diplomacy (and/or FPGA shells) (1) is probably the way to go.

sequencer:
I'll start from (1) :) Thank you!

PR #584, opened 2020-06-22
Bloop Support 

davidbiancolin:
See chipyard ucb-bar/chipyard#613.

On the firesim front bloop gives a 3x speed up on elaborating small designs (midasexamples) which is an enormous productivity win for me. Since managers aren't shared i think it might make sense to use bloop as the default when on F1.


sim/target-agnostic.mk
```diff
@@ -74,12 +74,12 @@ $(VERILOG) $(HEADER) $(fame_annos): $(FIRRTL_FILE) $(ANNO_FILE)
 conf: $(fame_annos)
 	mkdir -p $(GENERATED_DIR)
 	cd $(base_dir) && \
-	$(SBT) "project $(midas_sbt_project)" "runMain midas.stage.RuntimeConfigGeneratorMain \
+	$(call run_scala_main,$(firesim_sbt_project),midas.stage.RuntimeConfigGeneratorMain,\
 		-td $(GENERATED_DIR) \
-		-ggaf $(fame_annos) \
+		-faf $(fame_annos) \
```
albert-magyar, line 79: Did this just not get changed in #557 ?

scripts/machine-launch-script.sh
```diff
@@ -47,6 +47,12 @@ sudo yum -y install graphviz python-devel
 # used for CI
 sudo yum -y install expect
 
+# Optional: install bloop for fast scala builds on EC2 / CI
+curl -fLo coursier https://git.io/coursier-cli-linux &&
+sudo cp -f coursier /usr/local/bin
+sudo chmod 755 /usr/local/bin/coursier
```
albert-magyar, line 53: I think this is the only unmanaged install to a system-level path? Should we do this?

sim/target-agnostic.mk
davidbiancolin: I forgot to do the config generator. i should cherry-pick this onto master.

scripts/machine-launch-script.sh
davidbiancolin: You're right. I was thinking only from the perspective of running on EC2. How would you do it?

albert-magyar: Put it in `firesim/bloop-install` and put that on the path in `env.sh`?

albert-magyar: I'm also totally ignoring that this is in machine launch and not the setup script, so maybe just disregard that lol.

PR #588, opened 2020-06-25
Fix 'merging' of channels in memory model wrapper

albert-magyar:
When multiple channels merge without a queue, use the merging channels' valids to determine the ready signals that get sent back upstream without making any channel's ready depend on its valid.


albert-magyar:
This should be cherry-picked to master.

sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala
```diff
@@ -89,12 +89,15 @@ class ReadPort(val anno: ModelReadPort, val ports: Seq[Port]) extends IsMemoryPo
   def attachChannels(readCmd: Decoupled, readData: Decoupled): Seq[Statement] = {
     val iSignals = Seq(anno.addr, anno.en)
     val iValids  = iSignals.map(valid)
+    def iReady(i: ReferenceTarget) = {
+      And.reduce(readCmd.ready +: iSignals.filterNot(_ == i).map(valid))
```
davidbiancolin, line 93: Wait, there's a reduce method on `And`. Tight.

albert-magyar: You were already using it, lol

PR #589, opened 2020-06-25
Fix 'merging' of channels in memory model wrapper (bp #588)

albert-magyar:
@davidbiancolin how should we backport this?


davidbiancolin:
I think this is fine. I still need to figure out a consistent back-porting scheme.

PR #590, opened 2020-06-28
Fix ILATopWiring transform

zhemao:


sim/firesim-lib/src/main/scala/passes/ILATopWiring.scala
```diff
@@ -131,6 +130,15 @@ class ILATopWiringTransform extends Transform {
     state
   }
 
+  override def prerequisites = Forms.MidForm
+  override def optionalPrerequisites = Seq.empty
+  override def optionalPrerequisiteOf = Forms.MidEmitters
```
davidbiancolin, line 135: This should probably be Forms.HighEmitters because i think for the purposes of debugging GG emission we always want this pass to run, even if we specify high-form output. @albert-magyar am i understanding this correctly?

albert-magyar: Nothing you list in `optionalPrerequisiteOf` will change whether or not _this_ transform runs or not.

davidbiancolin: Because if it's listed as a target it always runs -- full stop?

albert-magyar: And the inverse -- if it's not listed as a target, the fact that it's listed as an `optionalPrerequisiteOf` _will not make it run_.

Not knowing much about this transform, I think listing any emitter under `optionalPrerequisiteOf` would be no more useful than listing `Nil`.

davidbiancolin: Can you have multiple emitters? Would it then be possible for a high form emitter to drop something that is lacking the result of this target? Or all targets guaranteed to run before all emitters?

PR #595, opened 2020-07-07
Add decoupling queue to hPort of AssertBridgeModule

timsnyder-siv:
Alleviates timing pressure in designs with large numbers of assertions.
Should probably make queue depth a parameter.  Hard coded to 2 for a start.

### DIFF
--- a/sim/midas/src/main/scala/midas/widgets/Assert.scala
+++ b/sim/midas/src/main/scala/midas/widgets/Assert.scala
@@ -22,7 +22,7 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
     val hPort = IO(HostPort(Input(UInt(numAsserts.W))))
     val resume = WireInit(false.B)
     val cycles = RegInit(0.U(64.W))
-    val q = Module(new Queue(hPort.hBits.cloneType,2))
+    val q = Module(new Queue(hPort.hBits.cloneType, 2))
     q.io.enq.valid := hPort.toHost.hValid
     hPort.toHost.hReady := q.io.enq.ready
     q.io.enq.bits := hPort.hBits
--- a/sim/midas/src/main/scala/midas/widgets/Assert.scala
+++ b/sim/midas/src/main/scala/midas/widgets/Assert.scala
@@ -22,23 +22,29 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
     val hPort = IO(HostPort(Input(UInt(numAsserts.W))))
     val resume = WireInit(false.B)
     val cycles = RegInit(0.U(64.W))
-    val asserts = hPort.hBits
+    val q = Module(new Queue(hPort.hBits.cloneType,2))
+    q.io.enq.valid := hPort.toHost.hValid
+    hPort.toHost.hReady := q.io.enq.ready
+    q.io.enq.bits := hPort.hBits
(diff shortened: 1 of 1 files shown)


sim/midas/src/main/scala/midas/widgets/Assert.scala
```diff
@@ -22,23 +22,29 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
...
-    // We only sink tokens, so tie off the return channel
-    hPort.fromHost.hValid := true.B
+    q.io.deq.ready := tFireHelper.fire(q.io.deq.valid)
     when (targetFire) {
       cycles := cycles + 1.U
     }
 
     genROReg(assertId, "id")
-    genROReg(assertFire && hPort.toHost.hValid, "fire")
+    genROReg(assertFire && q.io.deq.valid, "fire")
```
timsnyder-siv, line 47: When I tested this, the code was still `genROReg(assertFire, "fire")`  I think the correct thing to do is use the `deq` side `valid` from the queue but I haven't actually run with this and I'm not completely clear on what looks at this 'fire' register.

sim/midas/src/main/scala/midas/widgets/Assert.scala
```diff
@@ -22,23 +22,29 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
...
+
+    val asserts = q.io.deq.bits
+
     val assertId = PriorityEncoder(asserts)
     val assertFire = asserts.orR
 
     val stallN = (!assertFire || resume)
 
-    val tFireHelper = DecoupledHelper(hPort.toHost.hValid, stallN)
+    val tFireHelper = DecoupledHelper(q.io.deq.valid, stallN)
```
timsnyder-siv, line 39: When I ran the tests for this change, I still had `dummyPredicate` as a third argument to `DecoupledHelper()` and had parens on `tFireHelper.fire()` below but I don't expect those two changes coming from upstream will break anything here.

timsnyder-siv:
@davidbiancolin , I was using this Queue successfully with it based off of 1.10.  I rebased forward to dev to open this PR today but I haven't yet rerun the tests.  Feel free to mark this a draft until I update with test results if you'd like them rerun.  Just let me know.

sim/midas/src/main/scala/midas/widgets/Assert.scala
```diff
@@ -22,23 +22,29 @@ class AssertBridgeModule(assertMessages: Seq[String])(implicit p: Parameters) ex
     val hPort = IO(HostPort(Input(UInt(numAsserts.W))))
     val resume = WireInit(false.B)
     val cycles = RegInit(0.U(64.W))
-    val asserts = hPort.hBits
+    val q = Module(new Queue(hPort.hBits.cloneType,2))
```
davidbiancolin, line 25: Nit: Space between comma and additional parameters.

timsnyder-siv: Added space in fixup commit.

davidbiancolin:
Alright, now CI should run on this and sanity check the assertion tests still work.

sim/midas/src/main/scala/midas/widgets/Assert.scala
davidbiancolin: Yeah it should be fine.

PR #604, opened 2020-07-15
Mergify-based Automatic Backporting Support

davidbiancolin:
PRs made to dev can be labeled with "please backport", and Mergify will create a new PR against master. Developers will still need manually regenerate AGFIs (if necessary) and merge the backport PR. 

This was shamelessly ripped from the FIRRTL repo.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,19 @@
+pull_request_rules:
+  - name: backport to master
+    conditions:
+      - merged
+      - base=dev
+      - label="please backport"
+    actions:
+      backport:
+        branches:
+          - master
+        ignore_conflicts: True
+        label_conflicts: "bp-conflict"
+
+  - name: label Mergify backport PR
+    conditions:
+      - body~=This is an automated backport of pull request \#\d+ done by Mergify
+    actions:
+      label:
+        add: [backported]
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,19 @@
...
+          - master
+        ignore_conflicts: True
+        label_conflicts: "bp-conflict"
+
+  - name: label Mergify backport PR
+    conditions:
+      - body~=This is an automated backport of pull request \#\d+ done by Mergify
+    actions:
+      label:
+        add: [backported]
```
albert-magyar, line 19:
```suggestion
        add: [backport]
```

This goes on the PR to master

davidbiancolin: I was worried users would apply this to their dev branch since the label is what triggers the BP here versus the milestone. Maybe i'll change it to "is backport".

albert-magyar:
You can also do this whole process by just commenting on a PR like this, so maybe the rules aren't necessary. Mergify has to be enabled, and it won't work on PRs that modify `.mergify.yml`.

Also, changes to `.mergify.yml` should probably always go to master.

@Mergifyio backport master

davidbiancolin:
I'm going to leave the tag in for now. We can remove it later if it's confusing or burdensome.

PR #611, opened 2020-07-15
Fix TracerV Trigger Generation When Tracing Is Disabled

### LINKED ISSUE #600: Decouple TracerV-enable and Trigger-enable
In the current FireSim version, in order to use any bridge that may depend on a trigger, the TracerV bridge driver needs to be enabled.
I previous versions of FireSim, the trigger could be used in any AGFI that had the TracerV widget synthesized (all of them, by default), without having the trace written out to disk.

As a result, in order to use Autocounter in the current version of FireSim, the user must also output a tracefile (which really slows down simulation, because of writing the file to disk).

The rational behind the change was that TracerV was hurting FMR when it was not being used, since it was still offloading token to the host despite them not being written to disk.
The potential solution is to separate trigger-enable and tracer-enable

davidbiancolin:
In a regression introduced by multiclock support, if you weren't collecting an instruction trace with TracerV it's trigger sources would be disabled, which would effectively disable AutoCounter. This PR fixes that by separating trace disable from the trigger specification. 

Resolves #600. Perhaps a better longer term solution might be to pull the trigger source handling out of tracerV entirely.

### DIFF
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
@@ -155,22 +155,20 @@ tracerv_t::~tracerv_t() {
 
 void tracerv_t::init() {
     if (!this->trace_enabled) {
-      // Explicitly disable token collection inthe bridge by only collecting
-      // tokens from cycle 0 to cycle 0, saving DMA bandwidth and improving FMR
-      write(this->mmio_addrs->triggerSelector, 1);
-      write(this->mmio_addrs->hostTriggerCycleCountStartHigh, 0);
-      write(this->mmio_addrs->hostTriggerCycleCountStartLow, 0);
-      write(this->mmio_addrs->hostTriggerCycleCountEndHigh, 0);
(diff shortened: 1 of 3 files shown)


sim/firesim-lib/src/main/cc/bridges/tracerv.cc
```diff
@@ -188,15 +186,15 @@ void tracerv_t::init() {
...
               this->trigger_start_insn, this->trigger_start_insn_mask,
               this->trigger_stop_insn, this->trigger_stop_insn_mask);
     }
     else
     {
-      // Biancolin: should we not error here?
+      // Writing 0 to triggerSelector permanently enables the trigger
       write(this->mmio_addrs->triggerSelector, this->trigger_selector);
-      printf("TracerV: No trigger selected. Collecting trace from %lu to %lu cycles\n", 0, ULONG_MAX);
+      printf("TracerV: No trigger selected. Enable trigger enabled from %lu to %lu cycles\n", 0, ULONG_MAX);
```
alonamid, line 197: "Enable trigger enabled"?

PR #612, opened 2020-07-16
Make peek/poke tests more reliable for testing Golden Gate simulators

albert-magyar:
This fixes a few outstanding issues:
- Ensures that the tokens for the current cycle have arrived for all outputs before executing an `expect`
- Adds a pound delay to reset across the DPI boundary in VCS. This makes host reset deassertion behave more sensibly.

There still is the question of whether to add this constraint to peeks. I think a case could be made either way. Regardless, it's still possible to write peek/poke tests that have a broken underlying model of how test stimuli propagate, since the peek/poke bridge has different semantics from peek/poke Chisel tests, which let you do multiple combinationally propagated pokes per input per cycle.

### DIFF
--- a/sim/midas/src/main/cc/simif.h
+++ b/sim/midas/src/main/cc/simif.h
@@ -47,6 +47,9 @@ class simif_t
     }
     virtual void load_mem(std::string filename);
 
+    bool peekpoke_ready() { return read(this->defaultiowidget_mmio_addrs->READY); }
+    bool precise_peekable() { return read(this->defaultiowidget_mmio_addrs->PRECISE_PEEKABLE); }
+
   public:
     // Simulation APIs
     virtual void init(int argc, char** argv, bool log = false);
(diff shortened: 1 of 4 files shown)


albert-magyar:
> What's the rationale for plumbing out precise peekable to the master as opposed to adding it to the peek poke bridge directly?

Just to be consistent with `idle`/`done`, which is the most analogous extant thing.

davidbiancolin:
For some context, it's like that because the peek poke widget has long been the arbiter of simulation time by essentially controlling the flow of reset tokens. This doesn't really make sense anymore under multiclock, but i'll decouple peek/poke from the master in a later PR. (I've changed it in my dissertation branch)

albert-magyar:
> For some context, it's like that because the peek poke widget has long been the arbiter of simulation time by essentially controlling the flow of reset tokens. This doesn't really make sense anymore under multiclock, but i'll decouple peek/poke from the master in a later PR. (I've changed it in my dissertation branch)

If `precisePeekable` is directly visible from the peek/poke widget, then I would think that `done` should also be. I can certainly change this PR to do that.

sim/midas/src/main/cc/simif.h
```diff
@@ -41,12 +41,30 @@ class simif_t
...
 
+    std::string blockingFail = "The test environment has starved the simulator, preventing forward progress.";
+
     inline void take_steps(size_t n, bool blocking) {
       write(this->master_mmio_addrs->STEP, n);
       if (blocking) while(!done());
     }
     virtual void load_mem(std::string filename);
 
+    bool waitOn(size_t flag_addr, double timeout) {
```
davidbiancolin, line 52: ultra nit: maybe snake case the function names.

albert-magyar: Good call, gotta be C++-ing now :)

PR #614, opened 2020-07-18
Add unit tests for Golden-Gate-optimized simulators

albert-magyar:
This PR adds four unit tests for Golden Gate optimizations, including model multi-threading. In order to support these tests, the multi-threading pass is added to `MidasTransforms`, but it will function as an identity transform unless the appropriate annotations are added.


davidbiancolin:
> This PR adds four unit tests for Golden Gate optimizations, including model multi-threading. In order to support these tests, the multi-threading pass is added to `MidasTransforms`, but it will function as an identity transform unless the appropriate annotations are added.

But does it mean that it will function correct if you do add the annotations? (i.e., can people try to use it?)

albert-magyar:
> But does it mean that it will function correct if you do add the annotations?

Yes

PR #617, opened 2020-08-01
Make memory loading compatible with new testchipip API

zhemao:
I changed the way loadmem works in testchipip with this PR. 

https://github.com/ucb-bar/testchipip/pull/94

So this changes firesim to be compatible. As a helpful side-effect, it also allows firesim software sim to properly simulate multiple host channels.


zhemao:
Yes, it loads the same thing into each memory channel. This works out because of the way addressing is done.

PR #619, opened 2020-08-05
Allow synthesized asserts to be disabled at runtime

zhemao:


sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -34,6 +36,9 @@ firesim_top_t::firesim_top_t(int argc, char** argv)
         if (arg.find("+zero-out-dram") == 0) {
             do_zero_out_dram = true;
         }
+        if (arg.find("+disable-asserts") == 0) {
```
davidbiancolin, line 39: This should be handled in the assert bridge driver like plus args for other bridges. Remove the enable constructor parameter as well.

deploy/sample-backup-configs/sample_config_runtime.ini
```diff
@@ -57,6 +57,10 @@ suffixtag=
 # In general, this is not required to produce deterministic simulations on
 # target machines running linux. Enable if you observe simulation non-determinism.
 zerooutdram=no
+# For designs built with synthesized assertions, enabling (=yes) prints
+# assertion message and terminates simulation if assertion fires.
+# Disabling (=no) ignores assertion firing and continues simulation.
+asserts=yes
```
davidbiancolin, line 63: make this more verbose: `capture_synthesized_asserts` or perhaps reversing the polarity would make it more intuitive: `ignore_synthesized_asserts` and matches the variable names in python. 

@timsnyder-siv What name makes sense to you?

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -389,56 +394,64 @@ firesim_top_t::firesim_top_t(int argc, char** argv)
     add_bridge_driver(new synthesized_assertions_t(this,
```
davidbiancolin, line 394: As above, you can delete the enable asserts parameter once you move the plusarg handling into the bridge driver.

timsnyder:
Thanks for posting this.  It will be very helpful.

davidbiancolin:
This requires an AGFI regeneration.

deploy/sample-backup-configs/sample_config_runtime.ini
timsnyder: I think matching the polarity in the python makes sense.  I know some people have a strong dislike of `negative_name=true` semantics.  If you want a more verbose name, something like `asserts_terminate_sim=yes` might make more sense than `asserts=yes`.

`ignore_synthesized_asserts` is okay but I would wonder whether it would still print the message (when I didn't have the nice comment in the sample config).

deploy/sample-backup-configs/sample_config_runtime.ini
```diff
@@ -57,6 +57,10 @@ suffixtag=
 # In general, this is not required to produce deterministic simulations on
 # target machines running linux. Enable if you observe simulation non-determinism.
 zerooutdram=no
+# For designs built with synthesized assertions, enabling (=no) prints
```
davidbiancolin, line 60: Replacing "enabling (=no)" with "setting disable_synth_asserts=no" and later "Setting disable_synth_asserts=yes" makes this read more intuitively.

PR #623, opened 2020-08-27
Update AGFIs for new diplomatically clocked designs

jerryz123:


jerryz123:
Because the Chipyard divided clock config, which the firesim config inherits, is singlecore

davidbiancolin:
> Because the Chipyard divided clock config, which the firesim config inherits, is singlecore

We're remote so i can't sassily shake my head. Anyways, you're good to go.

davidbiancolin:
You still have to bump CY though...

PR #626, opened 2020-09-14
Waveform script

t14916:
Adding waveform scripts to reverse decoupling of host and target clock in firesim, as well as querying values from waveform.


davidbiancolin:
Subsumed by #642.

PR #630, opened 2020-09-22
[docs] Update config_runtime.ini example with profileinterval setting

albert-magyar:
The example `config_runtime.ini` that's included inline in the single-node docs is currently broken. This fixes that.


davidbiancolin (approved):
LGTM

PR #631, opened 2020-09-24
ILADepth override needs to be left-of BaseF1Config

timsnyder:
There isn't something special about this, the override needs to be to the left of `BaseF1Config` correct?


timsnyder:
@davidbiancolin this is a small change that I noticed while reading the docs this afternoon.  Hopefully I'm correct here.

davidbiancolin:
You are correct. Always air on the side of prefixing the config string. Appending will only produce a change if the key has not already been defined in the left side of the tree.

davidbiancolin (approved):
LGTM

PR #632, opened 2020-09-25
Support for Mirroring Chipyard's Idealized PLL

davidbiancolin:
See ucb-bar/chipyard#676

This no longer assumes the base clock is implicit when instantiating the ClockBridge. At least one clock must have a unity clock ratio --> that clock is assumed to be the base clock. (Previously, this unity-ratio clock was assumed to be the zeroth element of the clock vector). 

The old API has been pushed into a companion object.


albert-magyar (approved):
LGTM

PR #635, opened 2020-09-30
rm FpgaMMIOSize, instead enforce CtrlNastiKey.addrBits

timsnyder-siv:
Silent truncation of ctrl addresses is not needed.  Instead
add an assertion that ensures the design has not exceeded
the number of MMIO registers addressable on the ctrl bus.

### DIFF
--- a/sim/midas/src/main/scala/midas/Config.scala
+++ b/sim/midas/src/main/scala/midas/Config.scala
@@ -59,7 +59,6 @@ class SimConfig extends Config (new Config((site, here, up) => {
   case SynthAsserts     => false
   case SynthPrints      => false
   case DMANastiKey      => NastiParameters(512, 64, 6)
-  case FpgaMMIOSize     => BigInt(1) << 12 // 4 KB
   case AXIDebugPrint    => false
 
   // Remove once AXI4 port is complete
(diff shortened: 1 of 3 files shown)


timsnyder-siv:
@davidbiancolin here's my cut at fixing the truncation we were debugging last week.  I tested this quickly by hacking CtrlNastiKey.addrBits to only be 8 and running the default FireSimRocketConfig through GoldenGate to make sure the assertion fired.

Is there a place I could also add a scala unit test for this or anything else you think it needs?

davidbiancolin:
I think having a scala unit test to check this would be great but we don't really have an existing designs to check for compilation failures so i won't ask for one.

If you wanted to do it, i'd write a dummy target module that instantiated K autocounters, and host it with the other midasexamples. You wouldn't need to add a driver or anything. I'd then add a scalatest with the rest of the midas examples that attempted to compile that target design, but you'd wrap the gg compilation with one of scalatest's expect failure methods.  That's probably going to require some refactoring of the parent class [in here](https://github.com/firesim/firesim/blob/master/sim/src/test/scala/midasexamples/TutorialSuite.scala). 

What you could do is add two traits that expect to be mixed into TutorialSuite.

One would pull that default stuff out of the parent class (lines 90/91): 
 
trait RunsDefaultMLSim {
    runTest(backendSimulator)
}

another would be:

trait ExpectsCompilationFailure {
   <scala test expect failure magic> <invoke compilation>
}

Otherwise this LGTM.

davidbiancolin (approved):

timsnyder-siv:
Since there isn't an existing place to add a test, I'll save adding a test until after I'm done submitting PR's for the other things I need.  I merged `dev` into my branch so that the CI would run again for you before merging this.

timsnyder-siv:
@davidbiancolin this looks ready for merging when you have a chance.

PR #636, opened 2020-09-30
Add platform config options to selectively enable model multithreading

albert-magyar:


davidbiancolin (approved):
This LGTM. Could you add a note to the dev -> master PR. 

Could you also write a small doc page about the available optimizations and how to deploy them in a target?

Z-Y00:
Hi,
I'm confused about where to add the [firesim-boom-mem-opt] as described in your doc.
https://docs.fires.im/en/latest/Golden-Gate/Resource-Optimizations.html

Could you explain it a little bit more?
It would be much easier to understand if you can post some example configure file as in the Tutorial here:
https://docs.fires.im/en/latest/Running-Simulations-Tutorial/Running-a-Single-Node-Simulation.html

PR #638, opened 2020-09-30
Add support for target designs with SerialBridge but no DRAM

jerryz123:
Support target designs with no AXI4 backing DRAM. The changes modify firesim_fesvr to not use the loadmem channel if it is unavailable.

There will be a corresponding Chipyard PR when millennium is back.

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/SerialBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/SerialBridge.scala
@@ -10,16 +10,18 @@ import freechips.rocketchip.config.Parameters
 
 import testchipip.{SerialIO, SerialAdapter}
 
-class SerialBridge(memoryRegionName: String) extends BlackBox with Bridge[HostPortIO[SerialBridgeTargetIO], SerialBridgeModule] {
+case class SerialBridgeParams(memoryRegionNameOpt: Option[String])
+
+class SerialBridge(memoryRegionNameOpt: Option[String]) extends BlackBox with Bridge[HostPortIO[SerialBridgeTargetIO], SerialBridgeModule] {
   val io = IO(new SerialBridgeTargetIO)
   val bridgeIO = HostPort(io)
-  val constructorArg = Some(memoryRegionName)
+  val constructorArg = Some(SerialBridgeParams(memoryRegionNameOpt))
   generateAnnotations()
 }
 
 object SerialBridge {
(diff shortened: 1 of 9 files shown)


sim/midas/src/main/scala/midas/widgets/UsesHostDRAM.scala
```diff
@@ -49,7 +49,9 @@ trait HostDramHeaderConsts {
    * to give it an independent region.
    *
    */
-  def memoryRegionName: String
+  def memoryRegionNameOpt: Option[String]
```
davidbiancolin, line 52: Can you explain the rationale for this change. It has knock on effects for DRAM allocation.

jerryz123: The `HostDramHeaderConsts` trait is used by `SerialBridgeModule`, which does not have a `memoryRegionName` if there is no target DRAM.

davidbiancolin: I think in that case what we should do is remove the mixin from the serial bridge; and directly add an option member there. 

If we want bridges to be able to inspect the dram allocation map, i think we can push something into the parameters instance before FPGATop module instantiation.

davidbiancolin: Ill wait for the rest of the CY PR though.

sim/firesim-lib/src/main/scala/bridges/SerialBridge.scala
```diff
@@ -10,16 +10,18 @@ import freechips.rocketchip.config.Parameters
 
 import testchipip.{SerialIO, SerialAdapter}
 
-class SerialBridge(memoryRegionName: String) extends BlackBox with Bridge[HostPortIO[SerialBridgeTargetIO], SerialBridgeModule] {
+case class SerialBridgeParams(memoryRegionNameOpt: Option[String])
```
davidbiancolin, line 13: Can you add scaladoc for this explaining how it's used.

davidbiancolin (approved):
LGTM. Remember to add the dev->master PR message.

sim/midas/src/main/scala/midas/widgets/UsesHostDRAM.scala
jerryz123: CY PR is here: https://github.com/ucb-bar/chipyard/pull/682

PR #639, opened 2020-09-30
Improve QoR when threading models with SyncReadMems

albert-magyar:
Currently, synchronous-read memories get processed with `VerilogMemDelays` before FAME5 multi-threading, which leads to inconsistent BRAM inference and generally worse QoR. This PR fixes that by generating specialized implementations of natively multi-threaded synchronous-read memories. These implementations generally use RAM resources more effectively than the baseline unoptimized simulators.

This also copies in the fix to `VerilogMemDelays` from freechipsproject/firrtl#1908. This fix can be reverted when the FIRRTL upstream changes make it to FireSim, but since that will take a while, it's good to get this bugfix in.

### DIFF
--- a/sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
+++ b/sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
@@ -90,6 +90,7 @@ private[midas] class MidasTransforms extends Transform {
       new fame.EmitFAMEAnnotations("post-fame-transform.json"),
       new ResolveAndCheck,
       fame.MultiThreadFAME5Models,
+      fame.ImplementThreadedSyncReadMems,
       new ResolveAndCheck,
       new fame.EmitAndWrapRAMModels,
       new EmitFirrtl("post-gen-sram-models.fir"),
--- a/sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
+++ b/sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
@@ -90,7 +90,6 @@ private[midas] class MidasTransforms extends Transform {
       new fame.EmitFAMEAnnotations("post-fame-transform.json"),
       new ResolveAndCheck,
       fame.MultiThreadFAME5Models,
-      fame.ImplementThreadedSyncReadMems,
       new ResolveAndCheck,
       new fame.EmitAndWrapRAMModels,
       new EmitFirrtl("post-gen-sram-models.fir"),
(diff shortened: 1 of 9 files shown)


sim/midas/src/main/scala/midas/passes/fame/ImplementThreadedSyncReadMems.scala
```diff
@@ -0,0 +1,169 @@
...
+  def mapString(f: String => String): Statement = this.copy(name = f(name))
+  def mapInfo(f: Info => Info): Statement = this.copy(info = f(info))
+  def foreachExpr(f: Expression => Unit): Unit = ()
+  def foreachStmt(f: Statement => Unit): Unit = ()
+  def foreachType(f: Type => Unit): Unit = f(dataType)
+  def foreachString(f: String => Unit): Unit = f(name)
+  def foreachInfo(f: Info => Unit): Unit = f(info)
+}
+
+object ImplementThreadedSyncReadMems extends Transform {
```
davidbiancolin, line 57: Could you add the DependencyAPIMigration trait to this.

sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
```diff
@@ -89,6 +89,7 @@ private[midas] class MidasTransforms extends Transform {
       new fame.EmitFAMEAnnotations("post-fame-transform.json"),
       new ResolveAndCheck,
       fame.MultiThreadFAME5Models,
+      fame.ImplementThreadedSyncReadMems,
```
davidbiancolin, line 92: Could you put a comment here explaining the introduction of non-standard IR nodes and how they are handled by this pass ? Or maybe it makes sense to bundle these into a single wrapper SeqTransform?

davidbiancolin (approved):
Needs a bump, but LGTM.

sim/midas/src/main/scala/midas/passes/MidasTransforms.scala
albert-magyar: I changed it from a transform to just a utility, so now it's all contained within a single transform.

sim/midas/src/main/scala/midas/passes/VerilogMemDelays.scala
```diff
@@ -0,0 +1,190 @@
+// See LICENSE for license info.
```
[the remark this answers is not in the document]
albert-magyar: Yeah I'll do that when I merge this.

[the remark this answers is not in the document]
albert-magyar: Opened #655

PR #641, opened 2020-10-02
Unify fesvr with chipyard

jerryz123:
See https://github.com/ucb-bar/chipyard/pull/683


sim/firesim-lib/src/main/cc/bridges/serial.cc
```diff
@@ -43,20 +43,20 @@ serial_t::serial_t(simif_t* sim, const std::vector<std::string>& args, SERIALBRI
         }
     }
 
-    argv_arr = new char*[args_vec.size()];
+    argv_arr = new char*[args_vec.size()+1];
+    argv_arr[0] = (char*) "firesim-fesvr";
```
colinschmidt, line 47: Why not just prepend this to args_vec and delete all these ugly `+1`s

sim/firesim-lib/src/main/cc/fesvr/firesim_fesvr.cc
```diff
@@ -1,175 +1,74 @@
...
-    fesvr->run();
-
-    while (true)
-        fesvr->target->switch_to();
-
-    return 0;
-}
-
-firesim_fesvr_t::firesim_fesvr_t(const std::vector<std::string> args, bool has_loadmem) : htif_t(args), has_loadmem(has_loadmem)
+firesim_fesvr_t::firesim_fesvr_t(int argc, char** argv, bool can_have_loadmem) : testchip_fesvr_t(argc, argv, has_loadmem), is_busy(false)
```
colinschmidt, line 5: This is a strange style where some of the variable inits are up here but `has_loadmem(can_have_loadmem)` isn't.

sim/firesim-lib/src/main/cc/fesvr/firesim_fesvr.h
```diff
@@ -2,89 +2,50 @@
...
-#define FESVR_ADDR_CHUNKS 2
-#define FESVR_LEN_CHUNKS 2
-
-// Wraps read/write requests that will be shortcircuited through the LOADMEM widget
-struct fesvr_loadmem_t {
-    fesvr_loadmem_t(): addr(0), size(0) { }
-    fesvr_loadmem_t(size_t addr, size_t size): addr(addr), size(size) { }
-    size_t addr;
-    size_t size;
+struct firesim_loadmem_t {
```
colinschmidt, line 7: Nit: Is there not just a standard c++ tuple construct? Maybe just typedef that instead of creating your own version?

colinschmidt (approved):
LGTM. Just a few of nits.

sim/firesim-lib/src/main/cc/fesvr/firesim_fesvr.cc
jerryz123: `has_loadmem` is a property of the parent `testchip_fesvr_t` class, so it can't be added to the initialization list here.

For SW sim, `loadmem` is only enabled if the `+loadmem` flag is seen, so `testchip_fesvr_t` only sets `has_loadmem` if `can_have_loadmem` and `+loadmem` is present.

For FireSim, there is no `+loadmem` flag, so we have to override the behavior in `testchip_fesvr_t` and always set `has_loadmem` if `can_have_loadmem`

sim/firesim-lib/src/main/cc/fesvr/firesim_fesvr.h
jerryz123: This is just the old `fesvr_loadmem_t` renamed. I don't really want to change how this works here.

PR #644, opened 2020-10-15
Place FASED in a Seperate Clock Domain By Default

davidbiancolin:
Coupled with the CY bump this puts the memory bus in an asynchronous domain running at 1 GHz (the rest of the system still runs at 3.2 GHz), fixing a feature regression over MIDAS (and radically improving it). 

Related PRs: 
ucb-bar/chipyard#690
ucb-bar/testchipip#106

See:

https://groups.google.com/g/firesim/c/PQZOCKh681k/m/jlqzH3ZkCQAJ
https://groups.google.com/g/firesim/c/PjTQXR5yreA/m/1Vb5tFU1BAAJ

The fast clock does not fire in every timestep in this design which exposed a bug in the FAME1 xform. I've pulled in a change from my dissertation branch that coincidentally fixes this, and relaxes the requirement that inputs and outputs from a model must be latched and launched by state in the same domain of the hub model. A updated but still overly strict requirement is that output channels can only be combinationally coupled to inputs in the same domain (there are cases where this is OK...). 

This PR reveals an FMR regression (where models are being promoted even when multithreaded is disabled) that will be fixed in a separate PR.


sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -147,34 +147,63 @@ object FAMEModuleTransformer {
...
+     * combinational paths resolve based on initial tokens and register values.
+     * We can spoof this this by setting clockEnable for all domains without
+     * actually ungating those clocks.
+     */
+    val doneInit = hostFlagReg(ns.newName("doneInit"), resetVal = UIntLiteral(0))
+    val doneInitConnect = Connect(NoInfo, WRef(doneInit), Or(WRef(doneInit), WRef(finishing)))
+
+    case class TargetClockMetadata(
+      targetSourcePort: Port,
+      // I'm not convinced these are the right names for this...
```
timsnyder, line 161: I'm not convinced on the names either.  Are `clockLow` and `clockHigh` `negedge` and `posedge` with separate gating enable signals for a single target clock domain or is it something completely different?

sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -147,34 +147,63 @@ object FAMEModuleTransformer {
...
         // TODO: this should not include muxes in connectivity!
         val srcClockPorts = portDeps.getEdges(clock.name).map(portsByName(_))
         assert(srcClockPorts.size == 1)
         val clockRef = WRef(srcClockPorts.head)
-        val clockFlag = DoPrim(PrimOps.AsUInt, Seq(clockChannel.replacePortRef(clockRef)), Nil, BoolType)
+        val clockFlag = if (isInput) {
+          DoPrim(PrimOps.AsUInt, Seq(clockHighMap(clockRef)), Nil, BoolType)
+        } else {
+          DoPrim(PrimOps.AsUInt, Seq(clockLowMap(clockRef)), Nil, BoolType)
+        }
```
timsnyder, line 206: Yeah, I really don't understand what `clockHigh` and `clockLow` are and why input channels are High and output channels are Low.  Does the FAME transformed model sample inputs on `posedge` and latch new outputs on `negedge`?

I apologize for starting to learn about the transform guts and asking stupid newb questions in your excellent PR that I'm excited to be able to use.

davidbiancolin: No problem. The names derive from my somewhat poor understanding of how the old Chisel CPP backend worked, they should be renamed. Nothing happens on a clock negedge.

The resulting FAME 1-transformed is essentially a two cycle pipeline (or three depending on how you slice it), that acts on consecutive simulation time steps. The enables define when the a channel will be allowed to handshake again in the next host cycle (i.e., they reset their FiredReg). Essentially the FiredReg is the state for a single-bit FSM on each channel. 

In the first stage, clock tokens are unpacked and fanned out to all input channels according to their clock domain. When a cycle moves from stage 1 -> stage 2, firedRegs for all input channels in clock domains marked to fire in that clock token (essentially a timestep) are reset. 

In the second stage, the clock token is again fanned out, this time to all output channels. (The fired registers for output channels form the second set of pipeline registers). When a cycle moves from stage 2 -> stage 3, a target clock edge for that domain is launched, and output channel FiredRegs are reset.

In stage 3, combinational paths for the freshly fired clock-domain resolve. Setting their FiredRegs as handshakes occur. This will take multiple host cycles if a combinational path spans multiple models. 

That's a brain dump. Hopefully it makes some sense. :|

"Finishing" doesn't really imply a target cycle is finishing anymore so much as it means that handshakes on all output channels and all input channels are happening or have happened in the current host cycle. See `isFiredOrFiring`. In practice, since there is a infinite stream of clock tokens, the only source of stalls are the late arrival of input tokens and backpressure from output channels. Combinational paths manifest as both.  If we think of the following timesteps occupying the pipeline:

Stage1  |  Stage2 | Stage3
Time3   |  Time2   | Time1

Time1 generates a stall if:
1) An output channel is backpressuring (<channel_name)_source_ready is low) 
2) An output channel is combinationally dependent on an input channel that has not yet provided it's token (<CCDep channel name>_sink_valid is low). 
Time2 generates a stall if:
- Not all input tokens for domains scheduled to fire at Time 2 are available. Once the edge fires, these tokens should no longer be observable.

albert-magyar: In Chisel2-generated-C++-emulator parlance, "clock high" essentially means the phase where posedge-triggered events happen, and "clock low" means triggering the evaluation of the continuous assignments of combinational logic. The reliance on straight-line, topologically sorted code meant each was a very large basic block, so somehow these names were assigned to those blocks (way, way) back in the day.

Anyhow, I don't think they map onto what's being described here very well.

sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -147,35 +147,82 @@ object FAMEModuleTransformer {
...
+      targetSourcePort: Port,
+      outputChannelEnable: Expression,
+      inputChannelEnable: Expression,
+      clockBuffer: SignalInfo)
+
     // Multi-clock management step 4: Generate clock buffers for all target clocks
-    val targetClockBufs: Seq[SignalInfo] = clockChannel.ports.map { en =>
-      val enableReg = hostFlagReg(s"${en.name}_enabled", resetVal = UIntLiteral(1))
+    val clockMetadata: Seq[TargetClockMetadata] = clockChannel.ports.map { en =>
+      val enableReg = hostFlagReg(s"${en.name}_enabled")//, resetVal = UIntLiteral(1))
```
albert-magyar, line 173:
```suggestion
      val enableReg = hostFlagReg(s"${en.name}_enabled")
```

Leaving it commented out makes it look like you're aiming for "don't care," when I believe you are actually relying on the change to the default of zero?

sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -147,35 +147,82 @@ object FAMEModuleTransformer {
...
       val buf = WDefInstance(ns.newName(s"${en.name}_buffer"), DefineAbstractClockGate.blackbox.name)
       val clockFlag = DoPrim(PrimOps.AsUInt, Seq(clockChannel.replacePortRef(WRef(en))), Nil, BoolType)
       val connects = Block(Seq(
         Connect(NoInfo, WRef(enableReg), Mux(WRef(finishing), clockFlag, WRef(enableReg), BoolType)),
         Connect(NoInfo, WSubField(WRef(buf), "I"), WRef(hostClock)),
-        Connect(NoInfo, WSubField(WRef(buf), "CE"), And(And(WRef(enableReg), WRef(finishing)), nReset))))
-      SignalInfo(Block(Seq(enableReg, buf)), connects, WSubField(WRef(buf), "O", ClockType, SourceFlow))
+        Connect(NoInfo, WSubField(WRef(buf), "CE"),
+          Seq(WRef(enableReg), WRef(finishing), nReset).reduce(And.apply))))
```
albert-magyar, line 180:
```suggestion
          And.reduce(Seq(WRef(enableReg), WRef(finishing), nReset)))))
```

sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -147,35 +147,82 @@ object FAMEModuleTransformer {
...
+      *
+      * @param Port The reference to the original target port
+      *
+      * @param outputChannelEnable A boolean that, when asserted alongside
+      * targetCycleFinishing, serves to reset output channels in the associated
+      * clock domain so that new output tokens may be enqueued.
+      *
+      * @param inputChannelEnable Ditto, above. This is asserted one pipeline
+      * stage earlier, ensuring that all inputs are dequeued synchronously with
+      * the launching of the target clock.
```
albert-magyar, line 161: I am not really sure what the comments for `outputChannelEnable` / `inputChannelEnable` are saying.

As far as what the enables do, I think that "reset [input|output] channels" is maybe unclear vs. "lets an [input|output] channel synchronous to this clock [consume|produce] a token in this time step," since the reset as an FSM transition is just a convenient FIRRTL code golf thing.

And maybe it might make more sense to reverse the description and say that the `outputChannelEnable` is a pipelined version of `inputChannelEnable`.

albert-magyar (approved):
LGTM

davidbiancolin:
Alright i created a schematic for this F1 xform variant. It'll be included in my dissertation. In this example Ij and On are synchronous to clock0. 
image

PR #648, opened 2020-10-19
Fix RAW bug in firesim_tsi loadmem_read

jerryz123:
Wait for the write queue to empty before reading in loadmem.

This was a bug in 1.9.0.


davidbiancolin (approved):
LGTM

Issue #646, opened 2020-10-15
Use assembly FAT JARs in Chisel Build


timsnyder-siv:
@davidbiancolin , we've discussed this before as a path toward making #509 easier to implement.  There are others over here that would like to see me take a FAT JAR from our world and give it to FireSim instead of letting FireSim's makefiles call SBT.

I'm to the point where I'd like to implement this and submit a PR if you're still open to the idea.

davidbiancolin:
I'm open. How are you proposing to do that? 

Spitballing here, the problem as i see it is that there are three non-NOP recipes in the makefile that do:

1) Target elaboration 
2) GG compilation
3) Preprocessing and moving files into the vivado project directory (replace RTL)

(1) and (2) can be farmed out to a fat jar invocation on the remote. (3) can just be moved into a script. 

You can extend the buildafi function to compile your fat JAR. I think this should probably be a recipe in the makefiles anyways. Then you can dry-run invoke Make to get the arguments for steps (1), (2), (3).

timsnyder-siv:
You summarized how I was thinking about it.  I was planning add a target for creating the fat Jar.  Something like `firesim/sim/generated-src/f1/assembly.jar` since it is config independent.

Then I was going to make everything else that uses SBT call the fat JAR instead (and try not to break bloop or the other TARGET_PROJECTs in the process).

I had thought that the stuff you labeled target 3 for the preprocessing and moving stuff into the `aws-fpga/hdk/.../cl_firesim` copy should be broken out into a script.

timsnyder-siv:
As I'm starting to look into this in more detail and actually implement it, I'm realizing that Bloop doesn't support creation of a FAT jar.  If we make the fat_jar stuff an option, it is going to add a bunch of complexity to the makefiles.  If it isn't an option, I don't see how to keep the Bloop stuff.

@davidbiancolin do you want to see how ugly it is if I make the FAT_JAR stuff an option or do you think it's okay to remove the bloop support or is there a bloop expert that I should be talking to about this?

davidbiancolin:
What if the SBT flow just always built a fat jar and invoked that directly, that should streamline this right? I don't see why these two flows can't coexist. 

You could try upgrading to 1.4.0. The biggest problem is that jvm launch and SBT cross-project key resolution takes 10s of seconds, and specifically that hurts running MIDAS examples, where it can double or triple the runtime of small tests.

[...]

PR #651, opened 2020-10-26
SBT chisel flow uses assembly jar

timsnyder-siv:
fixes firesim/firesim#646

Run SBT once to create generated-src/<sbt_project>.jar and then run that jar for the generator and GoldenGate


sim/target-agnostic.mk
```diff
@@ -71,16 +73,16 @@ $(VERILOG) $(HEADER) $(fame_annos): $(FIRRTL_FILE) $(ANNO_FILE) $(SCALA_BUILDTOO
 # to generate a runtime configuration that is compatible with the generated
 # hardware (BridgeModule). Useful for modelling a memory system that differs from the default.
 .PHONY: conf
-conf: $(fame_annos)
+conf: $(fame_annos) $(FAT_JAR)
 	mkdir -p $(GENERATED_DIR)
 	# Runtime configuration generator must run under SBT currently; When
 	# launched via bloop some Console input and output is lost.
-	cd $(base_dir) && $(SBT) "project $(firesim_sbt_project)" "runMain midas.stage.RuntimeConfigGeneratorMain \
+	cd $(base_dir) && java $(JAVA_ARGS) -cp $(FAT_JAR) midas.stage.RuntimeConfigGeneratorMain \
```
timsnyder-siv, line 80: Looking at this again, if someone expects `make ENABLE_BLOOP=1 conf` to work, my change here will break that.  We could just roll this target back to be hardcoded for SBT.

jackkoenig:
Have you tried using SBT 1.4 with `sbtn` to see if it is fast enough in client-server mode to remove the need for Bloop?

sim/target-agnostic.mk
davidbiancolin: Defaulting to SBT seems reasonable.

sim/Makefile
```diff
@@ -50,21 +50,28 @@ BLOOP_NAILGUN_PORT ?= 8212
...
-endef
+  override SCALA_BUILDTOOL_DEPS += $(BLOOP_CONFIG_DIR)/TIMESTAMP
+  # Two notes about the bloop invocation:
+  # 1) the sed removes a leading {file:<path>} that sometimes needs to be
+  #    provided to SBT when a project but not for bloop.
+  # 2) Generally, one could could pass '--' to indicate all remaining arguments are
+  # 	 destined for the scala Main, however a bug in Bloop's argument parsing causes the
+  #    --nailgun-port argument to be lost in this case. Workaround this by prefixing
+  #    every main-destined argument with "--args"
+  define run_scala_main
```
davidbiancolin, line 61: Can you explain to me again why we need both a "run_scala_main" and a "run_main".

sim/build.sbt
```diff
@@ -87,6 +101,8 @@ lazy val firesim    = (project in file("."))
     Compile / doc := (doc in ScalaUnidoc).value,
     // Registers the unidoc-generated html with sbt-site
     addMappingsToSiteDir(mappings in (ScalaUnidoc, packageDoc), siteSubdirName in ScalaUnidoc),
-    concurrentRestrictions += Tags.limit(Tags.Test, 1)
+    concurrentRestrictions += Tags.limit(Tags.Test, 1),
+    mainClass in assembly := Some("chipyard.Generator"),
```
davidbiancolin, line 105: This just specifies a _default_ main class in the jar right?

davidbiancolin (commented):
Looks pretty good. After some more thought, i think it makes sense to bundle this in with the proposed manager changes.

sim/build.sbt
timsnyder-siv: Yes.  I have a suspicion that if you have more than one main class and you don't specify a default that assembly wigs out. I didn't test that theory, I just saw other projects were specifying a default and also did that.

sim/Makefile
timsnyder-siv: When using the fat jar, the generator and goldengate targets want to call java with the jar instead of the scala build tool.  When when not using the fat jar, we just call the scala build tool all the time.  

We decided to make SBT always use the fat jar and Bloop doesn't support fat jars, so it always calls the scala build tool, Bloop.

timsnyder-siv:
> i think it makes sense to bundle this in with the proposed manager changes.

Sounds good.  I probably also need to look into why the CI is failing.

sim/target-agnostic.mk
timsnyder-siv: Rolled this back in 9f611c63d

timsnyder-siv:
The CI jobs were failing because of a failure to find libff when trying to install `fabric` as a part of the machine-launch-script.sh.  I cherry-picked the fix from dev that I was missing and now I'm getting a legit failure when we try to run `make -C sim TARGET_PROJECT=midasexamples sbt SBT_COMMAND=...`  It looks like `sbt assembly` is being run for some reason and it is cribbing about `emulator.cc` being different.  I'm looking into it.

timsnyder-siv:
Makes sense why `sbt assembly` is running.  The toplevel make is calling:
```
make -C sim verilator TARGET_PROJECT=midasexamples DESIGN=GCD TARGET_CONFIG=NoConfig PLATFORM_CONFIG=HostDebugFeatures_DefaultF1Config PLATFORM=f1
```
And that starts by trying to build the FAT jar for the sbt project `firesim` and I didn't test that locally.    I am digging into how to resolve the error assembly encounters:
```
[error] (assembly) deduplicate: different file contents found in the following:
[error] /home/centos/project/target-design/chipyard/generators/rocket-chip/src/target/scala-2.12/rocketchip_2.12-1.0.jar:csrc/emulator.cc
[error] /home/centos/project/target-design/chipyard/generators/utilities/target/scala-2.12/utilities_2.12-1.0.jar:csrc/emulator.cc
```

I think this is working in the `firechip` project because the assembly merge strategies are different.  [Chipyard's build.sbt](https://github.com/ucb-bar/chipyard/blob/02c889b8b15b86587feae852f1165fade348bb4b/build.sbt#L13-L15) sets firechip to use:
```
  assemblyMergeStrategy in assembly := { _ match {
    case PathList("META-INF", "MANIFEST.MF") => MergeStrategy.discard
    case _ => MergeStrategy.first}}
``` 
Which is really aggressive, it just says use the first one whenever there's a conflict in any package.  I didn't change what Chipyard was doing in this PR because I wasn't sure if anyone else in chipyard was actually using the fat jars and knew they needed this.

firesim/sim/build.sbt did not have any merge stragegy settings for assembly and I added the settings I'm familiar with from another closed-source project into [firesim/sim/build.sbt](https://github.com/firesim/firesim/pull/651/files#diff-a2251bcdfc5057b878c3a1b331609e38a21a172e892a46db53b2b1ebf80a11e5R26-R38):
```
  assemblyMergeStrategy in assembly := {
    case PathList("META-INF", "services", xs @ _*) => MergeStrategy.concat
    // Discard Metadata, it's irrelevant
    case PathList("META-INF", xs @ _*) => MergeStrategy.discard
    // When any of our dependencies are different versions than those of firrtl.jar, there will be conflicts
    // When this occurs, pick last one which is stuff in .ivy2 (ie. not firrtl.jar)
    case PathList(xs @ _*) if xs.last.endsWith(".class") || xs.last.endsWith(".properties") => MergeStrategy.last
    // Just take the last matching joda/time/tz/data resource files
    case PathList("org", "joda", "time", "tz", "data", xs @ _*) => MergeStrategy.last
    case x =>
      val oldStrategy = (assemblyMergeStrategy in assembly).value
      oldStrategy(x)
  },
```
The more conservative `assemblyMergeStrategy` I used in firesim/sim/build.sbt doesn't have a rule to handle emulator.cc differences between [chipard/generators/utilities/src/main/resources/csrc/emulator.cc](https://github.com/ucb-bar/chipyard/blob/02c889b8b15b86587feae852f1165fade348bb4b/generators/utilities/src/main/resources/csrc/emulator.cc) and [chipyard/generators/rocket-chip/src/main/resources/csrc/emulator.cc](https://github.com/chipsalliance/rocket-chip/blob/653efa99a27dc155bd4b4706a7e71c5c930f62b1/src/main/resources/csrc/emulator.cc)

The diff of the two emulator.cc versions (going from RC as Version 1 and CY/utils as Version 2) can be seen at https://gist.github.com/timsnyder-siv/180c3a3253ffb996ee8c5aa365061791/revisions#diff-c8f58a8db10ec1bd79fe0f5f272fedba985e2b891bc523e9ee02adaa9d358d1f

sim/build.sbt
```diff
@@ -4,6 +4,30 @@ import Tests._
...
+      sbtassembly.AssemblyUtils.sourceOfFileForMerge(tempDir, f) match {
+        case (jar, _, _, true) if !jar.toString.contains("rocketchip") => f
+      }
+    }
+    if (filtered.size == 1) Right(Seq(filtered.head -> path))
+    else Left("still have multiple files after removing rocketchip for same target path:" +
+      filenames(tempDir, filtered).mkString("\n", "\n", "")
+      )
+  }
+}
```
timsnyder-siv, line 28: It seems like `emulator.cc` should come from whatever isn't rocketchip because it was overridden.  I added a custom merge strategy to the firesim/sim/build.sbt in 0c55c1c.  I don't like how hacky it looks.  Perhaps @jackkoenig might have some advice on making it more concise?  This should make the CI pass.

sim/build.sbt
jackkoenig: It's not strictly necessary to specify a `mainClass`, FIRRTL doesn't: https://github.com/freechipsproject/firrtl/blob/d1c0181e716c37142e233beed2efcea5c5794aa7/build.sbt#L65

But it is just the default.

sim/build.sbt
timsnyder-siv: Thanks @jackkoenig , I get confused about whether it's better to use `filter`, `map` or `collect`.  My understanding is that `collect` is just a combination of `filter` and `map`.

Now that I have the CI passing for FireSim using the fat jars and after looking at [chipyard/generators/utilities/src/main/scala/Simulator.scala's use of `getClass.getResourceAsStream()`](https://github.com/ucb-bar/chipyard/blob/02c889b8b15b86587feae852f1165fade348bb4b/generators/utilities/src/main/scala/Simulator.scala#L68-L75) to emit `/csrc/emulator.cc`, we definitely want to make sure that `chipyard/build.sbt` also updates it's `assemblyMergeStrategy` to be a bit more conservative and also handle picking `emulator.cc` that isn't coming from rocketchip.  I'll open a dependent Chipyard PR.

davidbiancolin: Pretty much the only case i use filter + map these days is if i can _very_ succinctly describe a function (T) => Bool. Otherwise i use collect. It really shines if you need to do any non-trivial pattern matching or want to filter on type, and tends to be more succinct. 

In general you should probably tend to using fewer higher-order functions for readability.

davidbiancolin:
The proposed CY sbt changes seem reasonable.

timsnyder-siv:
Marked this as a draft because I'm planning to add the manager changes to be able to use the FAT jar.

timsnyder-siv:
Hey David, as I'm testing the distributed elaboration flow, I'm realizing that I didn't remember that the firrtl generation tries to read the bootrom.img file from the file path.  I'm running into:
```
[192.168.1.150] out: [error] java.nio.file.NoSuchFileException: /home/centos/firesim_fat_jar/sim/./target-rtl/chipyard/generators/testchipip/bootrom/bootrom.rv64.img
```

Do you have an opinion on whether I should copy that file over or modify the bootrom stuff to access the bootrom.rv64.img as a Java resource instead?  If I do that, it will get bundled into the FAT jar and get pulled out of the jar when we try to run Java.  Shouldn't be too hard.  I'll just do something similar to [cy/generators/utilities/src/main/scala/Simulator.scala](https://github.com/ucb-bar/chipyard/blob/c083d5d94762878c1262190045fb9712ccd39ed9/generators/utilities/src/main/scala/Simulator.scala)  Well, wait a minute, I'm confused because I see stuff in that file accessing bootrom.img files as resources.  I also see `testchipip/bootrom/bootrom.rv64.img` in the fat jar.  

I guess that the RocketChip HasPeripheryBootROM  should be overridden by something that knows to get the bootrom out of the java resource instead of the file path?

davidbiancolin:
It's a little bit weird to me that the bootrom, which should be a generated artefact, would appear as a resource (maybe @jerryz123 or @abejgonzalez  can provide some insight there), so i'm not sure if those responsible for managing the `HasPeripheryBootROM` would welcome the change.  If not, given that it is treated as a resource in CY however, another option is the add a pre-elaboration phase that writes out the bootrom to generated-src. 

In general, having some pre-elaboration setup of generated-src (and thus supporting the copying of generated files to the remote before target-elaboration) seems useful. You could just handle copying the directory, and then punt to the user to make sure everything their generator needs is accessible in their design's generated-src dir prior to elaboration, instead of copying specific files.

abejgonzalez:
If I recall correctly, there was some work in the past to allow Chipyard to use published dependencies (https://github.com/ucb-bar/chipyard/pull/47) and this bootrom copying was a part of that. I think if `testchipip` was a published dependency, then there would be no path to the bootrom file so something (in this case `Simulator.scala` which is run before the generator) has to copy the file from the resources to a place that `HasPeripheryBootROM` can access it. I definitely think that getting the bootrom from a resource instead of a path would be ideal (but like David said... may not go over well with SiFive folks).

timsnyder-siv:
> another option is the add a pre-elaboration phase that writes out the bootrom to generated-src.

Looking at this more, it appears that the [firesim/sim/src/main/makefrag/firesim/Makefrag](https://github.com/firesim/firesim/blob/9700899c5a0ddb3684dbe364e7647c2509bd4028/sim/src/main/makefrag/firesim/Makefrag#L84-L86) is already symlinking the bootrom into the generated-src.  It seems like the work to be done here is : 
- making that DROMAJO_ROM target a prereq for the pre-elaboration phase 
   - possibly change it from creating a symlink to just copying over the bootrom.img so that I don't have to start worrying about symlinks in the rsyncing back and forth
   - It might also make sense to start calling it BOOT_ROM so that it looks less like it is only used for DROMAJO.  @davidbiancolin and @abejgonzalez, What do you guys think? 
- change `firesim.firesim.WithBootROM` to use the copied ROM instead of the source ROM.

timsnyder-siv:
Also, the copied `bootrom.rv{32,64}.img` is named `$(GENERATED_DIR)/$(long_name).rom` (i.e. `firesim.firesim.FireSim.FireSimRocketConfig.rom`).  I would expect something named with the `$(long_name)` to have the full contents of the bootrom that the elaborated design will have (i.e. the bootrom.img plus the compiled dtb).  The name being used makes it look like there is something specific from the config that is captured when there really isn't.

Am I picking at nits here or would it make sense to just copy it over as `bootrom.rv{32,64}.img`?  I ask because it will remove the need for `firesim.firesim.WithBootROM` to learn about `$(long_name)`.  It can just look for `bootrom.rv{32,64}.img` in the directory pointed to by `firrtl.options.TargetDirAnnotation`.

davidbiancolin:
[Accidentally edited your comment instead of replying]

> Am I picking at nits here or would it make sense to just copy it over as bootrom.rv{32,64}.img? I ask because it will remove the need for firesim.firesim.WithBootROM to learn about $(long_name). It can just look for bootrom.rv{32,64}.img in the directory pointed to by firrtl.options.TargetDirAnnotation.

This sounds better. It seems to me the safest thing to do is copy both 32b and 64b variants, retaining their names. @abejgonzalez thoughts?

timsnyder-siv:
> It would be nice if this change, if harmonized across CY, would let us delete the firesim variant of this config fragment -- but we can leave that to future work.

You mean, keep `chipyard.config.WithBootROM` and get rid of `firesim.firesim.WithBootROM`?  

> Could we not make rsync simply following the symlinks and copy the target instead? I think it's alright to copy the file in though.

I thought about this.  We could but right now I'm planning to rsync the generated-dir recursively to the build host and then back to the manager.  If we follow the links copying there, then copying back to the manager gets more complicated than I would like it to be.  Maybe there's a rsync option for handling that but I didn't look.

abejgonzalez:
> [Accidentally edited your comment instead of replying]
> 
> > Am I picking at nits here or would it make sense to just copy it over as bootrom.rv{32,64}.img? I ask because it will remove the need for firesim.firesim.WithBootROM to learn about $(long_name). It can just look for bootrom.rv{32,64}.img in the directory pointed to by firrtl.options.TargetDirAnnotation.
> 
> This sounds better. It seems to me the safest thing to do is copy both 32b and 64b variants, retaining their names. @abejgonzalez thoughts?

This seems reasonable to me. Right now the BootROMs aren't huge so I don't think its an issue copying both over. I also agree that we can change the name to not be `$(long_name)` since there is no config. specific changes to the BootROM.

timsnyder-siv:
@sagark , I've pushed my changes to the firesim manager that enable it to use a FAT jar from sbt-assembly and run the Chisel to Verilog elaboration on the build hosts.    The PR is currently based on the point I'm using internally and I see that I'll need to merge dev into my PR to resolve conflicts but I wanted to get this out for feedback

deploy/buildtools/buildafi.py
```diff
@@ -21,6 +24,32 @@ def get_deploy_dir():
...
+         prefix('export RISCV={}'.format(os.getenv('RISCV', ""))), \
+         prefix('export PATH={}'.format(os.getenv('PATH', ""))), \
+         prefix('export LD_LIBRARY_PATH={}'.format(os.getenv('LD_LIBRARY_PATH', ""))), \
+         prefix('source ./sourceme-f1-manager.sh'), \
+         prefix('cd sim/'), \
+         InfoStreamLogger('stdout'), \
+         InfoStreamLogger('stderr'):
+        return local(cmdline, capture=capture)
+
+def get_sim_makevar(buildconfig, varname):
```
davidbiancolin, line 40: This feels to me like it belongs in a more general place since it's probably more widely useful.

deploy/buildtools/buildafi.py
```diff
@@ -21,6 +24,32 @@ def get_deploy_dir():
...
+        varname
+    )
+
+    res = sim_local(cmdline, capture=True)
+    var_results = res.stdout.splitlines()
+    assert len(var_results) == 1, "get_sim_makevar() command " + \
+           res.real_command + " STDOUT is not a single line. It was:\n" + pformat(var_results)
+    return var_results[0].split(None, 2)[-1]
+
 def replace_rtl(conf, buildconfig):
```
davidbiancolin, line 53: rename to replace_rtl_local and update teh comment?

deploy/buildtools/buildafi.py
```diff
@@ -67,51 +97,107 @@ def aws_build(global_build_config, bypass=False):
...
+        # rsync fpgabuilddir to manager
+        rsync_cap = rsync_project(local_dir=pjoin(aws_fpga_root, fpgabuilddir),
+                                  remote_dir=pjoin(aws_fpga_root, fpgabuilddir, ''),
+                                  ssh_opts="-o StrictHostKeyChecking=no", upload=False, extra_opts="-l",
+                                  capture=True)
+        rootLogger.debug(rsync_cap)
+        rootLogger.debug(rsync_cap.stderr)
+        local("mkdir -p {}/results-build/{}/".format(ddir, results_builddir))
+        local("cp {}/design/cl_firesim_generated.sv {}/results-build/{}/cl_firesim_generated.sv".format(
+        pjoin(aws_fpga_root, fpgabuilddir),
```
davidbiancolin, line 176: Indent.

deploy/buildtools/buildafi.py
```diff
@@ -67,51 +97,107 @@ def aws_build(global_build_config, bypass=False):
...
+        rootLogger.debug(rsync_cap.stderr)
+        # sync the fat jar
+        rsync_cap = rsync_project(local_dir=fat_jar,
+                      remote_dir=fat_jar,
                       ssh_opts="-o StrictHostKeyChecking=no",
                       extra_opts="-l", capture=True)
         rootLogger.debug(rsync_cap)
         rootLogger.debug(rsync_cap.stderr)
+        # install java and dtc on the build host
+        sudo('yum install -y java dtc')
```
davidbiancolin, line 156: I wonder if this should be hoisted out and included in a machine-launch for build instances. Future work.

deploy/buildtools/buildafi.py
```diff
@@ -164,17 +250,21 @@ def aws_build(global_build_config, bypass=False):
...
+        rootLogger.debug("create-fpga-image result: " + str(agfi_afi_ids))
         ids_as_dict = json.loads(agfi_afi_ids)
         agfi = ids_as_dict["FpgaImageGlobalId"]
         afi = ids_as_dict["FpgaImageId"]
         rootLogger.info("Resulting AGFI: " + str(agfi))
         rootLogger.info("Resulting AFI: " + str(afi))
 
+    if global_build_config.distributed_elaboration:
+        # if we generated verilog (and driver headers) remotely, we now need to build the driver locally
+        rootLogger.info("Building FireSim-f1 driver")
```
davidbiancolin, line 262: I'm assuming the timestamps workout such that only the C++ is being compiled here? 
We could also just build the driver in all cases and drop the if.

sim/build.sbt
```diff
@@ -4,6 +4,30 @@ import Tests._
...
 
+def filenames(tempDir: File, fs: Seq[File]): Seq[String] =
+for(f <- fs) yield {
+  sbtassembly.AssemblyUtils.sourceOfFileForMerge(tempDir, f) match {
+    case (path, base, subDirPath, false) => subDirPath
+    case (jar, base, subJarPath, true) => jar + ":" + subJarPath
+  }
+}
+
+val not_rc_merge_strategy = new sbtassembly.MergeStrategy {
```
davidbiancolin, line 15: Camel casing + a brief comment explaining the intent of this code would be nice.

sim/src/main/makefrag/fasedtests/Makefrag
```diff
@@ -31,16 +31,29 @@ HEADER  := $(GENERATED_DIR)/$(DESIGN)-const.h
...
 chisel_src_dirs = \
 		$(addprefix $(firesim_base_dir)/,. midas midas/targetutils firesim-lib) \
 		$(addprefix $(chipyard_dir)/generators/, chipyard rocket-chip/src, rocket-chip/api-config-chipsalliance)
 
 chisel_srcs = $(foreach submodule,$(chisel_src_dirs),\
 	$(shell find $(submodule)/ -iname "[!.]*.scala" -print 2> /dev/null | grep 'src/main/scala'))
 
-$(FIRRTL_FILE) $(ANNO_FILE): $(chisel_srcs) $(FIRRTL_JAR) $(SCALA_BUILDTOOL_DEPS)
+# don't search for implicit rules to make the chisel_srcs (reduces useless output in --debug=i)
+$(chisel_srcs): ;
```
davidbiancolin, line 43: Nice

davidbiancolin (commented):
Pretty cool stuff; this turned out cleaner than i expected considering just how many things you have to touch...

deploy/buildtools/buildafi.py
timsnyder-siv: Yes, the timestamps should be preserved if I remember correctly from reading the Fabric rsync'ing code but I'll double check.  I thought it was useful to have the comment there.  You're correct that it won't hurt anything to remove the if.  I can do that and adjust the comment.

sim/build.sbt
timsnyder-siv: Will do.

deploy/buildtools/buildafi.py
timsnyder-siv: I thought about this but got lazy when it came time to figure out how to add it to the boto calls that launch the build instances.

deploy/buildtools/buildafi.py
timsnyder-siv: I'll take a look for a better place.

deploy/buildtools/buildafi.py
timsnyder-siv: I understand the desire to rename but @davidbiancolin what were you thinking you wanted to update in teh comment?

deploy/buildtools/buildafi.py
davidbiancolin: Mmm, that makes sense. I'll leave it to your discretion.

deploy/buildtools/buildafi.py
davidbiancolin: Something more to the effect of: runs the build locally instead of shipping a fat jar. The name will capture that though.

deploy/buildtools/buildafi.py
timsnyder-siv: @davidbiancolin do you have thoughts on where else it would be useful to have this?  

I thought about moving this stuff into `buildtools/buildconfig.py` but `buildconfig` doesn't have any fabric stuff in it and it seemed good to leave fabric out of the buildconfig.

Perhaps as a new module in the `util` package?  Maybe `util/simshim.py`?  Or, maybe `util/simmakefiles.py`?

If we start moving `get_sim_makevar`, it probably makes sense to put `sim_local` in the new place.  What about `buildafi.get_deploy_dir()`?  Should we move that too?  While we're talking about get_deploy_dir, it makes more sense to me that the GlobalBuildConfig and RunTimeConfig would capture the working directory and pass it down to be used rather than introspected multiple times by ssh'ing back from all build hosts to the manager to run `pwd`. 

I'm starting to feel like moving this somewhere else is part of a larger refactor of this code.  Let me know what you think.

deploy/buildtools/buildafi.py
timsnyder-siv: Ok.  GH is also doing a great job of hiding the third line of the comment "THIS ALWAYS RUNS LOCALLY". ;)

deploy/buildtools/buildafi.py
timsnyder-siv: Fabric 1.14 `rsync_project` uses `-pthrvz` as the default options. https://github.com/fabric/fabric/blob/1.14/fabric/contrib/project.py#L27

`-t` preserves the modification time and I don't see us passing in different default_opts to any of the calls or extra_opts that would change that.

I suppose there's always the chance that AWS system times may be inconsistent.  I've run into trouble with rsync'ing files correctly between hosts when the ntpd was jacked in the datacenter and having really weird things happen.  

I checked my firesim manager and noticed that it's setup using chrony pointing at the usual ntp.org servers.  It might be better to point the manager and build servers at the local Amazon Time Service by adding it's server info (`server 169.254.169.123 prefer iburst` as per https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/) to /etc/crony.conf and restarting chrony as a part of the startup script.  Hard to say whether it is worth it without starting to do some distributed builds and seeing if there are issues.  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html also has some useful info about Amazon Time Sync Service

Before adding the Amazon server, I saw:
```
bash-4.2$ chronyc sources -v
210 Number of sources = 4

  .-- Source mode  '^' = server, '=' = peer, '#' = local clock.
 / .- Source state '*' = current synced, '+' = combined , '-' = not combined,
| /   '?' = unreachable, 'x' = time may be in error, '~' = time too variable.
||                                                 .- xxxx [ yyyy ] +/- zzzz
||      Reachability register (octal) -.           |  xxxx = adjusted offset,
||      Log2(Polling interval) --.      |          |  yyyy = measured offset,
||                                \     |          |  zzzz = estimated error.
||                                 |    |           \
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
^* chl.la                        2   9   377   483   +478us[  +80us] +/-   12ms
^- 108.61.73.244                 2  10   377   673  -2898us[-3218us] +/-   54ms
^- 44.190.6.254                  3  10   317   307    -19ms[  -19ms] +/-   86ms
^- ns1.pulsation.fr              3  10   377   603  +6167us[+5818us] +/-  147ms
```

After adding the Amazon Server, the sync'd server at the top was the amazon server and it's estimated error was almost an order of magnitude smaller (not sure if that will matter at all)
```
bash-4.2$ chronyc sources -v
210 Number of sources = 5

  .-- Source mode  '^' = server, '=' = peer, '#' = local clock.
 / .- Source state '*' = current synced, '+' = combined , '-' = not combined,
| /   '?' = unreachable, 'x' = time may be in error, '~' = time too variable.
||                                                 .- xxxx [ yyyy ] +/- zzzz
||      Reachability register (octal) -.           |  xxxx = adjusted offset,
||      Log2(Polling interval) --.      |          |  yyyy = measured offset,
||                                \     |          |  zzzz = estimated error.
||                                 |    |           \
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
^* 169.254.169.123               3   4     7     2    -99ns[+1380us] +/-  511us
^- 50-205-244-108-static.hf>     2   6    17     1   +667us[ +667us] +/-   56ms
^- 159.203.82.102                4   6     7     3  -2911us[-1531us] +/-   44ms
^- las1.pawnode.com              3   6     7     2  -3055us[-1675us] +/-   47ms
^- pool-173-71-80-166.cmdnn>     2   6     7     2  +3149us[+4529us] +/-  140ms
```

I suppose if we want to avoid distributed time issues, we could always `make -t replace-rtl` after we copy the replace-rtl stuff back over.  I really dislike doing that if we don't have to.

deploy/buildtools/buildafi.py
davidbiancolin: I think you're probably right. We can punt on that, there's enough going on in this PR.

sim/Makefile
```diff
@@ -50,21 +54,27 @@ BLOOP_NAILGUN_PORT ?= 8212
...
+  	cd $(base_dir) && bloop --nailgun-port $(BLOOP_NAILGUN_PORT) run $(shell echo $(1) | sed 's/{.*}//') --main $(2)  $(addprefix --args ,$3)
+  endef
+  # recipes for generator and gg using bloop fall through to run_scala_main using Bloop
+  run_main = $(call run_scala_main,$(1),$(2),$(3))
+  PRE_ELABORATION_TARGETS += $(SCALA_BUILDTOOL_DEPS)
```
timsnyder-siv, line 70: @davidbiancolin as I'm starting to use this more and push on the rules a bit, I see a problem with this code.

The `PRE_ELABORATION_TARGETS` are used in the dependence list of target(s) in the TARGET_PROJECT_MAKEFRAG.  Dependence lists of targets are expanded when the target is defined in the makefile (assuming we aren't using .SECOND_EXPANSION) and thus this assignment of PRE_ELABORATION_TARGETS needs to happen before the `include $(TARGET_PROJECT_MAKEFRAG)`.

So does the assigment down on the new line 77.

sim/target-agnostic.mk
```diff
@@ -51,8 +53,16 @@ firesim_root_sbt_project := {file:$(firesim_base_dir)}firesim
...
-	$(call run_scala_main,$(firesim_sbt_project),midas.stage.GoldenGateMain,\
+$(GENERATED_DIR):
+	mkdir -p $@
+
+# empty recipe to help make understand multiple targets come from single recipe invocation
+# without using the new (4.3) '&:' grouped targets see https://stackoverflow.com/a/41710495
+$(VERILOG) $(HEADER) $(fame_annos): $(VERILOG).intermediate
+
+.SECONDARY: $(VERILOG).intermediate
+$(VERILOG).intermediate: $(FIRRTL_FILE) $(ANNO_FILE) $(RUN_MAIN_DEPS)
```
timsnyder-siv, line 64: Here's another problem.  I renamed `RUN_MAIN_DEPS` to `PRE_ELABORATION_DEPS` and forgot to change it here.

sim/src/main/makefrag/fasedtests/Makefrag
```diff
@@ -52,7 +52,7 @@ $(FAT_JAR): $(chisel_srcs) $(FIRRTL_JAR) $(SCALA_BUILDTOOL_DEPS)
 $(FIRRTL_FILE) $(ANNO_FILE): $(FIRRTL_FILE).intermediate ;
 
 .SECONDARY: $(FIRRTL_FILE).intermediate
-$(FIRRTL_FILE).intermediate: $(chisel_srcs) $(FIRRTL_JAR) $(RUN_MAIN_DEPS) | $(GENERATED_DIR)
+$(FIRRTL_FILE).intermediate: $(chisel_srcs) $(FIRRTL_JAR) $(PRE_ELABORATION_TARGETS) | $(GENERATED_DIR)
```
timsnyder-siv, line 55: This is fixing the variable rename I made partially

sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -191,7 +191,7 @@ vcs = $(GENERATED_DIR)/$(DESIGN)
 vcs_debug = $(GENERATED_DIR)/$(DESIGN)-debug
 vcs_args = +vcs+initreg+0 +vcs+initmem+0
 xsim = $(GENERATED_DIR)/$(DESIGN)-$(PLATFORM)
-sim_binary_basename := $(basename $(notdir $(SIM_BINARY)))
+sim_binary_basename = $(basename $(notdir $(SIM_BINARY)))
```
timsnyder-siv, line 194: change to recursive type so that --warn-undefined-variables doesn't complain about SIM_BINARY.  It is expected to be given on the cmdline.  It might be nice in the future to do an explicit check that if one of the `run-$EMUL-$DEBUG` targets is being run, that SIM_BINARY is defined and exists.

sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -241,7 +241,9 @@ endif
 disasm := 2>
 which_disasm := $(shell which spike-dasm 2> /dev/null)
 ifneq ($(which_disasm),)
-        disasm := 3>&1 1>&2 2>&3 | $(which_disasm) $(DISASM_EXTENSION) >
+    # Default to no extra spike-dasm arguments, Allowing cmdline addition
+    DISASM_EXTENSION ?=
+    disasm := 3>&1 1>&2 2>&3 | $(which_disasm) $(DISASM_EXTENSION) >
```
timsnyder-siv, line 246: Clarify that the default behavior is to run `spike-dasm` without any arguments but providing a variable that can be set on the cmdline.

sim/target-agnostic.mk
```diff
@@ -58,10 +58,10 @@ $(GENERATED_DIR):
 
 # empty recipe to help make understand multiple targets come from single recipe invocation
 # without using the new (4.3) '&:' grouped targets see https://stackoverflow.com/a/41710495
-$(VERILOG) $(HEADER) $(fame_annos): $(VERILOG).intermediate
+$(VERILOG) $(HEADER) $(fame_annos): $(VERILOG).intermediate ;
```
timsnyder-siv, line 61: This rule needs the semicolon to make it explicitly empty (without requiring whitespace or dumping extraneous prints during runtime)

sim/target-agnostic.mk
```diff
@@ -58,10 +58,10 @@ $(GENERATED_DIR):
...
 
 .SECONDARY: $(VERILOG).intermediate
-$(VERILOG).intermediate: $(FIRRTL_FILE) $(ANNO_FILE) $(RUN_MAIN_DEPS)
+$(VERILOG).intermediate: $(FIRRTL_FILE) $(ANNO_FILE) $(PRE_ELABORATION_TARGETS)
```
timsnyder-siv, line 64: The other missed rename.

sim/target-agnostic.mk
```diff
@@ -102,8 +102,8 @@ VERILATOR_MAKEFLAGS ?= -j8 VM_PARALLEL_BUILDS=1
 verilator = $(GENERATED_DIR)/V$(DESIGN)
 verilator_debug = $(GENERATED_DIR)/V$(DESIGN)-debug
 
-$(verilator) $(verilator_debug): export CXXFLAGS := $(CXXFLAGS) $(common_cxx_flags) $(VERILATOR_CXXOPTS) -D RTLSIM
-$(verilator) $(verilator_debug): export LDFLAGS := $(LDFLAGS) $(common_ld_flags)
+$(verilator) $(verilator_debug): export CXXFLAGS += $(common_cxx_flags) $(VERILATOR_CXXOPTS) -D RTLSIM
+$(verilator) $(verilator_debug): export LDFLAGS += $(common_ld_flags)
```
timsnyder-siv, line 106: There are several of these target-specific variable definitions that I modified.  While it is great to use simple variable types these statements cause`--warn-undefined-variables` to complain because you could potentially be using an undefined variable.  Since the expansion here is not expensive, switch to using simple recursive += 

If you really still want/need these to be simply expanded then adding an `ifndef` block for each of them to initialize them to be explicitly empty will also quell the `--warn-undefined-variables` messages.

sim/target-agnostic.mk
```diff
@@ -155,12 +155,12 @@ $(PLATFORM): $($(PLATFORM)) $(OUTPUT_DIR)/$(DESIGN).chain
...
 fpga_dir = $(firesim_base_dir)/../platforms/$(PLATFORM)/aws-fpga
 
-$(f1): export CXXFLAGS := $(CXXFLAGS) $(common_cxx_flags) $(DRIVER_CXXOPTS) -I$(fpga_dir)/sdk/userspace/include
+$(f1): export CXXFLAGS += $(common_cxx_flags) $(DRIVER_CXXOPTS) -I$(fpga_dir)/sdk/userspace/include
 # Statically link libfesvr to make it easier to distribute drivers to f1 instances
-$(f1): export LDFLAGS := $(LDFLAGS) $(common_ld_flags) -lfpga_mgmt
+$(f1): export LDFLAGS += $(common_ld_flags) -lfpga_mgmt
 
 # Compile Driver
-$(f1): $(HEADER) $(DRIVER_CC) $(DRIVER_H) $(midas_cc) $(midas_h) $(runtime_conf)
```
timsnyder-siv, line 153: I don't see `runtime_conf` being defined anywhere in the makefiles.  It's not hurting anything to add nothing here but removing it cleans up a `--warn-undefined-variables` message.

sim/target-agnostic.mk
```diff
@@ -214,13 +214,13 @@ replace-rtl: $(fpga_v) $(ila_work_dir) $(fpga_vh) $(fpga_tcl_env)
 
 .PHONY: replace-rtl
 
-$(firesim_base_dir)/scripts/checkpoints/$(target_sim_tuple): $(fpga_work_dir)/stamp
+$(firesim_base_dir)/scripts/checkpoints/$(name_tuple): $(fpga_work_dir)/stamp
```
timsnyder-siv, line 217: `target_sim_tuple` isn't defined anywhere either.  It looks like it may have been renamed to `name_tuple`.  I know these targets aren't encouraged to be used but I thought I'd clean them up.

sim/target-agnostic.mk
```diff
@@ -275,10 +275,12 @@ unittest_args = \
...
 		TOP_DIR=$(chipyard_dir)
 
-run-midas-unittests: $(chisel_srcs)
+.PHONY: run-midas-unittests
+run-midas-unittests:
 	$(MAKE) -f $(simif_dir)/unittest/Makefrag $@ $(unittest_args)
 
-run-midas-unittests-debug: $(chisel_srcs)
+.PHONY: run-midas-unittests-debug
+run-midas-unittests-debug:
```
timsnyder-siv, line 283: These targets look like they were promoted from a subdir to be `.PHONY`  targets of the toplevel makefile.  However, their dependency lists are a variable that isn't defined correctly for the subdir in question and is only defined by the TARGET_PROJECT_MAKEFRAGS for midasexamples and fasedtests.

I think the correct thing here is to make them explicitly `.PHONY` and remove their empty (or incorrect) dependency lists.

timsnyder-siv (commented):
After noticing the variable renames that I did halfway and that the inclusion of TARGET_PROJECT_MAKEFRAG needed to be after PRE_ELABORATION_TARGETS is defined, I went looking for usages of other variables that weren't defined in the makefiles.

We may not need to adopt all of these but this is what I would do.

sim/target-agnostic.mk
davidbiancolin: Cool, thanks. :)

deploy/buildtools/buildafi.py
```diff
@@ -67,51 +97,112 @@ def aws_build(global_build_config, bypass=False):
...
+	    rootLogger.debug(rsync_cap)
+	    rootLogger.debug(rsync_cap.stderr)
+	    # sync the fat jar
+	    rsync_cap = rsync_project(local_dir=fat_jar,
+			  remote_dir=fat_jar,
+			  ssh_opts="-o StrictHostKeyChecking=no",
+			  extra_opts="-l", capture=True)
+	    rootLogger.debug(rsync_cap)
+	    rootLogger.debug(rsync_cap.stderr)
+        with InfoStreamLogger('stdout'), InfoStreamLogger('stderr'):
```
timsnyder-siv, line 157: ugh.  This indenting is wrong.
```suggestion
      with InfoStreamLogger('stdout'), InfoStreamLogger('stderr'):
```

timsnyder-siv:
Since this PR makes changes to build.sbt and #658 also makes changes to build.sbt, @davidbiancolin and I have talked about waiting to merge this PR until the chisel3.4 (#658) changes get merged in.

I also found a slight bug that occur'd when I had a mixture of local replace-rtl and distributed in the same checkout.  Already pushed a fix to the gen-replace-rtl-script recipe in target-agnostic.mk

deploy/buildtools/buildafi.py
```diff
@@ -67,51 +97,112 @@ def aws_build(global_build_config, bypass=False):
...
         rootLogger.debug(rsync_cap.stderr)
-        rsync_cap = rsync_project(local_dir=ddir + "/../platforms/f1/aws-fpga/{}/*".format(fpgabuilddir),
-                      remote_dir='/home/centos/firesim-build/platforms/f1/aws-fpga/' + remotefpgabuilddir,
-                      exclude='build/checkpoints',
-                      ssh_opts="-o StrictHostKeyChecking=no",
-                      extra_opts="-l", capture=True)
-        rootLogger.debug(rsync_cap)
-        rootLogger.debug(rsync_cap.stderr)
+
+    if global_build_config.distributed_elaboration:
```
timsnyder-siv, line 134: @davidbiancolin good call having me put this behind an option.  I'm using this because I need to manually hack my way round the DCE issue for now.

sim/target-agnostic.mk
```diff
@@ -200,15 +214,36 @@ replace-rtl: $(fpga_v) $(ila_work_dir) $(fpga_vh) $(fpga_tcl_env)
...
+#  $(FAT_JAR) -> replace-rtl, regardless of whether they need to be done.  This is done so that
+#  replace-rtl.sh should always be legit commands and we don't have to handle shortcuts for output
+#  that look like "make: Nothing to be done for `replace-rtl'."
+#  --assume-old=$(fpga_work_dir)/stamp ensures that the replace-rtl.sh will be generated including the
+#  recipe for $(fpga_work_dir)/stamp.  This is needed because  distributed-elaboration mode of
+#  'firesim buildafi' assumes that it will rsync $(board_dir)/cl_firesim to build host and
+#  replace-rtl.sh is responsible for creating $(fpga_work_dir) from the template.
+.PHONY: gen-replace-rtl-script
+gen-replace-rtl-script: $(PRE_ELABORATION_TARGETS) | $(GENERATED_DIR)
+	$(MAKE) replace-rtl -n  --no-print-directory --assume-old=$(fpga_work_dir)/stamp \
```
timsnyder-siv, line 243: ugh.  I pushed this too quickly without testing (or reading carefully enough about what --assume-old does).  --assume-old is not what we want here.  That tells make to assume that you can't rebuild that file because it is lost-knowledge from ancient times that you don't know how to remake.  I thought it was the opposite of --assume-new, i.e. assume that the file is older than it's prereqs and must always be rebuilt.

I have a fix that adds the first prereq of $(fpga_work_dir)/stamp into the --assume-new list and that forces make to always include the recipe for $(fpga_work_dir)/stamp in the replace-rtl.sh script.

timsnyder-siv:
https://app.circleci.com/pipelines/github/firesim/firesim/650/workflows/ce102317-1090-4b5d-bf30-4af8c421cf23/jobs/1122 looks like it is failing after I merged latest dev because rocket-chip and sodor have different copies of `vsrc/SimDTM.v`:
```
[error] (assembly) deduplicate: different file contents found in the following:
[error] /home/centos/project/target-design/chipyard/generators/rocket-chip/src/target/scala-2.12/rocketchip_2.12-1.0.jar:vsrc/SimDTM.v
[error] /home/centos/project/target-design/chipyard/generators/riscv-sodor/target/scala-2.12/sodor_2.12-2.0.jar:vsrc/SimDTM.v
```

I'll look into this and push a fix.

timsnyder-siv:
Given that the Chipyard Docs say that Chipyard will use the RocketChip SimDTM in https://chipyard.readthedocs.io/en/latest/Advanced-Concepts/Chip-Communication.html#using-the-debug-module-interface-dmi we probably want the rocketchip SimDTM.v in the FAT JAR.

image
shows the differences between rocket and sodor.  I haven't thought through them completely but they seem nontrivial to me.

timsnyder-siv:
Now https://app.circleci.com/pipelines/github/firesim/firesim/653/workflows/0076351c-f4ab-450b-bee0-8c6e6df6e534/jobs/1129 is failing because of the macro-paradise 2.1.0 not being available for scala 2.12.12 like I mentioned in https://github.com/firesim/firesim/pull/668#discussion_r541237713

I'm going to wait until we adress that issue in #668 and then remerge local-chisel34

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/firesim/aws-fpga-firesim
 [submodule "target-design/chipyard"]
 	path = target-design/chipyard
-	url = https://github.com/ucb-bar/chipyard
+	url = https://github.com/sifive/chipyard
```
timsnyder-siv, line 6: Shouldn't merge with this in place.  Wasn't sure how to get the CI to work on the PR without pointing to my fork of CY since I don't have write access to the ucb-bar fork.  Since "pull request authors can't request changes on their own PR", (nice feature bro), I'm going to add the DO-NOT-MERGE label on this until I revert the chipyard url change.

sim/build.sbt
```diff
@@ -4,16 +4,60 @@ import Tests._
...
+      sbtassembly.AssemblyUtils.sourceOfFileForMerge(tempDir, f) match {
+        case (jar, _, _, true) if jar.toString.contains("rocketchip") => f
+      }
+    }
+    if (filtered.size == 1) Right(Seq(filtered.head -> path))
+    else Left("multiple candidates have rocketchip in their jar name for target:" +
+      filenames(tempDir, filtered).mkString("\n", "\n", "")
+      )
+  }
+}
```
timsnyder-siv, line 51: If @jackkoenig or anyone else knows of a more Scalatastic way to write this so that the code doesn't have to be duplicated, I'd be all for it.  I was thinking that I potentially need to create a new class that extends `sbtassembly.MergeStrategy` with another Bool parameter that is used to control the polarity and select the name and error string.

Maybe I'm trying too hard though.

sim/build.sbt
timsnyder-siv: Done.

.gitmodules
timsnyder-siv: Removed after realizing that as long as I have a PR open on ucb-bar/chipyard with the commit, it will be available from the upstream chipyard.

sim/src/main/makefrag/fasedtests/Makefrag
timsnyder-siv: This comment no longer makes sense after the force push of the rebase.

sim/target-agnostic.mk
timsnyder-siv: it wasn't missed now that I rebased and squashed a bunch of stuff together... :)

timsnyder-siv:
I'm starting to rebase this forward again to latest dev.

timsnyder-siv:
Okay.  It's been a crazy few months but I've finally rebased this forward and run some sanity checking today (manually ran `CIGroupA` and buildafi for the multiclock single-core rocket).

@davidbiancolin , I think in its current form, this PR may break `ENABLE_SBT_THIN_CLIENT`.  However, I wanted to push what I have and see what the CI does.  When I ran CI locally, I had the two fased tests in `CIGroupA` fail when run under SBT but then when I would go manually run the make, they would pass.  I'm not sure what that meant.  I was probably doing something wrong in how I was running them manually.

I might also want to look at making FAT_JAR be able to be a full classpath instead of a single file.  I can why offline.  That work could also be done in a follow-on PR if needed.

timsnyder-siv:
So... what does it mean when `CIGroupB` times out?  The failure looked like this at the end:
```
Running: make -C /home/centos/project/sim run-verilator TARGET_PROJECT=fasedtests DESIGN=AXI4Fuzzer TARGET_CONFIG=LLCDRAMConfig PLATFORM_CONFIG=DefaultF1Config PLATFORM=f1
make[1]: Entering directory `/home/centos/project/sim'
/home/centos/project/sim/src/main/makefrag/fasedtests/Makefrag:47: target `/home/centos/project/sim/midas/targetutils/src/main/scala/midas/annotations.scala' given more than once in the same rule.
cd /home/centos/project/sim/generated-src/f1/AXI4Fuzzer-LLCDRAMConfig-DefaultF1Config && ./VAXI4Fuzzer +mm_relaxFunctionalModel_0=0 +mm_openPagePolicy_0=1 +mm_backendLatency_0=2 +mm_schedulerWindowSize_0=8 +mm_transactionQueueDepth_0=8 +mm_dramTimings_tAL_0=0 +mm_dramTimings_tCAS_0=14 +mm_dramTimings_tCMD_0=1 +mm_dramTimings_tCWD_0=10 +mm_dramTimings_tCCD_0=4 +mm_dramTimings_tFAW_0=25 +mm_dramTimings_tRAS_0=33 +mm_dramTimings_tREFI_0=7800 +mm_dramTimings_tRC_0=47 +mm_dramTimings_tRCD_0=14 +mm_dramTimings_tRFC_0=160 +mm_dramTimings_tRRD_0=8 +mm_dramTimings_tRP_0=14 +mm_dramTimings_tRTP_0=8 +mm_dramTimings_tRTRS_0=2 +mm_dramTimings_tWR_0=15 +mm_dramTimings_tWTR_0=8 +mm_rowAddr_offset_0=18 +mm_rowAddr_mask_0=65535 +mm_rankAddr_offset_0=16 +mm_rankAddr_mask_0=3 +mm_bankAddr_offset_0=13 +mm_bankAddr_mask_0=7 +mm_llc_wayBits_0=3 +mm_llc_setBits_0=12 +mm_llc_blockBits_0=7 +mm_llc_activeMSHRs_0=8 +dramsim +max-cycles=1000000000  2> err
make: *** [sbt] Terminated
make[1]: *** [run-verilator] Hangup
Running: make -C /home/centos/project/sim verilator TARGET_PROJECT=fasedtests DESIGN=AXI4Fuzzer TARGET_CONFIG=DefaultConfig PLATFORM_CONFIG=ConstrainedIdHostConfig PLATFORM=f1

Too long with no output (exceeded 2h0m0s): context deadline exceeded
```

timsnyder-siv:
I decided to rerun from failed to see if the timeout was just a transient CircleCI issue.

timsnyder-siv:
This is strange.  If I check out chipyard dev at https://github.com/ucb-bar/chipyard/commit/1b16e759d9af5f395e357a978d5745c984c8acfb and bump firesim to 9b817ed5822b7129b05d7cc5ab368edb1657ebbe, running `make run-verilator TARGET_PROJECT=fasedtests DESIGN=AXI4Fuzzer TARGET_CONFIG=LLCDRAMConfig PLATFORM_CONFIG=DefaultF1Config PLATFORM=f1` locally didn't finish for me.  I have a weird suspicion that something in Chipyard dev that hasn't been merged to firesim dev might be causing this.

I'm going to open another PR that just bumps chipyard to it's latest dev and see what happens in the CI.

timsnyder-siv:
I reproduced the CI failure by only bumping Chipyard in #764.  I'm going to build it with VCS and dump some waves to see what is going on.

timsnyder-siv:
Looks like building the fasedtests with vcs is a bit stale, I'm running into:
```
 simif_emul.o: In function `simif_emul_t::init(int, char**, bool)':
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:90: undefined reference to `context_t::current()'
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:92: undefined reference to `context_t::init(void (*)(void*), void*)'
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:95: undefined reference to `context_t::switch_to()'
simif_emul.o: In function `simif_emul_t::advance_target()':
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:127: undefined reference to `context_t::switch_to()'
simif_emul.o: In function `finish()':
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:38: undefined reference to `context_t::switch_to()'
simif_emul.o: In function `__static_initialization_and_destruction_0':
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:21: undefined reference to `context_t::context_t()'
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/simif_emul.cc:21: undefined reference to `context_t::~context_t()'
vcs-harness.o: In function `tick':
/scratch/tims/firesim_use_fat_jar/sim/midas/src/main/cc/emul/vcs-harness.cc:301: undefined reference to `context_t::switch_to()'
collect2: error: ld returned 1 exit status
Makefile:106: recipe for target 'product_timestamp' failed
make[2]: *** [product_timestamp] Error 1
make[2]: Leaving directory '/scratch/tims/firesim_use_fat_jar/sim/generated-src/f1/AXI4Fuzzer-LLCDRAMConfig-DefaultF1Config/AXI4Fuzzer-debug.csrc'
Make exited with status 2
Verdi KDB elaboration done and the database successfully generated: 0 error(s), 0 warning(s)
rtlsim/Makefrag-vcs:47: recipe for target '/scratch/tims/firesim_use_fat_jar/sim/generated-src/f1/AXI4Fuzzer-LLCDRAMConfig-DefaultF1Config/AXI4Fuzzer-debug' failed
```

sim/src/main/makefrag/firesim/Makefrag
```diff
@@ -30,22 +30,57 @@ VERILOG := $(GENERATED_DIR)/FPGATop.v
...
-$(FIRRTL_FILE) $(ANNO_FILE): $(SCALA_SOURCES) $(FIRRTL_JAR) $(SCALA_BUILDTOOL_DEPS)
-	mkdir -p $(@D)
-	$(call run_scala_main,$(firesim_sbt_project),chipyard.Generator,\
+BOOTROM_FILES = bootrom.rv32.img bootrom.rv64.img
+CHIPYARD_ROMDIR ?= $(chipyard_dir)/generators/testchipip/bootrom
+# pattern rule to handle copying bootrom images into GENERATED_DIR so that they are available for
+# - distributed elaboration on a build host
+# - dromajo co-sim
+BOOTROM_TARGETS = $(addprefix $(GENERATED_DIR)/, $(BOOTROM_FILES))
+$(BOOTROM_TARGETS): $(GENERATED_DIR)/bootrom.%.img: $(CHIPYARD_ROMDIR)/bootrom.%.img | $(GENERATED_DIR)
```
abejgonzalez, line 48: Im not opposed to this... but if we want to dedup even more we could make the `none` Simulator option in `GenerateSimFiles` (in Chipyard) just copy the bootrom... I think that should work with the FPGA prototyping stuff.

abejgonzalez, line 48: Im not opposed to this... but if we want to dedup even more we could make the `none` Simulator option in `GenerateSimFiles` (in Chipyard) just copy the bootrom... I think that should work with the FPGA prototyping stuff.

sim/Makefile
```diff
@@ -60,6 +65,8 @@ SBT ?= java $(JAVA_OPTS) -jar $(rocketchip_dir)/sbt-launch.jar $(SBT_OPTS) $(SBT
 define run_scala_main
 	cd $(base_dir) && $(SBT) ";project $(1); runMain $(2) $(3)"
 endef
+# recipes for generator and gg using sbt are calls out to FAT_JAR instead of SBT
+run_main = cd $(base_dir) && java $(JAVA_ARGS) -cp $(FAT_JAR) $(2) $(3)
```
abejgonzalez, line 69: Reminder to change all `JAVA_ARGS` to `JAVA_OPTS` to match the change done in Chipyard.

sim/src/main/makefrag/midasexamples/Makefrag
```diff
@@ -30,23 +30,38 @@ else
...
 		$(addprefix $(firesim_base_dir)/,. midas midas/targetutils firesim-lib) \
 		$(addprefix $(chipyard_dir)/generators/, chipyard rocket-chip/src, rocket-chip/api-config-chipsalliance)
 
 chisel_srcs = $(foreach submodule,$(chisel_src_dirs),\
 	$(shell find $(submodule)/ -iname "[!.]*.scala" -print 2> /dev/null | grep 'src/main/scala'))
 
+# don't search for implicit rules to make the chisel_srcs (reduces useless output in --debug=i)
+$(chisel_srcs): ;
+
+$(FAT_JAR):  $(chisel_srcs) $(SCALA_BUILDTOOL_DEPS)
```
abejgonzalez, line 47: Why do we care about shipping FAT jars for the fastedtests and midasexamples? Or is this just done to keep things consistent across projects.

timsnyder-siv: I was just trying to keep things consistent.  Even though it isn't needed right now, I didn't think it was worth having them be different.

sim/src/main/makefrag/firesim/Makefrag
timsnyder-siv: I'll look at that and if I'm able to move it in there, I'll move this target into my private TARGET_PROJECT_MAKEFRAG (which doesn't use the CY generator)

timsnyder-siv: Hey @abejgonzalez, when I took the copying out of the makefiles and sanity checked it, I run into this error:
```
cd /home/centos/chipyard_oss_210430 && java -Xmx8G -Xss8M -XX:MaxPermSize=256M -Djava.io.tmpdir=/home/centos/chipyard_oss_210430/.java_tmp -cp /home/centos/chipyard_oss_210430/sims/firesim/sim/generated-src/firechip.jar chipyard.Generator --target-dir /home/centos/chipyard_oss_210430/sims/firesim/sim/generated-src/f1/FireSim-FireSimRocketConfig-BaseF1Config --name firesim.firesim.FireSim.FireSimRocketConfig --top-module firesim.firesim.FireSim --legacy-configs firesim.firesim:FireSimRocketConfig
OpenJDK 64-Bit Server VM warning: Ignoring option MaxPermSize; support was removed in 8.0
Elaborating design...
L2 InclusiveCache Client Map:
	0 <= blkdev-tracker0
	1 <= serial-tl
	2 <= Core 0 DCache
	3 <= Core 0 ICache

Interrupt map (2 harts 2 interrupts):
  [1, 1] => frontend
  [2, 2] => uart_0

[error] java.nio.file.NoSuchFileException: /home/centos/chipyard_oss_210430/sims/firesim/sim/generated-src/f1/FireSim-FireSimRocketConfig-BaseF1Config/bootrom.rv64.img
[error] 	...
[error] 	at freechips.rocketchip.devices.tilelink.BootROM$.contents$lzycompute$1(BootROM.scala:75)
```

This seems strange but I guess it is probably related to why [`firesim.firesim.WithBootROM` didn't use the path at `chipyard/bootrom` and instead reached all the way down into testchipip to get the ROM img ](https://github.com/ucb-bar/chipyard/pull/872/files#diff-e4468b96461e4c7fcb103a17a6a01bc40296e9aeeb712d1ea4218d57b0dd3e80L28-L29)

timsnyder-siv: Oh, `GenerateSimFiles` is a separate app that the chipyard makefiles call before the generator runs?  @davidbiancolin, head's up, I'm going to add a call to the `chipyard.GenerateSimFiles` for the none simulator in place of this copying.

timsnyder-siv: @abejgonzalez , Do you guys have trouble with having changes happen to files copied by GenerateSimFiles not causing the copy to happen again?  Looking at the Chipyard makefiles, it seems that ones the `$(build_dir)/sim_files.f` is created, `GenerateSimFiles` will never be rerun, even if the resources it copied have changed.  Am I missing something?

timsnyder-siv: After thinking about this more, I feel like adding `GenerateSimFiles` usage to FireSim might not be the right way to go because:
1. it isn't being used by the FireSim simulation makefiles (MIDAS-level) and adding it only for the bootrom files might confuse or conflict with ML sim
2. It looks like it breaks the dependency information in the makefiles so that files used in `sim_files.f` aren't kept up to date inside of the `target_dir` (but maybe I'm missing some other part of the CY flow).

sim/Makefile
timsnyder-siv: addressed in 7c299da of the rebase I just force-pushed.

sim/src/main/makefrag/firesim/Makefrag
abejgonzalez: Hmmm. I think we just missed that... Ideally, the `GenerateSimFiles` target depends on the items it potentially copies over but that isn't the case. I suppose no-one has gotten to that issue until now. Now I wonder if instead of using `GenerateSimFiles` we just do the same thing in the Makefiles (copy files over depending on if make is called from VCS or Verilator, and generating the `.f`). 

For now, I think it's fine to ignore it.

deploy/buildtools/buildafi.py
```diff
@@ -84,51 +114,120 @@ def on_build_failure():
...
+                                      remote_dir=pjoin(aws_fpga_root, fpgabuilddir, ''),
+                                      ssh_opts="-o StrictHostKeyChecking=no", upload=False, extra_opts="-l",
+                                      capture=True)
+            rootLogger.debug(rsync_cap)
+            rootLogger.debug(rsync_cap.stderr)
+            local("mkdir -p {}/results-build/{}/".format(ddir, results_builddir))
+            local("cp {}/design/cl_firesim_generated.sv {}/results-build/{}/cl_firesim_generated.sv".format(
+                  pjoin(aws_fpga_root, fpgabuilddir),
+                  ddir,
+                  results_builddir))
```
timsnyder-siv, line 202: Testing the latest code today, depending on how `replace-rtl.sh` fails, it is possible for the `cl_firesim_generated.sv` to not have been created.  In those failure modes, we will bomb out here and the build hosts are still left running.  I'm thinking that I will put this local behind  `file-existance or elab_result != 0` so that if we're in the process of bailing out after a failure, we don't bail out too early.

davidbiancolin:
My builds seemed to work..

abejgonzalez:
@timsnyder-siv Were you planning on getting this into the next CY release (happening this week)? Or would you rather put this in the next release (aka merge this into dev sometime this summer)?

sim/target-agnostic.mk
```diff
@@ -75,16 +87,14 @@ $(VERILOG) $(HEADER) $(fame_annos): $(FIRRTL_FILE) $(ANNO_FILE) $(SCALA_BUILDTOO
 # to generate a runtime configuration that is compatible with the generated
 # hardware (BridgeModule). Useful for modelling a memory system that differs from the default.
 .PHONY: conf
-conf: $(fame_annos)
+conf: $(fame_annos) $(GEN_CLASSPATH)
```
timsnyder-siv, line 90: oops, I forgot to handle converting an actual classpath into a list of dependencies here.  Should be something like 
```suggestion
conf: $(fame_annos) $(subst :, ,$(GEN_CLASSPATH))
```

Otherwise passing in an actual colon separated classpath will hit the "multiple pattern targets" error that I always have to lookup the meaning for when I hit it (see https://stackoverflow.com/questions/2100448/multiple-target-patterns-makefile-error)

timsnyder-siv: Resolved in cd39ff3

timsnyder-siv:
The implementation so far in this PR has a requirement for `local_dir == remote_dir` when using distributed elaboration.  This stems from how the `replace-rtl.sh` script is generated from the makefiles and the fact that the makefiles themselves use absolute paths.  I'm rethinking this limitation because I'd like to be able to use multiple volumes on my firesim manager (perhaps separating the root volume containing system software,  from project volumes that contain clones and experimental data).

I propose that we augment the makefiles to accept a `REMOTE_PREFIX` variable that gets pre-pended to all of the paths used in `replace-rtl` so that the generated `replace-rtl.sh` can continue to use absolute paths and also be able to support the manager wanting to put the working directory in a volume it knows exists on the build hosts.

@davidbiancolin do we want to handle this case in this PR or do it in a follow-on?

PR #652, opened 2020-10-28
Promote Passthrough Combinational Paths Through Models

davidbiancolin:
When bridges or modules destined for optimization are promoted, they tend to leave paths that run through their original instantiation site. As a result, direct connections between bridges and models must combinationally propagate through the hub model, potentially increasing FMR.

To be clear: 
Bridge -> Modules annotated with a BridgeAnnotation  (these are extracted from the target completely)
Model -> Modules annotated with a FAMEModelAnnotation (these remain in the transformed RTL)

At its heart, this PR optimizes this case by allowing Pipe-type to channels to "fan-out" instead of having to propagate serially through multiple models. 

This optimization works as follows:
1) `PromotePassthroughConnections` finds inter-model passthrough connectivity in the target (excluding clocks) and promotes it out of all models into the top level wrapper. There will now be connect statements in the top-level that share the same RHS (i.e., there is a fanout from common source model or bridge)

2) `ChannelExcision` can no longer assume all channel are to the hub model, nor that connections are 1:1, and has been generalized to reflect this.
- When a model drives multiple model sinks, a single source and multiple sinks are created. As before, FCCAs are retargeted to the new IO. If a channel fans out, the targeted FCCAs share the same "source" field. 
- When a connection to multiple models is driven by a bridge, the original bridge-emitted FCCA is duplicated. The group of channels is labelled as a fanout using the [[FAMEChannelFanoutAnnotation]]. Unlike a model-driven channel, these FCCAs have sources = None; and thus the extra annotation is required.

3) FindDefaultClocks has been adapted to remove the assumption that clock connects must exist between all models and the hub. This is needed for my dissertation work. 

4) `FAMETransform` has been adapted, mostly to change how the top is rewired
 - Inter-bridge channels do not drive any model; these need to be separately rewired (just a connect statement)
 - Transforms vestigial sources (that have been promoted to a fanout) into dummy wires, so they can still appear in expressions local to that model. No output channel port is generated.

5) SimWrapper generation emits forking channels to handle fanouts. All queues for each channel in a fanout must be ready to accept a new token before the source is dequeued from. 

Other changes:
- Updated performance summary
- Added a bridge that can be used as a random noise generator. 
- Midasexamples to check for FMR improvement resulting from passthrough optimizations.  

Thoughts / Questions:
- This PR could be dramatically simplified if bridges weren't extracted completely and instead handled like other models (as they are just models with host-side Chisel generators). This would require working the debug synthesis passes. In a future PR we could consider moving Bridge Extraction after the FAME Transform. 
- Should we add a switch to remove the passthrough optimization?


sim/midas/src/main/scala/midas/passes/fame/PromotePassthroughConnections.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.collection.mutable
+
+/** Removes all non-passthrough expressions in a circuit so that we can use
+  *  CheckCombLoops to return only passthrough connectivity (this represents
+  *  subset of of all combinationally connected paths).
+  *
+  * @param state the CircuitState to simplify
+  */
+object RemoveNonWirePrimitives {
+  def apply(state: CircuitState): CircuitState = {
```
albert-magyar, line 24: This should probably work on a `Circuit`, rather than a `CircuitState`, since it's not part of a coherent transform.

sim/midas/src/main/scala/midas/passes/fame/PromotePassthroughConnections.scala
```diff
@@ -0,0 +1,136 @@
...
+  override def optionalPrerequisiteOf = Seq(Dependency[FAMEDefaults])
+
+  def execute(state: CircuitState): CircuitState = {
+    val moduleMap = state.circuit.modules.map(m => m.name -> m).toMap
+    val topModule = moduleMap(state.circuit.main)
+
+    def collectInstances(insts: mutable.ArrayBuffer[WDefInstance])(s: Statement): Unit = {
+      s match {
+        case wdef: WDefInstance => insts += wdef
+        case o => Nil
```
albert-magyar, line 73:
```suggestion
        case _ => s.foreach(collectInstances(insts))
```

Golf

sim/midas/src/main/scala/midas/passes/fame/PromotePassthroughConnections.scala
```diff
@@ -0,0 +1,136 @@
...
+      }
```
albert-magyar, line 74:
```suggestion
```

sim/midas/src/main/scala/midas/passes/fame/PromotePassthroughConnections.scala
```diff
@@ -0,0 +1,136 @@
...
+      s.foreach(collectInstances(insts))
```
albert-magyar, line 75:
```suggestion
```

sim/midas/src/main/scala/midas/passes/fame/PromotePassthroughConnections.scala
```diff
@@ -0,0 +1,136 @@
...
+        }
+        sinkNode match {
+          case Some(sinkNode) =>
+            val sourceNode = sink2sourceMap(sinkNode)
+            val newRHS = sourceNode match {
+              case LogicNode(portName, Some(instName), None) => WSubField(WRef(instName), portName)
+              case LogicNode(portName, None, None) => WRef(portName)
+              case o => throw new Exception(s"memport field of source LogicNode should be unset.") 
+            }
+            c.copy(expr = newRHS)
```
albert-magyar, line 126: This approach (anything driven by a passthrough gets directly driven by the upstream source) is reasonable. Since it happens before channel excision, it also looks like we could use DCE to delete any newly-unused wires.

sim/midas/src/main/scala/midas/passes/fame/FindDefaultClocks.scala
```diff
@@ -6,46 +6,96 @@ import firrtl._
...
+        fcca.copy(clock = Some(getDefaultClock(clockGraph, sourceModelInst)))
       } else {
         fcca
       }
     case a => a
   }
 
+  // Hub self-loops can be introduced as a result of extracting passthroughs in
+  // satellite models.  For these channels, a default clock cannot be inferred by
+  // looking at Connect statements.  Instead find a second channel that shares the
```
albert-magyar, line 83: Is this always guaranteed to exist?

sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -182,6 +197,18 @@ class NestedModelsF1Test extends TutorialSuite("NestedModels")
 
 class MultiRegF1Test extends TutorialSuite("MultiReg")
 
+class PassthroughModelTest extends TutorialSuite("PassthroughModel") {
+  expectedFMR(2.0)
+}
+
+class PassthroughModelNestedTest extends TutorialSuite("PassthroughModelNested") {
+  expectedFMR(2.0)
```
albert-magyar, line 205: This is very handy. I see that it goes up to 5.0 if the passthroughs are left in.

albert-magyar (approved):
LGTM

sim/midas/src/main/scala/midas/passes/fame/FindDefaultClocks.scala
davidbiancolin: The way i've currently written this, the model from which the passthrough was extracted should remain a sink in a fanout inclusive of this self-loop, even if it's unused the satellite model.

PR #658, opened 2020-11-11
Update to Chisel 3.4

timsnyder-siv:
These are the changes I needed to make when updating FireSim to work with Chisel 3.4 and FIRRTL 1.4.  I'm going to point to this branch in a ChipYard PR that will also point to the version of RC and barstools that I know to be capable of running Chisel->GoldenGate->Verilog without blowing up (and was passing midas-level `run-asm-tests` but that isn't being regressed yet).

NOTE: I also need to pass `--no-dce` when running GoldenGate because I'm hitting a case where DCE seems to be removing a dont_touch'd AutoCounter node.

### DIFF
--- a/sim/midas/src/main/scala/midas/core/SimWrapper.scala
+++ b/sim/midas/src/main/scala/midas/core/SimWrapper.scala
@@ -16,15 +16,15 @@ import chisel3._
 import chisel3.util._
 import chisel3.experimental.{Direction, chiselName, ChiselAnnotation, annotate}
 import chisel3.experimental.DataMirror.directionOf
-import firrtl.annotations.{SingleTargetAnnotation, ReferenceTarget}
+import firrtl.annotations.{SingleTargetAnnotation, ReferenceTarget, IsModule}
 
 import scala.collection.immutable.ListMap
 import scala.collection.mutable.{ArrayBuffer}
 
 case object SimWrapperKey extends Field[SimWrapperConfig]
 
-private[midas] case class TargetBoxAnnotation(target: ReferenceTarget) extends SingleTargetAnnotation[ReferenceTarget] {
-  def duplicate(rt: ReferenceTarget): TargetBoxAnnotation = TargetBoxAnnotation(rt)
+private[midas] case class TargetBoxAnnotation(target: IsModule) extends SingleTargetAnnotation[IsModule] {
+  def duplicate(rt: IsModule): TargetBoxAnnotation = TargetBoxAnnotation(rt)
 }
 
 class SimReadyValidRecord(es: Seq[(String, ReadyValidIO[Data])]) extends Record {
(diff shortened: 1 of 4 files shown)


davidbiancolin:
Is the DCE problem reproducible with the default make target?

timsnyder-siv:
It's been a few days since I ran into the DCE thing and started disabling it (don't remember if the default target hits it or if it requires one of our bigger designs).  I told @jackkoenig  that I was going to try to get back to giving him a repro later today or tomorrow so he can debug and fix.

timsnyder-siv:
@davidbiancolin I'm now discovering that the PlusArgReaderRemoval pass doesn't seem to be working in my chisel3.4 setup.  I'm just starting to look into it.

timsnyder-siv:
@davidbiancolin , the issue I'm seeing with the PlusArgReaderPass is related to needing to use `--no-dce`.  It looks like DCE is the pass that fully removes the plusarg_readers out of the design.  Even though all of my plusarg_reader's are defined in the emitted Verilog, I'm ending up with plusarg_reader.v in my firrtl_black_box_resource_files.f and that causes an error with multiple definitions of a module when I try to build verilator (and I assume vivado will freak out as well).

Do you know why the PlusArgReaderPass is needed?  Defining SYNTHESIS should be enough to make vivado and any other SYNTHESIS tools happy when I look at https://github.com/chipsalliance/rocket-chip/blob/6eb1a3de082e27c752d9e4c1ae971c693cc192eb/src/main/resources/vsrc/plusarg_reader.v

davidbiancolin:
> @davidbiancolin , the issue I'm seeing with the PlusArgReaderPass is related to needing to use `--no-dce`. It looks like DCE is the pass that fully removes the plusarg_readers out of the design. Even though all of my plusarg_reader's are defined in the emitted Verilog, I'm ending up with plusarg_reader.v in my firrtl_black_box_resource_files.f and that causes an error with multiple definitions of a module when I try to build verilator (and I assume vivado will freak out as well).
> 
> Do you know why the PlusArgReaderPass is needed? Defining SYNTHESIS should be enough to make vivado and any other SYNTHESIS tools happy when I look at https://github.com/chipsalliance/rocket-chip/blob/6eb1a3de082e27c752d9e4c1ae971c693cc192eb/src/main/resources/vsrc/plusarg_reader.v

I don't remember the exact motivations for it; it was one of the final things donggyu upstreamed from his dev branches. IIRC it was problematic for downstream transformations -- I suspect it can be removed.

timsnyder-siv:
> I don't remember the exact motivations for it; it was one of the final things donggyu upstreamed from his dev branches. IIRC it was problematic for downstream transformations -- I suspect it can be removed.

Ok.  I'll experiment with removing PlusArgReaderPass and putting `define SYNTHESIS in the right spot.  If it works, I'll add the required changes to this PR.

timsnyder-siv:
These commits were merged into the local-chisel34 branch by @abejgonzalez outside of the PR flow.  I'm going to close this as unmerged and move over to using local-chisel34

PR #661, opened 2020-11-16
cannot import name docevents

apaj:
The 
```
yum install python-pip
pip install boto3
```
sequence wouldn't let me run successfully aws-setup. It turns out that the developers of the boto package did something (described here in detail: [https://github.com/boto/boto3/issues/2596](https://github.com/boto/boto3/issues/2596)) that requires the awscli package to be upgraded via pip.


davidbiancolin (approved):
LGTM. Thanks for the PR and the detailed commit message. :)

PR #662, opened 2020-11-20
Provide a Simple Mechanism For Triaging Hanging Simulators

davidbiancolin:
This just keeps coming up. It's probably a misuse of the term to call this a watchdog unless we put it on a timer; i'm open to other names. `heartbeat.csv` can also double as a means to generate an FMR / effective target frequency plot.

It's not clear to me what the plus args for this thing should be, we probably want one to disable it, and one to change the polling interval.


davidbiancolin:
@timsnyder-siv always interested in your comments.

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/Debugging-Hanging-Simulators.rst
```diff
@@ -0,0 +1,61 @@
+.. _debugging-hanging-simulators:
+
+Debugging a Hanging Simulator
+=============================
+
+A common symptom of a failing simulation is that appears to
```
alonamid, line 6:
```suggestion
A common symptom of a failing simulation is that it appears to
```

sim/midas/src/main/cc/bridges/host_watchdog.cc
```diff
@@ -0,0 +1,29 @@
...
+void host_watchdog_t::tick(){
+    if (trip_count == polling_interval)  {
+        trip_count = 0;
+        uint64_t current_cycle = sim->actual_tcycle();
+        has_timed_out |= current_cycle == last_cycle;
+
+        time_t current_time;
+        time(&current_time);
+        struct tm * tm_local = localtime(&current_time);
+        log << current_cycle << ", " << current_time - start_time << std::endl;
```
alonamid, line 24: I was originally concerned the file size might get very big for long simulations, but after doing the estimate math it seems to be fine

sim/midas/src/main/cc/bridges/host_watchdog.cc
```diff
@@ -0,0 +1,29 @@
+#include "host_watchdog.h"
+
+host_watchdog_t::host_watchdog_t(simif_t* sim, std::vector<std::string> &args):
+        bridge_driver_t(sim),
+        sim(sim) {
+    log.open("heartbeat.csv", std::ios_base::out);
+    if (!log.is_open()) {
+        fprintf(stderr, "Could not heartbeat output file.\n");
```
alonamid, line 8:
```suggestion
        fprintf(stderr, "Could not open heartbeat output file.\n");
```

alonamid (approved):
LGTM other than the possible typos. I hope the timeout isn't going to come back to bite us

davidbiancolin:
> LGTM other than the possible typos. I hope the timeout isn't going to come back to bite us

Maybe if we have dessert or something. But I doubt it will, because if the hot-loop is free-spinning without making any forward progress you almost definitely have a problem If this we're a timer-driven watchdog (which is probably should be) then your concern would be well justified.

alonamid:
Is it possible that Dromajo co-sim could cause it to stall?

davidbiancolin:
> Is it possible that Dromajo co-sim could cause it to stall?

No because the way it detects stall is by counting transits through the main loop. Dromajo can take as long as it needs, since it's tick is blocking afaict.  We'll need to reevaluate this when we go multithreaded.

timsnyder-siv (commented):
This is great information.  Plusarg for polling interval that can be set to -1 seems like enough config to me.  If the file size does become an issue, it might be useful to be able to specify a maximum size for the logfile.

heartbeat might be better than watchdog as far as a name goes.

PR #663, opened 2020-11-20
Add DRAMMasEnums.backendLatencyBits

timsnyder-siv:
It is ocasionally useful to set mm_backendLatency to values larger than typical
DRAM Timing parameters.  Create a unique DRAMMaxEnums.backendLatencyBits to
describe the width of mm_backendLatency


davidbiancolin (approved):
LGTM. Can you add a comment in the dev->master PR for the release notes, or do you not have permission to do so?

PR #664, opened 2020-11-20
Enhance info print statements for AutoCounter

timsnyder-siv:
@davidbiancolin these are extra println statements that I found useful when starting to use the AutoCounter feature.


sim/midas/src/main/scala/midas/widgets/Lib.scala
```diff
@@ -208,7 +208,7 @@ class MCRFileMap() {
   private val regList = ArrayBuffer[MCRMapEntry]()
 
   def allocate(entry: MCRMapEntry): Int = {
-    Predef.assert(!name2addr.contains(entry.name), "name already allocated")
+    Predef.assert(!name2addr.contains(entry.name), s"name already allocated '${entry.name}'")
```
davidbiancolin, line 211: You ran into this eh. Where was the collision?

timsnyder-siv: I think it was because we had coverpoints with the same label in the same module.

sim/midas/src/main/scala/midas/passes/AutoCounterCoverTransform.scala
```diff
@@ -225,17 +226,24 @@ class AutoCounterTransform extends Transform with AutoCounterConsts {
       case o => remainingAnnos += o
     }
 
+    println(s"[AutoCounter] There are ${counterAnnos.length} counterAnnos available for selection in the following modules:")
+    counterAnnos.map(_.target.module).distinct.foreach({ i => println(s"  ${i}") })
+
     //extract the module names from the methods mentioned previously
     val covermodulesnames = moduleAnnos.map(_.target.module).distinct
+    println("[AutoCounter] covermodulesnames are:")
```
alonamid, line 234: I'm generally in favor of giving a more "human readable" descriptions in log lines. i.e. something like "Selected Modules for cover-function based AutoCounter annotation" rather than "covermodulesnames"

alonamid (commented):
Thanks for this! I'm always in favor of more informative logging/debug. LGTM other than that one comment.

sim/midas/src/main/scala/midas/passes/AutoCounterCoverTransform.scala
timsnyder-siv: Good point @alonamid , what do you think of :
```suggestion
    println("[AutoCounter] selected modules for cover-function based annotation:")
```

alonamid: works for me

alonamid (approved):

PR #665, opened 2020-11-24
Enable customruntimeconfig to be relative path above firesim

timsnyder-siv:
When using FireSim as a library in another project, it is useful to keep
configuration files in the repo of the toplevel project. This is a minimal
change that enables a user to provide a path like ../../../toplevel-configs/somegreat.conf

infrasetup already copies the file correctly.  This change only modifies
RuntimeHWConfig.get_local_runtimeconf_binaryname() to explicitly strip the
path to os.path.basename


timsnyder-siv:
@davidbiancolin I think this is the minimal change needed so that I can store my custom runtime configs in my toplevel repo instead of inside `firesim/sim/custom-runtime-configs`.  I know we had talked briefly about making [runtools.runtime_config.CUSTOM_RUNTIMECONFS_BASE](https://github.com/sifive/firesim/blob/allow-relpaths-in-customruntimeconfig/deploy/runtools/runtime_config.py#L22) configurable (or perhaps even making it a list of paths that is searched in order).  I wasn't sure where we would want to configure such a search path because config_hwdb.ini doesn't have a 'global' section and I wouldn't want to specify it in more than one place.

Let me know what you think.

davidbiancolin (commented):
Looks good. Could make sure the docs are updated?

timsnyder-siv:
Sure.  When I read through them, I thought my change is almost a bugfix that makes the code behave the way it is currently documented:
https://github.com/sifive/firesim/blame/allow-relpaths-in-customruntimeconfig/docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst#L499-L506

Specifically that they say `Then, set this field to the relative name of the config.`  I think they can be simplified to say that you can set that to a path relative to sim/custom-runtime-configs.  

It might be a good time to incorporate some of the things we talked about back in June in https://groups.google.com/g/firesim/c/PjTQXR5yreA  Specifically, I'm thinking of mentioning:
- default runtime.conf assumes 3.2GHz core clock (is it assuming 2:1 and that the periphery clock is running at 1.6GHz) and the default runtime.conf does not automatically adjust DRAM timing if you modify the clock frequency of FASED
- runtime.conf is used to add extra plusargs to the FireSim cmdline that are currently used to provide runtime configuration to FASED  
- `make -C sim conf` target being used to create a runtimeconf

Are you still planning to add more docs about FASED?

davidbiancolin (approved):
LGTM

davidbiancolin:
> Sure. When I read through them, I thought my change is almost a bugfix that makes the code behave the way it is currently documented:
> https://github.com/sifive/firesim/blame/allow-relpaths-in-customruntimeconfig/docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst#L499-L506
> Specifically that they say `Then, set this field to the relative name of the config.` I think they can be simplified to say that you can set that to a path relative to sim/custom-runtime-configs.

Ah, true.
> It might be a good time to incorporate some of the things we talked about back in June in https://groups.google.com/g/firesim/c/PjTQXR5yreA Specifically, I'm thinking of mentioning:
> 
> * default runtime.conf assumes 3.2GHz core clock (is it assuming 2:1 and that the periphery clock is running at 1.6GHz) and the default runtime.conf does not automatically adjust DRAM timing if you modify the clock frequency of FASED

This is changed now on dev such that FASED runs at 1GHz. But indeed.

> * runtime.conf is used to add extra plusargs to the FireSim cmdline that are currently used to provide runtime configuration to FASED
> * `make -C sim conf` target being used to create a runtimeconf
> 
> Are you still planning to add more docs about FASED?

That needs to happen at some point, but all my writing bandwidth is going to the dissertation right now. I'll try to write the beginnings of something with those points.

timsnyder-siv:
> That needs to happen at some point, but all my writing bandwidth is going to the dissertation right now. I'll try to write the beginnings of something with those points.

No rush.  I was asking because I thought I might be able to take a stab at it if you haven't been able to yet.

PR #666, opened 2020-11-25
Add init-submodules.sh step to simulation steps

hngenc:


davidbiancolin (approved):
Quite the auspicious PR number; good thing there's no code here. :D

LGTM

Issue #605, opened 2020-07-15
Gracefully Remove Autocounter Annotations When Not Enabled


davidbiancolin:
Presently, if AutoCounter is not enabled, AutoCounterAnnotations present in the source will produce renaming errors. It might not be obvious to the user this is being created by an AutoCounterAnnotation if they didn't add it themselves. The compiler should gracefully remove these, perhaps warning the user it is doing so.

davidbiancolin:
Alon what you have to do here is:
1) Convert the firrtl autocounter anntotations to use "HasDontTouches" and remove the DontTouchs in the annotator utilities.
See [this trait](https://github.com/freechipsproject/firrtl/blob/ecb96e83324ea17cf38b7b90753d745d3c7f51bd/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala#L19). There's also `DontTouchAllTargets`which may be acceptable.

2) Delete the Autocounter annotations at the end of the transform.

timsnyder-siv:
I think that there may also be a transform ordering issue that has to be dealt with.  The renaming error I ran into happens when DeadCodeElimination is run because it wants to remove the dead, unused AutoCounter signals.  The AutoCounter transform currently happens after DeadCodeElimination (see [MidasTransforms.scala](https://github.com/firesim/firesim/blob/659ddbf418d033f3b2fa3edb5bce46479dd0e3be/sim/midas/src/main/scala/midas/passes/MidasTransforms.scala#L39-L50))

This might be easier to do than fixing chipsalliance/firrtl#1997 as a quick way to avoid [the intersection of dead blackboxes and the ILA flow](https://forums.xilinx.com/t5/Synthesis/synth-design-no-longer-removes-unreachable-blackboxes-Synth-8/td-p/1181298).

@alonamid have you started looking at this?  I might be motivated to give it a try (instead of writing a script that will do a lame text delete on the dead blackboxes I know about in my design).

timsnyder-siv:
You can observe the assertion in 'make -C sim replace-rtl' using the local-chisel34 branch after reverting 5ccad8ab9bb6f175b4bbd6e0e46b74c21fee31ae  (that commit adds --no-dce to the GodenGate invocation to avoid this issue)

alonamid:
@timsnyder-siv , I was planning to start looking at this late next week, but if you have time to give it a try beforehand then go ahead!

[...]

PR #668, opened 2020-12-01
Bump to Chisel 3.4

timsnyder-siv:
Supersedes #658 using the local-chisel34 branch @abejgonzalez created.


timsnyder-siv:
@abejgonzalez head's up, I have more things that I want to push to this branch.  Should I bump FireSim on the Chipyard local-chisel34 branch as well?

abejgonzalez:
Yes. That would be great.

sim/firesim-lib/src/main/scala/passes/EC2F1Artefacts.scala
```diff
@@ -31,7 +31,7 @@ object EC2F1Artefacts extends Transform {
   // Capture FPGA-toolflow related verilog defines
   def generateHostVerilogHeader(targetDir: File) {
     val headerName = "cl_firesim_generated_defines.vh"
-    writeOutputFile(new File(targetDir, headerName), s"\n")
+    writeOutputFile(new File(targetDir, headerName), s"`define SYNTHESIS\n")
```
timsnyder-siv, line 34: @davidbiancolin I'm not sure that we really want to do this here because this is supposed to be things that the host platform requires.  Anyway, just wanted to cal it out

abejgonzalez:
A note about thin client: You cannot run the CI with the thin client... it looks like you can't call SBT from within SBT otherwise it just hangs. Is that a problem for normal FireSim use?

davidbiancolin:
> A note about thin client: You cannot run the CI with the thin client... it looks like you can't call SBT from within SBT otherwise it just hangs. Is that a problem for normal FireSim use?

In my flow i often run tests-in-the-loop, what you're describing sounds like it would break my flow. 

basically i type
`make sbt TARGET_PROJECT=midasexamples`

And then
`~testOnly firesim.midasexamples.<sometest>`

abejgonzalez:
Well that sucks... I'll look into it and see if there is a nice solution. If not I'll check if bloop works from the FireSim side still.

davidbiancolin:
@timsnyder-siv 4ec71df Works around the DCE issue by fixing how we manage DontTouches for targetutils annotations. We were planning to do this anyways. See #605.

davidbiancolin:
A word of caution, but CI is pretty broken. I've yet to do an accounting of all failures.

sim/firesim-lib/src/main/scala/passes/EC2F1Artefacts.scala
davidbiancolin: This seems like a reasonable place.

davidbiancolin:
CI is borked because of SBT changes, but the current failures are:
[error] Error during tests:
[error]         firesim.midasexamples.MultiSRAMF1Test
[error]         firesim.midasexamples.TrivialMulticlockF1Test
[error]         firesim.fasedtests.AXI4FuzzerMultiChannelTest
[error]         firesim.midasexamples.MultiRegF1Test
[error]         firesim.midasexamples.MultiRegfileF1Test
[error]         firesim.midasexamples.TwoAddersF1Test
[error]         firesim.midasexamples.ParityF1Test
[error]         firesim.midasexamples.NestedModelsF1Test
[error]         firesim.fasedtests.BaselineMultichannelTest
[error]         firesim.midasexamples.TriggerWiringModuleF1Test
[error]         firesim.midasexamples.GCDF1Test

abejgonzalez:
> CI is borked because of SBT changes, but the current failures are:

I think you just need to set `JAVA_ARGS` for the CI runs... but I'm not too sure.

jackkoenig:
Note that Chisel 3.4.1 is out. I intend to bump rocket-chip to using it this week. I also intend to use the published artifacts.

sim/firesim-lib/src/main/scala/passes/EC2F1Artefacts.scala
timsnyder-siv: I was looking at my generated verilog yesterday and noticed that the emitted Verilog for assertions uses `ifundef SYNTHESIS`.  I need to look at it more closely this morning and make sure that `SYNTHESIS` isn't causing the assertions to be synthesized incorrectly in FireSim.

timsnyder-siv: As I thought, [Vivado seems to define SYNTHESIS on it's own](https://forums.xilinx.com/t5/Synthesis/Does-vivado-define-word-quot-SYNTHESIS-quot-automatically-How/m-p/815280#M23524)

Most other synthesis tools I've used in the past did this as well, I just didn't find a reference that indicated Vivado did the right thing before I added it.  

@davidbiancolin, I think we should probably back this one line out, what do you think?
```suggestion
    writeOutputFile(new File(targetDir, headerName), s"\n")
```

davidbiancolin: Yup, let's apply this change.

davidbiancolin:
> Note that Chisel 3.4.1 is out. I intend to bump rocket-chip to using it this week. I also intend to use the published artifacts.

@jackkoenig Firrtl.testutils aren't published right?

sim/firesim-lib/src/main/scala/configs/CompilerConfigs.scala
```diff
@@ -83,7 +83,6 @@ class WithAutoCounterPrintf extends Config((site, here, up) => {
 class BaseF1Config extends Config(
   new WithWiringTransform ++
   new WithAsyncResetReplacement ++
-  new WithPlusArgReaderRemoval ++
```
davidbiancolin, line 86: @timsnyder-siv Remind me why this had to be removed again?

timsnyder: From the commit message when I took it out ff837e5a0dd0b5b34204905b720ca9755c5f23b3:

> The PlusArgReaderPass changes extmodule plusarg_reader to
a module that drives its output from DEFAULT but does not remove
BlackBox* Annotations.  It relies on DeadCodeElimination coming along
later to clean up the [dead blackboxes]
> 
> Rather than using the PlusArgReaderPass, the verilog SYNTHESIS `define
accomplishes the same effect in the plusarg_reader blackbox verilog.

Now that we have DCE re-enabled it might be possible to put this back in.  Want me to check?

My question would be why put it back in?  The PlusArgReader's don't seem to cause problems for downstream transforms (or there are silent problems that need to be made noisier).  They are also useful during MIDAS-level simulation.

davidbiancolin: Yeah it seems to me that everything works fine now with it out. We put it in initially because MIDAS couldn't process black boxes. If the PlusArgReader interacts sanely with ML sim I see little reason to remove it.

sim/build.sbt
```diff
@@ -8,12 +8,9 @@ lazy val commonSettings = Seq(
...
-  traceLevel   := 15,
   scalacOptions ++= Seq("-deprecation","-unchecked","-Xsource:2.11"),
-  libraryDependencies += "org.scalatest" %% "scalatest" % "3.1.0" % "test",
-  // Remove after bump to FIRRTL 1.3
-  libraryDependencies += "org.scalatestplus" %% "scalacheck-1-14" % "3.1.0.1" % "test",
-  libraryDependencies += "org.json4s" %% "json4s-native" % "3.6.1",
+  libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.2" % "test",
+  libraryDependencies += "org.json4s" %% "json4s-native" % "3.6.10",
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
   addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
```
timsnyder-siv, line 15:
```suggestion
  addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
```
The recent bump of scalaVersion to 2.12.12 for mdf causes me to hit:
```
[error] stack trace is suppressed; run last mdf / update for the full output
[error] (mdf / update) sbt.librarymanagement.ResolveException: Error downloading org.scalamacros:paradise_2.12.12:2.1.0
[error]   Not found
[error]   Not found
[error]   not found: /home/centos/.ivy2/local/org.scalamacros/paradise_2.12.12/2.1.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: https://oss.sonatype.org/content/repositories/snapshots/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: https://oss.sonatype.org/content/repositories/releases/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: /home/centos/.m2/repository/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
```

Bumping paradise to version 2.1.1 in CY fixes the issue.  We may want to bump to 2.1.1 here as well.

abejgonzalez (approved):
LGTM

PR #671, opened 2020-12-05
Allow instance threading by non-power-of-two factors

albert-magyar:
This change allows the threading optimization to handle threading by non-power-of-two factors for all types of designs. It conditionally generates slightly more complicated logic to translate the addresses of memory operations to reflect the true location of a particular instance thread's data.

Unfortunately, the `mv` + moderate changes to `ImplementThreadedMems.scala` did not make a nice diff on GitHub, but the changes are not quite as big as they look.

This replaces 081dfef, which was a WIP commit that I inadvertently left in when I came back after not looking at #639 for a little too long.


sim/midas/src/main/scala/midas/passes/fame/ImplementThreadedMems.scala
```diff
@@ -0,0 +1,172 @@
...
+
+object ImplementThreadedMems {
+  private def wsf(e: Expression, f: String): WSubField = WSubField(e, f)
+  private def wssf(e: Expression, f0: String, f1: String): WSubField = WSubField(wsf(e, f0), f1)
+  private def prim(op: PrimOp, args: Seq[Expression]) = DoPrim(op, args, Nil, UnknownType)
+
+  private def implement(tMem: ThreadedMem, moduleName: String): Module = {
+    val info = FAME5Info.info
+    val tIdxMax = UIntLiteral(tMem.nThreads-1)
+    val bramToBufferPipeDepth = if (tMem.nThreads < 4) 0 else 1
```
davidbiancolin, line 45: I know this existed previously, but it would be comment why this threshold is what it is.

davidbiancolin (approved):
Chilllllllll. 

If the docs and release notes mention a power-of-two restriction can you update that too.

sim/midas/src/main/scala/midas/passes/fame/ImplementThreadedMems.scala
albert-magyar: Good idea

albert-magyar:
> Chilllllllll.
> 
> If the docs and release notes mention a power-of-two restriction can you update that too.

Fortunately, they don't. Agile :)

PR #672, opened 2020-12-07
Add faster async-read mem model for lower FMR penalty

albert-magyar:


timsnyder-siv:
I think this is the PR @davidbiancolin mentioned would improve the FMR if I try to start using the MCRams optimization, please correct me if I've wandered into a different PR.  @albert-magyar, if I want to try this out, I need to do something to make the `LabelSRAMMacros` pass use this model, right?  I'm starting to look at that code but if you could give me a pointer on how to try this out, I'd really appreciate it!

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
```diff
@@ -0,0 +1,138 @@
...
+  // Divide target writes as evenly as possible into nPhysWrites banks
+  val wBankMaxSize = (channels.write_cmds.length + nPhysWrites - 1) / nPhysWrites
+
+  // Pick the winning write for each write bank
+  //   - Hook it up to the underlying memory's corresponding port
+  //   - Mark it done
+  (writes.grouped(wBankMaxSize).toSeq zip physWriteCmds).foreach {
+    case (wBank, underlyingWC) =>
+      underlyingWC.en := false.B
+      underlyingWC.mask := true.B
```
davidbiancolin, line 95: Is this not overridden somewhere? If it's unsupported you should add some assertions.

davidbiancolin (changes requested):
Super clean implementation. 

Is there a page in your dissertation with a block diagram for this? It might be nice to add a comment in the code that references it.

Can you explain to me what's happening with the write mask? And can you add requires for parameterization corners that are unsupported (if applicable).

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
albert-magyar: What thing being unsupported are you referring to?

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
```diff
@@ -0,0 +1,138 @@
+package midas.models.sram
+
+import chisel3._
+import chisel3.util.{PriorityEncoderOH, DecoupledIO, log2Ceil}
+
+object BankedAsyncMemChiselModel {
+  /**
+    * NOTE: Enabling read skips can change simulation results for buggy targets that use undefined read data
```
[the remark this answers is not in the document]
albert-magyar: This implementation is sensitive to the order of arrival of write tokens in terms of which write wins for write-write collisions (a separate distinction from that in the note). As long as all the write ports follow the same "model path" (which will generally just go from one model containing the CPU to a regfile model) from a register to the write port, this will not introduce any non-determinism. However, it is possible to construct situations where it would. A potential resolution would be to add an assertion that asserts a lack of write-write collisions and error with "warning, your setup can produce non-determinism" if they exist.

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
davidbiancolin: The mask

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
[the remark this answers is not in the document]
davidbiancolin: Ah, right. That's subtle. You could also check that the write-token order never changes from cycle to cycle. since as you point out, it's probably going to be static in 99.9% of foreseeable cases. I think starting with a write collision check is reasonable.

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
albert-magyar: The target mask is supported.

albert-magyar: Keep in mind these are ground-type memories. The scheduler picks a write only when it is active (mask and en both true) and skips it otherwise, so the value of the target mask is implemented correctly; however, the mask of the internal host memory is always set to true. This makes both false-mask and false-enable ports "skippable," which is some trivial performance benefit.

davidbiancolin: Duh, ok. It might be worth just add a small comment noting that.

albert-magyar: I added a comment.

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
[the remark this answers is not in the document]
albert-magyar: > I think this is the PR @davidbiancolin mentioned would improve the FMR if I try to start using the MCRams optimization, please correct me if I've wandered into a different PR. @albert-magyar, if I want to try this out, I need to do something to make the `LabelSRAMMacros` pass use this model, right? I'm starting to look at that code but if you could give me a pointer on how to try this out, I'd really appreciate it!

Sure -- basically, all you need to do is change [this line](https://github.com/firesim/firesim/blob/9598b7f4448e693fe9f3b18689a55503a7ea08c8/sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala#L195) to use `BankedAsyncMemChiselModel` instead of `AsyncMemChiselModel`. They take the same parameters, aside from some optional ones for the banked version that have sensible defaults and control the configuration of the underlying host memory.

(Sorry for the late reply; perhaps you already found this out.)

albert-magyar:
> I think this is the PR @davidbiancolin mentioned would improve the FMR if I try to start using the MCRams optimization, please correct me if I've wandered into a different PR. @albert-magyar, if I want to try this out, I need to do something to make the `LabelSRAMMacros` pass use this model, right? I'm starting to look at that code but if you could give me a pointer on how to try this out, I'd really appreciate it!

Sure -- basically, all you need to do is change [this line](https://github.com/firesim/firesim/blob/9598b7f4448e693fe9f3b18689a55503a7ea08c8/sim/midas/src/main/scala/midas/passes/fame/EmitAndWrapRAMModels.scala#L195) to use `BankedAsyncMemChiselModel` instead of `AsyncMemChiselModel`. They take the same parameters, aside from some optional ones for the banked version that have sensible defaults and control the configuration of the underlying host memory.

(Sorry for the late reply; perhaps you already found this out.)

sim/midas/src/main/scala/midas/models/sram/BankedAsyncMemModel.scala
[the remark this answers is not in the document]
albert-magyar: Something else to consider: there is no guarantee that multi-write-port memories will be deterministic in the *target* under write collisions, since this is undefined under FIRRTL. This can affect both FPGA prototypes and unoptimized models that contain these memories. In practice, combinational-read memories have deterministic write collisions, but this is relying on implementation-defined behavior. However, anything that uses BRAMs to implement 2-RW synchronous-read memories (including the new "FPGA backend") will result in the non-deterministic write collision behavior of true dual-port BRAMs being exposed to unoptimized simulators of target designs that generate write collisions to such memories.

albert-magyar:
> And can you add requires for parameterization corners that are unsupported (if applicable).

I am not aware of any parameterizations that will not work at the moment. I've fuzz tested instances with a few numbers of target read and write ports from 1+1 to 12+6 and host ports from 1+1 to 4+2, so I don't see anything obvious, but nothing comprehensive.

davidbiancolin (approved):
Seems to me this is good to merge. Can you open an issue as a reminder to switch to this by default once we have bumped firrtl to bring in the FPGA backend.

PR #677, opened 2020-12-10
Midas: import scala.language.implicitConversions

ingallsj:
**Type of change**: paying off technical debt

**Impact**: no functional change

**What was changed**
fix compile warning:
```
[warn] /firesim/sim/midas/src/main/scala/midas/widgets/CppGeneration.scala:73:16: implicit conversion method toStrLit should be enabled
[warn] by making the implicit value scala.language.implicitConversions visible.
[warn] This can be achieved by adding the import clause 'import scala.language.implicitConversions'
[warn] or by setting the compiler option -language:implicitConversions.
[warn] See the Scaladoc for value scala.language.implicitConversions for a discussion
[warn] why the feature should be explicitly enabled.
[warn]   implicit def toStrLit(str: String): CStrLit = CStrLit(str)
[warn]                ^
```


timsnyder-siv (commented):
@davidbiancolin , we're trying to clean up Scala compiler warnings and this is one that was found.  

https://docs.scala-lang.org/tour/implicit-conversions.html (end of the page) has an explanation of what `import scala.language.implicitConversions` is doing.

davidbiancolin (approved):
Thanks @ingallsj for the PR. LGTM.

PR #678, opened 2020-12-10
fix compile warning: Chisel3 withReset no longer experimental

ingallsj:
**Type of change**: paying off technical debt

**Impact**: no functional change

**What was changed**
fix compile warning:
```
[warn] /firesim/sim/firesim-lib/src/main/scala/bridges/BlockDevBridge.scala:107:26: value withReset in package experimental is deprecated (since 3.2): Use the version in chisel3._
[warn]     chisel3.experimental.withReset(reset.toBool || targetReset) {
[warn]                          ^
```


davidbiancolin:
While i'm down to fix chisel3 deprecation warnings, given just how many there are across RC / CY / FireSim, it might be best to batch this up with more changes.

ingallsj:
> While i'm down to fix chisel3 deprecation warnings, given just how many there are across RC / CY / FireSim, it might be best to batch this up with more changes.

I'm just starting with the smaller and easier ones.  They will get bigger.  :)

davidbiancolin:
> > While i'm down to fix chisel3 deprecation warnings, given just how many there are across RC / CY / FireSim, it might be best to batch this up with more changes.
> 
> I'm just starting with the smaller and easier ones. They will get bigger. :)

Sounds good to me. :)

davidbiancolin (approved):
LGTM

PR #680, opened 2020-12-11
Fix a FASED Deadlock Bug under Wide MBus Configurations

davidbiancolin:
This is a bandaid solution for targets with wide memory buses. The code change definitely fixes a logical error in the code, although it's not entirely clear to me why it works.  It might just mask an underlying problem with how i'm doing width-adaption by putting less load on the memory system. A dedicated AXI4 width widget is forthcoming.

This PR also:
- cleans up the width handling by adding the nodes directly into the main diplomatic graph instead of instantiating them in a dedicated module. 
- lets the latency histograms support more than one outstanding req per ID


timsnyder-siv (approved):
I can confirm that these changes enable my design with `targetWidthBytes > hostWidthBytes` to boot Linux whereas before this bandaid, it would hang during Linux boot.

I think we should go ahead and merge this unless you want others to review.

PR #682, opened 2020-12-11
repo_state_summary.sh capture state from toplevel superproject down

timsnyder-siv:
@davidbiancolin this is a small change to your repo_state_summary.sh script so that it will capture all of the git state when FireSim is being used as a library.  This should be useful for chipyard as top as well as my usecase.

I verified that `git rev-parse --show-superproject-working-tree` exists in git 2.16.6 (the version that the manager launch script installed in /usr/bin/git when I setup my manager).


davidbiancolin (approved):
Nice change. Given that this is only every really run on the manager, LGTM.

What does happen if you run it on a system with an older version of git?

timsnyder-siv:
> What does happen if you run it on a system with an older version of git?

It will attempt to `cd --show-superproject-working-tree` and that will error with the usage of cd and cause the repo_state_summary.sh script to exit non-zero, causing `make replace-rtl` to fail.

We could tweak it a bit so that the error is more obvious by explicitly checking to make sure that `$(git rev-parse --show-superproject-working-tree) != "--show-superproject-working-tree"` and printing a better message about requiring newer git.

PR #683, opened 2020-12-11
Refactor tar2afi logic from aws_build into seperate function

### LINKED ISSUE #669: Interest in tar->AGFI as standalone manager command?
I have some builds I kicked off (with a hacked up manager) that finish Vivado fine and fail to copy back to the manager.

I'm breaking out [the logic that submits the tarball for AGFI creation](https://github.com/firesim/firesim/blob/659ddbf418d033f3b2fa3edb5bce46479dd0e3be/deploy/buildtools/buildafi.py#L141-L236) into a function that can continue to be called from `buildafi.aws_build` and also be called from a new manager 'task' function.  Just to reiterate, I'm not proposing any change in behavior for buildafi, only adding a way to do just the tar->AFI portion if something goes wrong during buildafi and you end up with a useful tarball via more manual methods.

One thing that is required for this to work is being able to pass in a cmdline argument to set launch_time  to an existing prefix of one of the results-build/* directories.

I'm not sure how useful this would be for others.  It could possibly be confusing since it's only useful when things go wrong in buildafi and you have to pick up the pieces.  Might be useful if the build fails and then you can manually fix it up from the checkpoints?

timsnyder-siv:
Enables one to run 'firesim tar2afi --launchtime <timestamp>' to run only
the tar->afi portion of buildafi for the configs listed in config_build.ini

Behavior of buildafi is unchanged

fixes #669


deploy/buildtools/buildafi.py
```diff
@@ -171,7 +189,7 @@ def on_build_failure():
 
     # if we're unlucky, multiple vivado builds may launch at the same time. so we
     # append the build node IP + a random string to diff them in s3
-    global_append = "-" + env.host_string + "-" + ''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(10)) + ".tar"
+    global_append = "-" + str(env.host_string) + "-" + ''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(10)) + ".tar"
```
timsnyder-siv, line 192: When using tar2afi, Fabric's env.host_string will be None.  explicitly casting it to str makes this avoid an exception and it should be fine for the host element to be empty for tar2afi.  To be more clear, we could check for None and use 'localhost' if that's preferable.

deploy/firesim
```diff
@@ -149,6 +149,22 @@ def buildafi(globalbuildconf):
...
+    Iterate the globalbuildconf.get_builds_list() using a multiprocessing.Pool
+    running the logic for submitting the tarball to `aws ec2 create-fpga-image`,
+    wait for afi to be available and copy to all known regions
+
+    NOTE: this is only useful if you also pass the --launchtime argument to firesim
+    and specify an already existing launchtime for a previous, tarball generation
+    """
+
+    for buildconf in globalbuildconf.get_builds_list():
+        execute(aws_create_afi, globalbuildconf, buildconf, hosts=['localhost'])
```
timsnyder-siv, line 165: This could be improved.  As it is now, all of the configs will execute serially.  I don't see how to get Fabric to execute them in parallel on localhost.  I don't really want to have to spin up a buildfarm only for the purposes of then running things in parallel on localhost.  I thought about using multiprocessing or threading to do this in parallel but it seems ugly to import another subprocess execution library.

I think this is useful even with them running in series.

Thoughts?

deploy/firesim
```diff
@@ -240,6 +257,8 @@ def construct_firesim_argparser():
                         default=-1)
     parser.add_argument('-q', '--forceterminate', action='store_true',
                         help='For terminaterunfarm, force termination without prompting user for confirmation. Defaults to False')
+    parser.add_argument('-t', '--launchtime', type=str,
+                        help='Give the "Y-m-d--H-M-S" prefix of results-build directory. Useful for tar2afi when finishing a partial buildafi')
```
timsnyder-siv, line 261: I currently only added support for `--launchtime` to the GlobalBuildConfig.  It will also be useful for the RuntimeConfig if we refactor the monitor loop in `runworkload` into a separate method that can be restarted if/when the monitor dies.

davidbiancolin (approved):
This LGTM, but it'd be good to get @sagark's eyes on it.

deploy/firesim
davidbiancolin: I could've sworn i ran stuff in parallel on the localhost in the past. Maybe you can work around this by specifying repeated instances of the localhost in the worker pool?

PR #684, opened 2020-12-11
add try/catch blocks to avoid VCS DPI-UED

timsnyder-siv:
If an uncaught exception percolates up into VCS, it gives you a very
unhelpful message:
Error-[DPI-UED] C++ Exception detected
  Import DPI routine invoked at file
  '/scratch/tims/federation_readd_fsim_flow/firesim/sim/midas/src/main/verilog/vcs_top.v'(line
  1020) has C++ exceptions not caught. C++ exceptions shall not propagate out
  of any imported subroutine.
  Fix it in DPI-C code before running simulation.

To avoid that, we must add try-catch blocks around the toplevel of each
thread of the FireSim driver.

This is a first cut of doing that but needs a corresponding CY change to
the fesvr now that we're not using a separate FireSim fesvr implementation.


timsnyder-siv:
This may also need try/catch block added into CY fesvr because I was looking at a version before #641 was merged in.

davidbiancolin (approved):
LGTM

PR #690, opened 2020-12-21
provide vcs -top to avoid spurious toplevel in hierarchy (#618)

### LINKED ISSUE #618: F1Shim still appears as top level module in waveform
@davidbiancolin 
Current firesim version is still simulating F1Shim as a top level module in addition to emul/FPGATop.

f1shim

timsnyder-siv:
From pg 4-69 of https://spdocs.synopsys.com/dow_retrieve/R-2020.09/vg/VCS/PDFs/vcs.pdf

VCS has the -top compile-time option for specifying the configuration that describes the
top-level configuration or module of the design. For example:vcs -top top_cfg ...If you have
coded your design to have more than one top-level module, you can enter more than one -top
option, or you can append arguments to the option using the plus delimiter. For example:-top
top_cfg+test+Using the -top option tells VCS not to create extraneous top-level modules, that
is, one that you do not specify.

fixes #618

### DIFF
--- a/sim/midas/src/main/cc/rtlsim/Makefrag-vcs
+++ b/sim/midas/src/main/cc/rtlsim/Makefrag-vcs
@@ -18,6 +18,7 @@
 VCS ?= vcs -full64
 override VCS_FLAGS := -quiet -timescale=1ns/1ps +v2k +rad +vcs+initreg+random +vcs+lic+wait \
 	-notice -line +lint=all,noVCDE,noONGS,noUI -quiet -debug_pp +no_notifier -cpp $(CXX) \
+	-top emul \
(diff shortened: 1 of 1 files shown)


davidbiancolin (approved):
Good stuff. I'll handle verilator in a separate PR. (I personally only ever use waves with VCS)

timsnyder-siv:
> Good stuff. I'll handle verilator in a separate PR. (I personally only ever use waves with VCS)

I use Verilator to gen waves when I'm on AWS and too lazy to migrate back to a host with VCS installed.  I'm not sure the problem exists in Verilator because we already provide `--top-module` in the verilator cmdline.

PR #694, opened 2021-01-06
Prevent repeated AbstractClockGate definitions

albert-magyar:
This makes the pass defining the clock gate model idempotent.


davidbiancolin (approved):
lgtm

PR #689, opened 2020-12-18
bump chipyard for rc chisel firrtl 3.4.1.x

timsnyder-siv:
Haven't found any compile-time changes needed inside firesim for bumping RC in ucb-bar/chipyard#742.  The CI looks like it is running alright locally.  Opening a pr with the chipyard bump so that the CI will run cleanly in the cloud as well.

PR #699, opened 2021-01-14
Fix Two GG-Compile Time Errors

davidbiancolin:
Fixes two bugs that break default target elaboration on dev, this should enable the RC bump to proceed forward.

- Pulls in a update to DedupModules that fixes a loss of connectivity when using Chisel bulk connects 

- As a workaround, removes now-unneeded `DontTouchAnnotations` emitted by the AutoCounter cover implementation. These were being emitted into modules that were later inlined. I suspect they should be renamed to point to nodes in the parent but, for still unknown reasons, weren't. This wasn't a problem previously because, until 5c890ead6, we were never running `InlineInstances`.  

It might be worthwhile to regenerate AGFIs for this change. Alternatively it can be folded into the RC bump.


davidbiancolin:
h/t @jackkoenig for saving me on the dedup fix.

timsnyder-siv (approved):
Using this PR, I'm able to get past the issue with the DontTouchNotFoundException I mentioned in #689 this morning.

PR #705, opened 2021-01-24
Add a Rocket-Based Test to CI

davidbiancolin:
We have more memory available to us in the JVM now that we aren't running two instances of SBT.


davidbiancolin:
This is subsumed by #727.

PR #706, opened 2021-01-24
Support Multibit Events in AutoCounter

davidbiancolin:
Previously you could not accumulate non-boolean events using AutoCounter, which forced users to instantiate multiple counters and do the reduction out-of-band (e.g., each lane of SS machine would have its own event for instructions retired). If you already have the reduction handy it's convenient to annotate just the one multibit event.


sim/midas/targetutils/src/main/scala/midas/annotations.scala
```diff
@@ -202,10 +202,15 @@ case class AutoCounterCoverModuleAnnotation(target: String) extends ChiselAnnota
...
 object PerfCounter {
   /**
-    * Annotates a Bool representing a target event (ex. L1 D$ miss)  that
-    * should be tracked by AutoCounter
+    * Labels a signal as an event for which an host-side counter (an
+    * "AutoCounter") should be generated).  Events can be multi-bit to encode
+    * multiple occurances in a cycle (e.g., the number of instructions retired
+    * in a superscalar processor). NB: Golden Gate will not generate the
+    * coutner unless AutoCounter is enabled in your the platform config. See
+    * the docs for more info.
```
timsnyder-siv, line 210: Should this be a ref to a specific point in the docs or is that too much to maintain?

timsnyder-siv:
Just reading through the code, this seems to make sense to me and I don't see anything that jumps out as wrong.  However, I'd really like to try it or see a test added before approving.

alonamid:
It took me a while to understand the usage model here while reading the code. I think I understand it now, but let me try to repeat it to see if I understand correctly. If that's indeed the case, I would echo @timsnyder 's comment about adding a more detailed description of the semantics of the PerfCounter annotation arguments to the docs, because I think this feature adds some ambiguity to the "target" argument to the PerfCounter annotation (and perhaps even changes it's semantics):

In the "traditional" case of counting a boolean signal, by providing the boolean signal as an argument to the PerfCounter annotation, AutoCounter would count how many cycles that signal was high. Hence, in this "legacy" case, the semantic of the PerfCounter annotation was that it would count the number of cycles the target signal was high.

In the case described by this PR, the semantic of the the target argument to the PerfCounter annotation is not longer "a signal" to be counted, but rather an *increment value* for a counter to increment by. In the case of the "legacy" boolean signal, these *happen* to be the same, since because the counter needs to increment by 1 every cycle that the target signal is 1. In the case of signal bitvectors such as "retired instructions", we expect the user to perform a reduction in the source code, and provide only the increment amount as the target argument in PerfCounter. 
While I agree this is a reasonable usage model (since user may choose to compute the increment using different bitvector reductions such as XOR/AND/OR and we don't want to strictly enforce any of those), and do this we should be clear about this semantic, and how is just happen to overlap with the "trivial" boolean usage model.

timsnyder-siv:
Maybe I should elaborate more than +1.  @alonamid, I think you understand the new semantics and I agree with your feedback that it would be good to document the semantics.

sim/src/main/scala/midasexamples/AutoCounterModule.scala
```diff
@@ -27,6 +27,11 @@ class AutoCounterModuleDUT(
 
   PerfCounter(enabled4, "ENABLED_DIV_4", "Count the number of times the enabled cycle count is divisible by 4. Should be equal to number of cycles minus reset cycles divided by 4")
 
+  // Multibit event
```
davidbiancolin, line 30: FWIW, this is the test.

timsnyder-siv: Oh.  Cool.   I wasn't sure if it was or not because I didn't see something checking the counted value against some expectation.  I need to learn more about how the tests work. :)

davidbiancolin:
I'll expand the docs to explain what happens in the multibit case a little better.

sim/src/main/scala/midasexamples/AutoCounterModule.scala
davidbiancolin: How these particular tests work is by diffing the AutoCounter-generated output against an printf that's been formatted to match that output (see below). It's a little bit hacky, but at least it's simple.

timsnyder-siv: I was sort of seeing it before but didn't see anything that was parsing the output of the print statements and making sure they were the same, so I (very incorrectly) assumed that this wasn't one of the CI tests but was something that could be visually inspected for correctness.  

In other words, I was forgetting the leap back into 
https://github.com/firesim/firesim/blob/fec239847ac57c7b13b1782ba933b9fc91de9121/sim/src/test/scala/midasexamples/TutorialSuite.scala#L110-L123  
where `diffSynthesizedLog` is being called.  I was also being thrown off by the 'test' code living in 'src/main' instead of 'src/test'.

Not trying to complain about the tests being hacky.  It's great that they are there.  I just forget about how they are plumbed together because it goes through a few layers of SBT, makefile, SBT and I get confused.

davidbiancolin: One way to rationalize how this works, is that these duts are targets with their own generators. The tests merely consist of a means to invoke parts of the make-based flow and diff generated outputs. It started off this way because the midasexamples were suppose to a simple way to demonstrate how you could use MIDAS not a target that wasn't rocket-chip based. I've since added a bunch more targets that both demonstrate simple features but increasingly just act as tests. It probably makes sense to pull them into src/test, and with a little SBT chicanery everything will work the same way. 




The tests are wrappers consist of wrappers around existing

timsnyder-siv: While it might be possible to reorganize things to put these into 'src/test', I think that a less drastic way to aleviate what I'm describing would be adding Scaladoc for `AutoCounterModule`, `AutoCounterCoverModule` and `AutoCounterPrintfModule` that captures what you just wrote: 'this is a  target its their own generator that demonstrates XXX and is suitable for use in testing by extending the abstract `TutorialSuite`'.

I was going to suggest a single file-scope Scaladoc but looking more closely, that doesn't seem to be a thing in Scaladoc as it is in C++ Doxygen.  I'll see if I can propose them as comments on this PR in the appropriate places.

timsnyder-siv: Grr.  Of course I can't.  The lines I want to propose patches for are too far away from your original patches so why the heck would I ever want to comment on them!  Thanks Github!!!

I'll submit my suggestions as another PR into this PR (rather than doing nothing or pushing them straight into your branch).

davidbiancolin:
@alonamid I updated the docs. LMK if you'd like specific changes.

alonamid (approved):

Issue #695, opened 2021-01-11
Default FPGADebug Annotations & ILA Platform Config Switch


davidbiancolin:
Two things:
- It would be nice to annotate some critical signals as FPGADebug by default (check them in). The or-reduction in the assert bridge being a good example. 
- To enable ^ without wasting resources, it would probably be good to put auto ILA on its own PLATFORM_CONFIG switch, and disabled by default. Of our debugging utilities, AutoILA is the only one that does not do this, so being consistent would be good.

timsnyder-siv:
Would it be reasonable to get some of this into 1.11 if we restrict the scope to things that are in firesim and chipyard and I do a first cut PR?  Getting critical signals in rocket and boom (maybe other generators) might take more time.  I was thinking of annotating:
* `assertFire`, `assertId`, `q.io.deq.valid`, and `cycles` from [Assert.scala](https://github.com/firesim/firesim/blob/dev/sim/midas/src/main/scala/midas/widgets/Assert.scala#L25-L53)
* DRAM interface.  Would `EgressReq` and `EgressResp` in [dram/Interfaces.scala](https://github.com/firesim/firesim/blob/dev/sim/midas/src/main/scala/midas/models/dram/Interfaces.scala#L44-L57) be the right place to annotate the DRAM interface so that one could at least see memory transactions coming to/from the DUT?
* Something that fleshes out what @davidbiancolin means by "using an AutoILA" in  [Debugging-Hanging-Simulators.rst](https://github.com/firesim/firesim/blame/f1d8f5fc5ade787ce57c265301d1de00654921c3/docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/Debugging-Hanging-Simulators.rst#L45)  Depending on how much would need to be annotated, maybe this would be too much to have always going into the ILA.  If so, we could explain more in the documentation about what you should annotate when debugging a hang.
* what else in FireSim would be generally useful for debugging purposes?

davidbiancolin:
Is there any need to push for release inclusion? 

> Would it be reasonable to get some of this into 1.11 if we restrict the scope to things that are in firesim and chipyard and I do a first cut PR? Getting critical signals in rocket and boom (maybe other generators) might take more time. I was thinking of annotating:
> 
[...]

PR #712, opened 2021-02-02
 #695 must mixin WithAutoILA to platform config to populate ILA

timsnyder-siv:
Addresses part of #695 by adding a new `WithAutoILA` PLATFORM_CONFIG mixin that is required to enable the AutoILA population.  This will enable us to keep interesting `FpgaDebug()` annotations in our design or to add default annotations in FireSim IP.


timsnyder-siv:
To provide a migration path, we agreed that all of the recipes in https://github.com/firesim/firesim/blob/dev/deploy/sample-backup-configs/sample_config_build_recipes.ini should mix `WithAutoILA` into their PLATFORM_CONFIG to maintain the current behavior for the default builds.  I'll push that as another commit to this PR.

timsnyder-siv:
I rebased this forward to resolve the conflict in sample_config_build_recipes.ini

alonamid (approved):
LGTM

PR #716, opened 2021-02-04
Enhance launchrunfarm to patiently try harder

### LINKED ISSUE #713: Enhance launchrunfarm to patiently try harder
We have a workload that contains 60 jobs.  Haven't ever had a problem getting hosts but yesterday and today, we have started hitting this case:
https://github.com/firesim/firesim/blob/9ffba8dbf4ec22a17269ecd15f0a6ad761c8e82b/deploy/awstools/awstools.py#L174-L178

It's annoying because:
* we've already allocated a goodly portion of what we need
* `run_workload.sh` very politely does `terminaterunfarm` after `launchrunfarm` bombs out
* then we start trying to allocate again.

Seems to me like when AWS is having trouble meeting demand for f1 hosts that `launchrunfarm` should:
* look for existing hosts marked with `fsimcluster` and only try to allocate the remaining (so that it can be rerun after a failure)
* not give up completely after failing to allocate a host across all subnets.  I've locally hacked it so that it just tries forever right now.  It should probably be a config_runtime.ini parameter for how long we try to allocate hosts before giving up.  I say 'patiently' in the title because I do think we should sleep for some amount of time when we've tried all of the subnets to give AWS a chance to rebalance it's blend of f1 instance types in a given region before we start hammering on the create-instance API again.

@davidbiancolin , how often do you guys run into this type of thing?  Do my changes to `launchrunfarm` seem sensible?  If so, I'm happy to open a PR.

timsnyder-siv:
fixes #713

- [x] expose  `always_expand_runfarm` to runtime.ini or remove it.

### DIFF
--- a/deploy/awstools/aws_setup.py
+++ b/deploy/awstools/aws_setup.py
@@ -1,0 +1,64 @@
+#/usr/bin/env python3
+
+""" This script configures your AWS account to run FireSim. """
+
+import boto3
+
+vpcname = 'firesim'
+secgroupname = 'firesim'
+
+def aws_setup():
+    ec2 = boto3.resource('ec2')
+    client = boto3.client('ec2')
+
+    # get list of avail zones in the region, we will need it later
+    avail_zones = list(map(lambda x: x['ZoneName'], client.describe_availability_zones()['AvailabilityZones']))
+    av_zones_with_3octet = zip(range(len(avail_zones)), avail_zones)
+
+    print("Creating VPC for FireSim...")
+    vpc = ec2.create_vpc(CidrBlock='192.168.0.0/16')
+    vpc_id = vpc.id
+    # confirm that vpc is actually available before running commands
+    client.get_waiter('vpc_exists').wait(VpcIds=[vpc_id])
+    client.get_waiter('vpc_available').wait(VpcIds=[vpc_id])
+
+    vpc.create_tags(Tags=[{"Key": "Name", "Value": vpcname}])
+    vpc.wait_until_available()
+
+    ig = ec2.create_internet_gateway()
+    vpc.attach_internet_gateway(InternetGatewayId=ig.id)
+
+    route_table = vpc.create_route_table()
+    route = route_table.create_route(
+        DestinationCidrBlock='0.0.0.0/0',
+        GatewayId=ig.id
+    )
+    print("Success!")
+
+    print("Creating a subnet in the VPC for each availability zone...")
+    subnets = []
+    # create a subnet in each availability zone for this vpc
+    for ip, zone in av_zones_with_3octet:
+        subnets.append(ec2.create_subnet(CidrBlock='192.168.' + str(ip) + '.0/24', VpcId=vpc.id, AvailabilityZone=zone))
+        client.get_waiter('subnet_available').wait(SubnetIds=[subnets[-1].id])
+        client.modify_subnet_attribute(MapPublicIpOnLaunch={'Value': True}, SubnetId=subnets[-1].id)
+        route_table.associate_with_subnet(SubnetId=subnets[-1].id)
+    print("Success!")
+
+    print("Creating a security group for FireSim...")
+    sec_group = ec2.create_security_group(
+        GroupName=secgroupname, Description='firesim security group', VpcId=vpc.id)
+
+    # allow all egress rule exists by default
+
+    # ingress rules
+    sec_group.authorize_ingress(IpPermissions=[
+        {u'PrefixListIds': [], u'FromPort': 60000, u'IpRanges': [{u'Description': 'mosh', u'CidrIp': '0.0.0.0/0'}], u'ToPort': 61000, u'IpProtocol': 'udp', u'UserIdGroupPairs': [], u'Ipv6Ranges': [{u'Description': 'mosh', u'CidrIpv6': '::/0'}]},
+        {u'PrefixListIds': [], u'FromPort': 22, u'IpRanges': [{u'CidrIp': '0.0.0.0/0'}], u'ToPort': 22, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
+        {u'PrefixListIds': [], u'FromPort': 10000, u'IpRanges': [{u'Description': 'firesim network model', u'CidrIp': '0.0.0.0/0'}], u'ToPort': 11000, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': [{u'Description': 'firesim network model', u'CidrIpv6': '::/0'}]},
+        {u'PrefixListIds': [], u'FromPort': 3389, u'IpRanges': [{u'Description': 'remote desktop', u'CidrIp': '0.0.0.0/0'}], u'ToPort': 3389, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': [{u'CidrIpv6': '::/0', u'Description': 'rdp'}]}])
+
+    print("Success!")
+
+if __name__ == '__main__':
+    aws_setup()
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -159,7 +159,7 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior, 
         assert False, "INVALID INSTANCE MARKET TYPE."
 
 def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavior, spotmaxprice, blockdevices=None,
-                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), additive=True):
+                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), expand_runfarm_by_count=True):
     """ Launch `count` instances of type `instancetype`
 
     Using `instancemarket`, `spotinerruptionbehavior` and `spotmaxprice` to define instance market conditions
@@ -186,10 +186,10 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
         Path to readable file.  Contents of file are passed as `UserData` to AWS
     timeout : datetime.timedelta, default=timedelta() (immediate timeout after attempting all subnets)
         `timedelta` object representing how long we should continue to try asking for instances
-    additive : bool, default=True
+    expand_runfarm_by_count : bool, default=True
         When true, create `count` instances, regardless of whether any already exist. When False, only
         create instances until there are `count` total instances that match `tags` and `instancetype`
-        If `tags` are not passed, `additive` must be `True` or `ValueError` is thrown.
+        If `tags` are not passed, `expand_runfarm_by_count` must be `True` or `ValueError` is thrown.
 
     Return type
     -----------
@@ -197,12 +197,12 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
 
     Returns
     -------
-    list of instance resources.  If `additive` is True, this list contains only the instances created in this
-    call.  When `additive` is False, it contains all instances matching `tags` whether created in this call or not
+    list of instance resources.  If `expand_runfarm_by_count` is True, this list contains only the instances created in this
+    call.  When `expand_runfarm_by_count` is False, it contains all instances matching `tags` whether created in this call or not
     """
 
-    if tags is None and not additive:
-        raise ValueError("additive=False requires tags to be given")
+    if tags is None and not expand_runfarm_by_count:
+        raise ValueError("expand_runfarm_by_count=False requires tags to be given")
 
 
     aws_resource_names_dict = aws_resource_names()
@@ -230,7 +230,7 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
     # starting with the first subnet, keep trying until you get the instances you need
     startsubnet = 0
 
-    if tags and not additive:
+    if tags and not expand_runfarm_by_count:
         instances = instances_sorted_by_avail_ip(get_instances_by_tag_type(tags, instancetype))
     else:
         instances = []
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -159,12 +159,51 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior, 
         assert False, "INVALID INSTANCE MARKET TYPE."
 
 def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavior, spotmaxprice, blockdevices=None,
-                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(minutes=0)):
-    """ Launch count instances of type instancetype, optionally with additional
-    block devices mappings and instance tags
+                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), additive=True):
+    """ Launch `count` instances of type `instancetype`
 
-         This will launch instances in avail zone 0, then once capacity runs out, zone 1, then zone 2, etc.
+    Using `instancemarket`, `spotinerruptionbehavior` and `spotmaxprice` to define instance market conditions
+    (see also: construct_market_conditions)
+
+    This will launch instances in avail zone 0, then once capacity runs out, zone 1, then zone 2, etc.
+    The ordering of availablility zones can be randomized by passing`randomsubnet=True`
+
+    Parameters
+    ----------
+    instancetype : str
+        String acceptable by `boto3.ec2.create_instances()` `InstanceType` parameter
+    count : int
+        The number of instances to launch
+    instancemarket
+    spotinterruptionbehavior
+    spotmaxprice
+    blockdevices
+    tags : dict of tag names to string values, default=None
+        Dict of tags
+    randomsubnet : bool, default=False
+        If true, subnets will be chosen randomly instead of starting from 0 and proceeding incrementally.
+    user_data_file : str, default=None
+        Path to readable file.  Contents of file are passed as `UserData` to AWS
+    timeout : datetime.timedelta, default=timedelta() (immediate timeout after attempting all subnets)
+        `timedelta` object representing how long we should continue to try asking for instances
+    additive : bool, default=True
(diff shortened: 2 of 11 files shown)


deploy/awstools/awstools.py
```diff
@@ -169,14 +174,20 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
 
     # starting with the first subnet, keep trying until you get the instances you need
     startsubnet = 0
-    instances = []
-    while len(instances) < count:
-        if not (startsubnet < len(subnets)):
-            rootLogger.critical("we tried all subnets, but there was insufficient capacity to launch your instances")
-            rootLogger.critical("""only the following {} instances were launched""".format(len(instances)))
-            rootLogger.critical(instances)
-            return
```
timsnyder-siv, line 178: I moved this down into the `except` block (and added the timeout logic) so that I could re-`raise` the exception, but then I didn't like how the output read.  I thought it was confusing to have the stacktrace at the end and switched to just using `sys.exit(1)`.

deploy/awstools/awstools.py
```diff
@@ -169,14 +174,20 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
...
 
+    if 'fsimcluster' in tags.keys():
+        instances = instances_sorted_by_avail_ip(get_instances_by_tag_type(tags['fsimcluster'], instancetype))
```
timsnyder-siv, line 179: This is slightly wonky.  There is nothing in the signature of `launch_instances` that requires `fsimcluster` to be a key in `tags` but there may be other tags that should be checked.  Do build hosts use this same `launch_instances` helper or do they have a different one?

deploy/awstools/awstools.py
```diff
@@ -207,12 +218,31 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
 
         except client.exceptions.ClientError as e:
             rootLogger.info(e)
-            rootLogger.info("This probably means there was no more capacity in this availability zone. Try the next one.")
```
timsnyder-siv, line 210: I always thought the imperative 'Try the next one' was confusing in this message, like it was telling me as the user to do something instead of informing me that the script is going to try the next one.

deploy/runtools/runtime_config.py
```diff
@@ -264,6 +265,11 @@ def __init__(self, runtimeconfigfile, configoverridedata):
         self.m4_16xlarges_requested = int(runtime_dict['runfarm']['m4_16xlarges']) if 'm4_16xlarges' in runtime_dict['runfarm'] else 0
         self.f1_2xlarges_requested = int(runtime_dict['runfarm']['f1_2xlarges']) if 'f1_2xlarges' in runtime_dict['runfarm'] else 0
 
+        if 'launch_instances_timeout_minutes' in runtime_dict['runfarm']:
+            self.launch_timeout = timedelta(minutes=int(runtime_dict['runfarm']['launch_instances_timeout_minutes']))
+        else:
+            self.launch_timeout = timedelta(minutes=60)
```
timsnyder-siv, line 271: I default to a one-hour timeout if it isn't specified.  I suppose, some people might want the old behavior (if they care about making sure that they have minimized the number of availability zones.  Let me know if you disagree with this default.  I'm happy to change it to 0 so that the old behavior is the default.

deploy/awstools/awstools.py
timsnyder-siv: Also, if we find existing instances in the `fsimcluster`, do we want to check them to make sure they aren't already running `FireSim-f1`?  Might be a good feature but I stopped short of throwing it in with this PR.

deploy/runtools/runtime_config.py
sagark: I would prefer to keep the original behavior if there's no launch_instances_timeout_minutes key, just so that there are no surprises for users relying on it to error if there aren't enough instances. 

IIRC some of the isca workloads that require many instances expect this behavior and we probably wouldn't want it to keep retrying for an hour while leaving the most of the instances running if someone starts the script and walks away.

For the explicit default later, I think 60 mins is fine.

deploy/awstools/awstools.py
sagark: I agree, good catch, thanks!

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -51,6 +51,23 @@ Note that these values are ONLY used to launch instances. After launch, the
...
+
+Integer number of minutes that the ``launchrunfarm`` command will attempt to
+request new instances before giving up.  This limit is used for each of the types
+of instances being requested.  For example, if you leave it set to the default of 60,
+and you are requesting all four types of instances, ``launchrunfarm`` will try
+to launch each instance type for 60 minutes, possibly trying up to a total of
+four hours.
+
+This limit starts to be applied from the first time ``launchrunfarm`` has received a
+``ClientError`` response in all AvailabilityZone's for your region.  In other words,
```
sagark, line 65: AvailabilityZone's -> AvailabilityZones

deploy/awstools/awstools.py
sagark: Yeah, builds use this same helper, see: https://github.com/firesim/firesim/blob/master/deploy/buildtools/buildconfig.py#L38

sagark: One concern here is if there are workloads that rely on the current "additive" nature of launchrunfarm, i.e. they call it N times asking for one instance each time and expect to end up with N instances after all those calls.

sagark:
Thanks for this @timsnyder-siv ! This looks good to me overall, left a few comments. The only two major things I have are:

* I think not having the key for retrying should default to not retrying, for backwards compat.
* Is the "check for existing instances with the fsimcluster tag and only add more if necessary" feature required for your use case? This is another change that might be surprising to users expecting the current behavior. Perhaps this could be another option in the config ini?

timsnyder-siv:
> Thanks for this @timsnyder-siv ! 

Thanks for the feedback @sagark.  I'l make those changes and push when I'm able to.  Given that my power and internet are sporadically down, it may be a few days but I'll push updates for your feedback.

davidbiancolin:
@timsnyder-siv You think this can make 1.13?

timsnyder-siv:
> @timsnyder-siv You think this can make 1.13?

Yes. Will prioritize.

deploy/awstools/awstools.py
timsnyder-siv: Ok.  So, looking at this again, the wonkiness I was referring to is that `fsimcluster` is hardcoded in the logic here. It is also hardcoded in `get_instances_by_tag_type()`.  `get_instances_by_tag_type()` should be generalized to take a dict of tags that gets mapped into a filter in a more generic way instead of having `fsimcluster` hardcoded in either place.

I also need to add another control to address what Sagar calls 'the current "additive" nature of launchrunfarm'.  I think it would be helpful if there was a documented example of a workload that uses that feature because I have trouble understanding it.  I know that's a limitation on my end.

davidbiancolin: @abejgonzalez Do your changes to support other host environments also support growing the runfarm iteratively like this?

abejgonzalez: Oh. I never thought of a run farm being iteratively added to. Frankly, I haven't checked but I presume it should still work for EC2.

davidbiancolin: Yes, but it'd need to be supported by the allocators for other hosts. If we think it's a pain in the ass, then we should put this behavior on a deprecation schedule.

abejgonzalez: Do we know of any workloads that use this additive functionality? If SiFive and Berkeley don't use it then I think its probably safe to deprecate? IMO this functionality isn't clear from the `.ini`s and is tricky to use (to terminate you would have to call terminaterunfarm multiple times to terminate a subset of the instances? to terminate all of them?).

timsnyder-siv: > to terminate you would have to call terminaterunfarm multiple times to terminate a subset of the instances? to terminate all of them?

No. The docs are pretty clear that the instance count from the runtime.ini is only used when launching instances.  All other commands use the runfarmtag. https://github.com/firesim/firesim/blame/master/docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst#L50-L52

Regardless, I agree that this should be deprecated somehow to simplify the manager.

timsnyder-siv:
@davidbiancolin my tests for this will require https://github.com/spulec/moto/pull/4853 and that will be included in an upcoming release of moto.  The newer version of moto also requires newer python (because of code in it's dependent responses library).

I'm going to test things locally and make these new tests smart enough to 'skip' if the version of moto is broken but we'll want to bump Python and moto in a separate PR.

deploy/awstools/awstools.py
timsnyder-siv: In the most recent changes that I just pushed, I added an `additive` parameter to `launch_instances` that defaults to `True` and has the old behavior.  If we decide that we don't need, I can take it out and remove the tests for it.

deploy/runtools/runtime_config.py
timsnyder-siv: When the key isn't included in runtime.ini, the timeout defaults to 0 now.

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
timsnyder-siv:
```suggestion
``ClientError`` response in all AvailabilityZones for your region.  In other words,
```

deploy/awstools/aws_setup.py
```diff
@@ -0,0 +1,64 @@
+#/usr/bin/env python3
+
+""" This script configures your AWS account to run FireSim. """
```
timsnyder-siv, line 3: This used to live at `scripts/aws-setup.py`.  I moved it here because it's easier to import and use in tests for code that assume this has been run.  `scripts/aws-setup.py` still exists and just imports this and runs it.

deploy/awstools/awstools.py
```diff
@@ -20,6 +24,7 @@
 # this needs to be updated whenever the FPGA Dev AMI changes
 # You can find this by going to the AMI tab under EC2 and searching for public images:
 # https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#Images:visibility=public-images;search=FPGA%20Developer;sort=name
+# And whenever this changes, you also need to update deploy/tests/test_amis_snake.json
```
timsnyder-siv, line 27: It might be possible to change `tests/conftest.py` to automagically generate the json so that it will have the correct ami name string and doesn't have to be manually updated.  I stopped short of doing that in this PR.

deploy/awstools/awstools.py
```diff
@@ -220,17 +271,39 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
...
             }
-            if user_data_file is not None:
+            if user_data_file:
                 with open(user_data_file, "r") as f:
                     instance_args["UserData"] = ''.join(f.readlines())
 
             instance = ec2.create_instances(**instance_args)
             instances += instance
 
         except client.exceptions.ClientError as e:
```
timsnyder-siv, line 281: We might want to explicitly check for `e.value.response["Error"]["Code"] == "InsufficientInstanceCapacity"` to make it less likely that someone would have a non-capacity `ClientError` and sit around waiting.

deploy/awstools/awstools.py
```diff
@@ -153,13 +158,53 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior,
...
+    blockdevices
+    tags : dict of tag names to string values, default=None
+        Dict of tags
+    randomsubnet : bool, default=False
+        If true, subnets will be chosen randomly instead of starting from 0 and proceeding incrementally.
+    user_data_file : str, default=None
+        Path to readable file.  Contents of file are passed as `UserData` to AWS
+    timeout : datetime.timedelta, default=timedelta() (immediate timeout after attempting all subnets)
+        `timedelta` object representing how long we should continue to try asking for instances
+    additive : bool, default=True
```
davidbiancolin, line 189: This is more of a nit, but perhaps a name like "expand_runfarm_limit", or something that gets at the fact that the goal is what is being added to, not the existing count, would be good.

deploy/awstools/awstools.py
```diff
@@ -184,14 +229,20 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
...
-            rootLogger.critical(instances)
-            return
 
+    if tags and not additive:
+        instances = instances_sorted_by_avail_ip(get_instances_by_tag_type(tags, instancetype))
+    else:
+        instances = []
+
+    if len(instances):
+        rootLogger.info("Already have {} of {} {} instances.".format(len(instances), count, instancetype))
```
davidbiancolin, line 239: It seems to me this would underreport the number of instances in the additive mode.

davidbiancolin (commented):
This looks reasonable, thanks Tim. 

I'm still of the opinion we should remove the additive mode though. Lets get Abe to sign off on this change too.

deploy/awstools/awstools.py
```diff
@@ -153,13 +158,53 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior,
...
+    randomsubnet : bool, default=False
+        If true, subnets will be chosen randomly instead of starting from 0 and proceeding incrementally.
+    user_data_file : str, default=None
+        Path to readable file.  Contents of file are passed as `UserData` to AWS
+    timeout : datetime.timedelta, default=timedelta() (immediate timeout after attempting all subnets)
+        `timedelta` object representing how long we should continue to try asking for instances
+    additive : bool, default=True
+        When true, create `count` instances, regardless of whether any already exist. When False, only
+        create instances until there are `count` total instances that match `tags` and `instancetype`
+        If `tags` are not passed, `additive` must be `True` or `UserError` is thrown.
```
abejgonzalez, line 192: Nit: Shouldn't this be `ValueError`?

deploy/awstools/awstools.py
```diff
@@ -220,17 +271,39 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
                 ]),
                 "InstanceMarketOptions":marketconfig,
             }
-            if user_data_file is not None:
+            if user_data_file:
```
abejgonzalez, line 274: I'm not a Python pro by any means, but wouldn't this error out in the `open`?

deploy/awstools/awstools.py
abejgonzalez, line 192: Nit: Shouldn't this be `ValueError`?

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -51,6 +51,25 @@ Note that these values are ONLY used to launch instances. After launch, the
...
+
+Integer number of minutes that the ``launchrunfarm`` command will attempt to
+request new instances before giving up.  This limit is used for each of the types
+of instances being requested.  For example, if you set to 60,
+and you are requesting all four types of instances, ``launchrunfarm`` will try
+to launch each instance type for 60 minutes, possibly trying up to a total of
+four hours.
+
+This limit starts to be applied from the first time ``launchrunfarm`` receives a
+``ClientError`` response in all AvailabilityZones for your region.  In other words,
```
abejgonzalez, line 65: Nit: Can you say AvailabilityZones (AZs)

abejgonzalez (approved):
SGTM

deploy/awstools/awstools.py
timsnyder-siv: In the additive mode, it won't print because instances will be empty at this point.

deploy/awstools/awstools.py
timsnyder-siv: This was just a nit that I gardened.  Usually `is not None` can be simplified to just using truthiness of the thing.  What are you worried about erroring out in the open call?

deploy/awstools/awstools.py
timsnyder-siv: Yes.  This is a good catch.  I typed the docs before actually looking at the builtin Exception types and remembering that UserError isn't one of them...
```suggestion
        If `tags` are not passed, `additive` must be `True` or `ValueError` is thrown.
```

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
timsnyder-siv: Sure.  
```suggestion
``ClientError`` response in all AvailabilityZones (AZs) for your region.  In other words,
```

deploy/awstools/awstools.py
timsnyder-siv: How about `expand_runfarm_by_count`?  When I read `expand_runfarm_limit`, my brain thinks that it should take an integer limit, not be a bool.

abejgonzalez: That works for me

deploy/awstools/awstools.py
abejgonzalez: Oh. Derp. Yes this is fine.

timsnyder-siv:
I realized that I didn't expose `expand_runfarm_by_count` to the runtime.ini.  Are we going to rip it out or keep it for this PR?

deploy/awstools/awstools.py
timsnyder-siv: As I'm plumbing this in, I'm liking `always_expand_runfarm`.  The name makes more sense when looking at the runfarm ini and I'm going to change it again unless it gives people heartburn.

@abejgonzalez, David told me to plumb it in via other channels so I'll be pushing that shortly.

timsnyder-siv:
@davidbiancolin and @abejgonzalez thinking more about this, I think it makes sense to keep `always_expand` (fka `additive`, `expand_by_count`) on `launch_instances()` because other callers of the awstools function might want that behavior at some point.  I still think it is debatable whether `launchrunfarm` should ever do this but it might make sense to do this in some other context (perhaps if you want the whole 'task' to crash and burn if `launch_instances` throws).

timsnyder-siv:
Closing this in favor of the local PR #940

PR #717, opened 2021-02-06
Multibit ac doc tutorialsuite test targets

timsnyder-siv:
Okay @davidbiancolin , this is a concrete example of what I meant in https://github.com/firesim/firesim/pull/706/files#r571310835

Also, I found a link to the old https://github.com/ucb-bar/midas-examples in the RST docs that I also changed

### DIFF
--- a/docs/Advanced-Usage/Generating-Different-Targets.rst
+++ b/docs/Advanced-Usage/Generating-Different-Targets.rst
@@ -119,7 +119,7 @@ resides in ``sim/``.  These projects are:
    either BOOM or rocket pipelines, and should be your starting point if you're
    building an SoC with the Rocket Chip generator.
 2. **midasexamples**: the `Golden Gate example designs
-   <https://github.com/ucb-bar/midas-examples>`_, a set of simple chisel
+   <https://github.com/firesim/firesim/tree/dev/sim/src/main/scala/midasexamples>`_, a set of simple chisel
    circuits like GCD, that demonstrate how to use Golden Gate.  These are useful test
    cases for bringing up new Golden Gate features.
 3. **fasedtests**: designs to do integration testing of FASED memory-system timing models.
--- a/sim/src/main/scala/midasexamples/AutoCounterModule.scala
+++ b/sim/src/main/scala/midasexamples/AutoCounterModule.scala
@@ -71,6 +71,15 @@ class AutoCounterModuleChild extends MultiIOModule {
   io.oddlfsr := odd_lfsr
 }
 
+/** Demonstrate explicit instrumentation of AutoCounters via PerfCounter 
+ *
+ * Toplevel Chisel class suitable for use as a GoldenGate 'target' as described by the docs
+ *
+ * @see https://docs.fires.im/en/latest/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.html#ad-hoc-performance-counters
+ * @see https://docs.fires.im/en/latest/Advanced-Usage/Generating-Different-Targets.html
+ * @see AutoCounterF1Test
+ * @see PerfCounter
+ */
 class AutoCounterModule(implicit p: Parameters) extends PeekPokeMidasExampleHarness(() => new AutoCounterModuleDUT)
 
 class AutoCounterCoverModuleDUT extends Module {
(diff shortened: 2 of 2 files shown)


docs/Advanced-Usage/Generating-Different-Targets.rst
```diff
@@ -119,7 +119,7 @@ resides in ``sim/``.  These projects are:
    either BOOM or rocket pipelines, and should be your starting point if you're
    building an SoC with the Rocket Chip generator.
 2. **midasexamples**: the `Golden Gate example designs
-   <https://github.com/ucb-bar/midas-examples>`_, a set of simple chisel
+   <https://github.com/firesim/firesim/tree/dev/sim/src/main/scala/midasexamples>`_, a set of simple chisel
```
timsnyder-siv, line 122: Or, would you rather reference them like you do below without creating a hyperlink (see https://github.com/firesim/firesim/pull/717/files#diff-6ea46dd280d3c59fa262dd9fb27b7472124db0ceb33642c5e898d579ebda3707R152)

sim/src/main/scala/midasexamples/AutoCounterModule.scala
```diff
@@ -71,6 +71,15 @@ class AutoCounterModuleChild extends MultiIOModule {
   io.oddlfsr := odd_lfsr
 }
 
+/** Demonstrate explicit instrumentation of AutoCounters via PerfCounter 
+ *
+ * Toplevel Chisel class suitable for use as a GoldenGate 'target' as described by the docs
+ *
+ * @see https://docs.fires.im/en/latest/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.html#ad-hoc-performance-counters
```
timsnyder-siv, line 78: Is there a better way to reference explicit links into the docs that is more maintainable?

sim/src/main/scala/midasexamples/AutoCounterModule.scala
```diff
@@ -71,6 +71,15 @@ class AutoCounterModuleChild extends MultiIOModule {
...
+ * @see https://docs.fires.im/en/latest/Advanced-Usage/Generating-Different-Targets.html
+ * @see AutoCounterF1Test
```
timsnyder-siv, line 80: Do you guys publish the rendered ScalaDoc somewhere?  I'm curious about how these `@see` links will render but I'm pretty sure that having this here would have let me grok the testiness of what you were doing more quickly.

timsnyder-siv:
@davidbiancolin , if you like these example scaladocs, I wouldn't mind doing the same for the rest of the midasexamples.  It would give me a chance to read through them.

sim/src/main/scala/midasexamples/AutoCounterModule.scala
davidbiancolin: that's a good question. I'm not sure. There's both the brittleness of using a URL and the versioning problem we have to contend with here.

sim/src/main/scala/midasexamples/AutoCounterModule.scala
davidbiancolin: You can do it from sbt to build it yourself.  If you want to see the published stuff:
https://fires.im/firesim/latest/api/
Or for the release: 
https://fires.im/firesim/1.11.0/api/

davidbiancolin:
> @davidbiancolin , if you like these example scaladocs, I wouldn't mind doing the same for the rest of the midasexamples. It would give me a chance to read through them.

I'm down. I don't think you need to do them for all the examples. We might actually consider removing some of the ones that don't have an explicit purpose (specifically some of the holdovers from the chisel example designs).

timsnyder-siv:
> I don't think you need to do them for all the examples.

Maybe I should use the list of examples that are being used in TutorialSuite.scala as a guide for the ones to look at and add scaladoc to because they are actually being used as tests?

sim/src/main/scala/midasexamples/AutoCounterModule.scala
timsnyder-siv: Thanks for the pointer.  I was being lazy and not going to look for how they get published in the circleci stuff or trying to remember how to regen them locally.  I didn't see a link to the published scaladoc from either https://fires.im or https://docs.fires.im.

davidbiancolin: Where do you think is an appropriate place. You think we should just put the link in the front page of the docs? 
I think the only slight annoyance is getting the url to point to the right version. I'm sure there's a RST macro (or something) for that available in readthedocs.

timsnyder-siv:
@davidbiancolin can we merge this into your multibit autocounter branch?

davidbiancolin:
Go for it. :)

PR #722, opened 2021-02-10
[ci] Upgrade installed pip2 on ci instances

davidbiancolin:


timsnyder-siv (approved):
Sweet!  It makes things pass again.  You even made it match the version on the AMI.

One small thought (but maybe for future PR), should we document or potentially refactor everything that has to change when the AMI gets upgraded in one place?

davidbiancolin:
That's probably a good idea. It might make sense to do that during the process? 

There are other differences between the CI and the developer AMI, so we'll likely run into this again. Honestly, i'm hoping to change to using github actions + AWS instances to run CI. We have both funding and permission to go ahead with that.

PR #724, opened 2021-02-11
Tests For Printf Region-of-Interest, Trigger Predication; TriggerSink Sugar

davidbiancolin:
In light of all the activity around printfs and triggers recently, this PR:

- Adds tests to check printf cycle-based ROI definition
- Lengthens Printf based tests to ensure they require multiple `simif_t::pull`s
- Adds a method `whenEnable` to TriggerSink. This is sugar for a common use case
- Adds a midasexample to demonstrate predicating a printf with the trigger system


sim/midas/targetutils/src/main/scala/midas/annotations.scala
```diff
@@ -343,4 +343,21 @@ object TriggerSink {
       def toFirrtl = TriggerSinkAnnotation(targetWire.toTarget, clock.toTarget)
     })
   }
+
+  /**
+    * Syntatic sugar for a when context that is predicated by a trigger sink.
+    * Example usage:
+    * {{{
+    * TriggerSink.whenEnabled {
```
alonamid, line 351: I'm guessing this is sugar to enable the functionality while not embedding the trigger in the printf transform?

alonamid (approved):
LGTM

sim/midas/targetutils/src/main/scala/midas/annotations.scala
davidbiancolin: Yeah, you don't need to use this function, but it serves as shorthand for how i expected users should. 

We could couple printfs and triggers, as we do for autocounter. That might be a question for a wider discussion. I opted not to because it was simpler in the short run, and keeping things decoupled makes it easier to test these systems.

PR #726, opened 2021-02-12
Improve Chisel Naming via plugin; Remove chiselName Annotation

davidbiancolin:


sim/build.sbt
```diff
@@ -47,6 +50,12 @@ lazy val testchipip    = ProjectRef(chipyardDir, "testchipip")
 lazy val sifive_blocks = ProjectRef(chipyardDir, "sifive_blocks")
 lazy val firechip      = ProjectRef(chipyardDir, "firechip")
 
+lazy val chiselLib = "edu.berkeley.cs" %% "chisel3" % chiselVersion
```
jackkoenig, line 53: Why do you have `chiselLib` here if you're not using it?

abejgonzalez: Agreed. There is no point in using this if we aren't using SBT-Sriracha in FireSim.

abejgonzalez (approved):
LGTM

PR #727, opened 2021-02-23
First-pass AWS-hosted CI based off Abe's BOOM FireSim CI

davidbiancolin:
This transitions CI to run on AWS. This will enable us to run a more complete set of tests and, in the future, enable regeneration of AGFIs and running FPGA-based tests. Unlike BOOM CI, the support here spins up a fresh manager instance on each CI run.  Since we'll have multiple parallel workflows in flight, this seemed like the most trivial way to provide isolation.  Managers are identified using instance tags derived from each workflow's ID. 

The general process is:
1. Spin up an instance (`launch-manager-instance.py`)
2. Initialize the instance (machine-launch-script + build-setup, `initialize-manager.py`)
3. (Parallel) Run scala tests

CI instances are stopped and terminated using one of two mechanisms established by `initialize-manager`:
- a simple time-to-live sleeps all instances after the time has elapsed
- `workflow-monitor` tracks the state of the active workflow
   - on success, it terminates instances 
   - on failure, it stops instances 

CI runs a cleanup workflow every 12h to terminate instances that are >24h old, regardless of their state. This will give us some time to restart an instance with a failure with a debug.

Current limitations:
 - Currently only those credentials under my biancolin-aws-educate will be able to interact with these instances
    - TODO Open issue to generalize access to a workflows instance to authenticated users
 - Rerun from failed, and debug with SSH do _not_ function correctly since these generate a new workflow ID
    - You must rerun from start 
    - TODO Open Issue to reuse a manager across these sessions where possible.
  
Long Term TODOs:
- Method for passing aws credentials to manager instance that is less sketchy
- Restrict IAM permissions for CI credentials


.circleci/initialize-manager.py
```diff
@@ -0,0 +1,52 @@
...
+    max_runtime (seconds): The maximum uptime this manager should have be for it is terminated
+        This covers potential livelock scenario in one of the jobs (that produces fs activity)
+
+    filesystem_timeout (seconds):  If no fs activity is detected in ~/firesim after 2x this
+        interval, the manager is terminated.  CircleCI provides no mechanism for
+        doing cleanup after all jobs are done (even if some of those jobs fail).
+        So here we use FS activity as a proxy for doneness
+    """
+
+    # Catch any exception that occurs so that we can gracefully teardown the local jjjjjjjjjjjjjjjjjjjjjjj
```
abejgonzalez, line 21: jjjjjjjjjjjjjjjjj

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        put(ci_workdir + "/scripts/machine-launch-script.sh", manager_home_dir)
+        with cd(manager_home_dir):
+            put(ci_workdir + "/scripts/machine-launch-script.sh", ".")
```
abejgonzalez, line 25: Why are you re-copying the `machine-launch-script`?

abejgonzalez: Side note: Maybe this is better named `machine-launch-init-script`?

.circleci/config.yml
```diff
@@ -8,6 +10,9 @@ executors:
     environment:
         JVM_MEMORY: 3500M # Default JVM maximum heap limit
         LANG: en_US.UTF-8 # required by sbt when it sees boost directories
+  aws:
+    docker:
+    - image: cimg/python:2.7-node
```
abejgonzalez, line 15: Is this just a basic image with Python 2.7?

.circleci/launch-manager-instance.py
```diff
@@ -0,0 +1,54 @@
...
+
+def main():
+    manager_instance = get_manager_instance()
+    if (manager_instance is not None):
+        print("There is an existing manager instance for this CI workflow:")
+        print(instance_metadata_str(manager_instance))
+        sys.exit(0)
+
+
+    manager_instance = awstools.launch_instances(
```
abejgonzalez, line 28: Wow. So nice.

deploy/awstools/awstools.py
```diff
@@ -43,6 +44,11 @@ def aws_resource_names():
         'runfarmprefix':     None,
     }
 
+    # The code below assumes we're running on an AWS instance. When running on a CI instance,
+    # early out with the default dictionary
+    if os.environ.has_key("CIRCLE_WORKFLOW_ID"):
```
abejgonzalez, line 49: Why is this needed? Won't you need this in the future when running normal FireSim FPGA simulations? Or maybe I'm misunderstanding what this is?

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,52 @@
...
+        with cd(manager_fsim_dir), prefix("source ./sourceme-f1-manager.sh"):
+            run(".circleci/firesim-managerinit.expect {} {} {}".format(
+                os.environ["AWS_ACCESS_KEY_ID"],
+                os.environ["AWS_SECRET_ACCESS_KEY"],
+                os.environ["AWS_DEFAULT_REGION"]))
+
+        with cd(manager_ci_dir):
+            # Set up a pair of crude checks to powerdown the instance
+            # The sleep 1 is required for the screen command to not be DOA
+            run("screen -d -m ./manager-watchdog.sh {} {} {}; sleep 1".format(max_runtime, filesystem_timeout, ci_workflow_id))
```
abejgonzalez, line 43: I agree with your description assessment. This should probably be polling the workflow state to shut down in this current setup. 

With that being said, this might be problematic when running longer jobs (AFI creation, FPGA-sims). In that case, there is no "nice" way to run a long CI job (IIRC longer than 5hrs). For example, this prevents you from having a single job called `afi-creation` that lasts a long time. In that case, the BOOM CI creates separate workflows (spawned by the CircleCI API) to finish up when the AFI build passes/fails. This creates an issue where a single manager instance can have multiple workflows associated with it (assuming some version of what I did... spawning new workflows... is done here). Maybe the solution, in this case, is to associate one manager with multiple workflows and if all workflows are completed then the manager instance terminates everything on it and shuts down (note how "shutting down" is now a bit more complicated because if FPGA-sims are going... then you need to terminate those F1 instances). However, I think this should be fairly straight forward to do and much cleaner in Python than in Bash :).

abejgonzalez:
Overall, I think this is an extremely clean implementation of the CI work previously done (thanks `fabric` + `boto`). I think the next jump is how to do this cleanly for AFI builds / FireSim-sims... but that's out-of-scope for this PR for now.

I'll see if I can get some help from Chick on some of this AFI/FireSim-sim stuff.

.circleci/initialize-manager.py
abejgonzalez: Actually on second thought... the following statement is wrong:

> Maybe the solution, in this case, is to associate one manager with multiple workflows and if all workflows are completed then the manager instance terminates everything on it

Technically, if we are using CircleCI to do this there might be periods where there are no workflows running but the AWS FireSim manager is running (building an AFI, running SPEC, etc). So there has to be some sort of tracking on the manager... just like what I did in the BOOM CI (have some file indicate what is running or isn't).

davidbiancolin: Yeah, some manager side tracking will be required.

Although, I guess you could continuously spin up new polling jobs when they are about to reach their time limit... #hacks

.circleci/config.yml
davidbiancolin: Yes.

.circleci/initialize-manager.py
davidbiancolin: Oops, i'll remove that.

.circleci/config.yml
timsnyder-siv: It's the newer 'cimg' instead of 'circleci' namespace python image with python 2.7 and node.js installed.  

See:
- https://circleci.com/docs/2.0/circleci-images/#next-gen-circleci-images  regarding 'cimg' vs 'circleci' dockerhub namespace
- https://hub.docker.com/r/cimg/python for a description of what the '2.7-node' tag means.

.circleci/manager-watchdog.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+max_runtime_seconds=$1
+timeout_seconds=$2
+ci_workflow_id=$3
+
```
timsnyder-siv, line 6: It might be nice to have a comment that describes the high-level intent of this watchdog script.

.circleci/run-scala-test.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+
+import sys
+
+from fabric.api import execute
```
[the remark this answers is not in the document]
davidbiancolin: Yeah, that's right. I was fooling around with something and forgot to put it back in.

davidbiancolin:
@abejgonzalez @timsnyder-siv This is ready for a re-review.

timsnyder-siv (commented):
I think there is only one or two broken things, the rest are comments you can feel free to take or leave. 

Broken stuff:
* `DryRun=True` in `cull-old-ci-instances.py`

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,55 @@
...
+        with cd(manager_fsim_dir), prefix("source ./sourceme-f1-manager.sh"):
+            run(".circleci/firesim-managerinit.expect {} {} {}".format(
+                os.environ["AWS_ACCESS_KEY_ID"],
+                os.environ["AWS_SECRET_ACCESS_KEY"],
+                os.environ["AWS_DEFAULT_REGION"]))
+
+        with cd(manager_ci_dir):
+            # Put a baseline time-to-live bound on the manager.
+            # Instances will be stopped and cleaned up in a nighlty job.
+            # Hack: The final sleep 1 ensures the screen comes up.
```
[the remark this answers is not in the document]
davidbiancolin: I'll try adding `-nohup`. That does sound like it could be an issue.

.circleci/workflow-monitor.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# Terminate instances if:
+#   - the workflow is successful (all jobs complete successfully)
+#   - the workflow is cancelled
+# Stop instances if:
+#   - the workflow has failed (all jobs have completed, but at least one has failed)
+#
+# Other states to consider: not_run, on_hold, unauthorized
+
+import httplib
```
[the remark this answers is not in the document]
davidbiancolin: I had not, thanks. :)

.circleci/workflow-monitor.py
```diff
@@ -0,0 +1,63 @@
...
+            state = res_dict["status"]
+
+            print "Workflow {} status: {}".format(workflow_id, state)
+            if state == "success" or state == 'canceled':
+                terminate_workflow_instances(workflow_id)
+                exit(0)
+            elif state == "failed":
+                stop_workflow_instances(workflow_id)
+                exit(0)
+
```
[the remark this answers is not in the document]
davidbiancolin: There are lots of states where we should do nothing, (e.g, failing --> we need to let the other jobs finish) maybe it's better if I explicitly codify that?

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,55 @@
...
+            run("screen -S ttl -dm bash -c \'sleep {}; ./change-workflow-instance-states.py {} stop\'; sleep 1".format(max_runtime, ci_workflow_id))
```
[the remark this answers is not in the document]
davidbiancolin: Hmm, that may be a good idea. In the short term it may be simpler to just remove it altogether.

.circleci/workflow-monitor.py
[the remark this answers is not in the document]
timsnyder-siv: I think it would be good to codify them.  To make the code less clunky, I'd probably use a dict of callables to map from the state to the action, like this:
```
# add an exit=None parameter to your terminate/stop_workflow_instances helpers
# and 'exit(exit) if exit' at the end of them.  How redundant is that?  Maybe call it exit_code?

term_and_exit_0 = lambda: terminate_workflow_instances(workflow_id, exit=0)
stop_and_exit_1 = lambda: terminate_workflow_instances(workflow_id, exit=0)
# my fabulous no-op lambda isn't needed anymore with the do_nothing_states factored
# out into a list that is checked after looking for an actionable state in wkflow_actions but here
# it is if you end up deciding to list the do_nothing_states in the wkflow_actions and get rid of the separate list
do_nothing = lambda: None

# setup closures to call based on workflow states where something needs to be done
wkflow_actions = {
   success:  term_and_exit_0
   canceled: term_and_exit_0
   failed: stop_and_exit_0
}

do_nothing_states = [
    'failing',
    # enumerate the remaining ones
]


action = wkflow_actions.get(state)
if action:
    action()
elif state in do_nothing_states:
    pass
else:
    print('this is where you should have an error condition where you complain about not knowing what to do')
    raise FreakOut
```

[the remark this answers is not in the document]
timsnyder-siv: Warning: I wrote the big block of code above only in the PR and didn't try to even run it through python .  Sharp edges.

.circleci/initialize-manager.py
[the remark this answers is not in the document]
timsnyder-siv: Unless you're confident that any use of the simple timeout would happen way, way after circleci will have put the workflow in the timeout corner.  Then, it wouldn't matter.  I just wasn't certain of whether you could end up in some weird state where circleci thinks the workflow is still going but the manager has been torn down.  I guess that you'll get an error to the circleci console when Fabric running on the circlci runner loses connectivity with the manager.

I was overthinking this.  I think you should just leave what you have.

.circleci/config.yml
```diff
@@ -17,6 +22,13 @@ commands:
           command: |
             cd scripts && /usr/bin/bash ./machine-launch-script.sh
 
+  install-fabric:
+    description: "Installs fabric for manager-managing CI container"
```
timsnyder-siv, line 26: Maybe call this `install-ci-python-reqs` instead?  For the description "Install minimal python dependencies for CI"?

.circleci/initialize-manager.py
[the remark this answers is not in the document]
timsnyder-siv: I guess in the spirit of overthinking, if you want a simple timeout that stops the instance, you might be best off by appending `sudo shutdown -h +<minutes_to_live>` at the end of the userdata of every host you launch.  Then, you're no longer depending on the manager being able to contact the EC2 API and get it to do the shutdown, each worker, everywhere will know its simple ttl and shut itself off when it expires.

It also has the added benefit of being able to differentiate between hosts that shut themselves down vs being shutdown by the manager (or anything else using the EC2 API) via https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Instance.state_reason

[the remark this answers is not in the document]
timsnyder-siv: If you decide to distribute the simple shutdown to all hosts, you might also want to explicitly specify `InstanceInitiatedShutdownBehavior='stop'` when you create_instances.  I'm pretty sure that's the default because I use `shutdown -h ` and it stops my FPGA AMI instances.

.circleci/cull-old-ci-instances.py
```diff
@@ -0,0 +1,32 @@
...
+    current_time = datetime.datetime.utcnow().replace(tzinfo=pytz.UTC)
+
+    # Grab all instances with a CI-generated tag
+    all_ci_instances_filter = {'Name': 'tag:' + unique_tag_key, 'Values' : ['*']}
+    all_ci_instances = get_instances_with_filter([all_ci_instances_filter], allowed_states=['*'])
+
+    client = boto3.client('ec2')
+    print "Terminated Instances:"
+    for inst in all_ci_instances:
+        lifetime_secs = (current_time - inst["LaunchTime"]).total_seconds()
```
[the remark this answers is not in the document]
timsnyder-siv: I'm glad I went and looked for more details on this.  Time since last state transition isn't a thing I guess.  I think I was remembering that you can cull instances that have been stopped for a period of time by parsing the datetime out of the transition reason string as in https://stackoverflow.com/questions/41231630/checking-stop-time-of-ec2-instance-with-boto3

Maybe not exactly what you're trying to do here but in case you find it useful there it is.

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,55 @@
...
+import traceback
+
+from fabric.api import *
+
+from common import *
+# This is expected to be launch from the ci container
+from ci_variables import *
+
+def initialize_manager(max_runtime):
+    """ Preforms the prerequisite tasks for all CI jobs that will run on the manager instance
```
abejgonzalez, line 12:
```suggestion
    """ Performs the prerequisite tasks for all CI jobs that will run on the manager instance
```

.circleci/initialize-manager.py
```diff
@@ -0,0 +1,55 @@
...
+            run("chmod +x machine-launch-script.sh")
+            run("sudo ./machine-launch-script.sh")
+            run("git clone https://github.com/firesim/firesim.git")
+
+        with cd(manager_fsim_dir):
+            run("git checkout " + ci_commit_sha1)
+            run("./build-setup.sh --fast")
+
+        with cd(manager_fsim_dir), prefix("source ./sourceme-f1-manager.sh"):
+            run(".circleci/firesim-managerinit.expect {} {} {}".format(
```
abejgonzalez, line 36: FYI we can simplify the `expect` here by just using a configuration file and having this expect just click enter: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html.

.circleci/cull-old-ci-instances.py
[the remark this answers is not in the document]
davidbiancolin: Yeah I was aware of the string in the transition reason. It's something to consider doing in the future.

.circleci/initialize-manager.py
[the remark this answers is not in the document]
davidbiancolin: I'll consider switching to this when i start to scale this out.

.circleci/initialize-manager.py
davidbiancolin: Yeah that's much clearner. Though I think we should just split out the one-time aws configure from managerinit, as @timsnyder-siv has previously suggested (IIRC),  and then eventually drop the expect altogether.

deploy/awstools/awstools.py
```diff
@@ -43,17 +44,17 @@ def aws_resource_names():
...
+    # first get this instance's ID, if we're not on AWS, fallback to default dictionary
     resp = None
     try:
+        res = None
+        with hide('everything'):
+            res = local("""curl -s http://169.254.169.254/latest/meta-data/instance-id""", capture=True)
+        instanceid = res.stdout
+        rootLogger.debug(instanceid)
+
+        # try to get tags. if we don't have permission for this. return False
```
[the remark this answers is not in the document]
davidbiancolin: I don't really want to screw with catching a more specific exception unless Sagar can confirm the AuthorizationError is what he sees in practise. I'll clean up the curl handling though.

.circleci/workflow-monitor.py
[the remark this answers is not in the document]
timsnyder-siv: I'd still be worried about AWS adding some new state that this code doesn't handle but it isn't a big enough deal to postpone merging this.

timsnyder-siv (approved):
This is going to be awesome!  Thanks for doing this.

PR #728, opened 2021-02-25
Add option to remove -retiming flag from Vivado

jerryz123:
Depends on https://github.com/firesim/aws-fpga-firesim/pull/34

### DIFF
--- a/sim/firesim-lib/src/main/scala/util/Configs.scala
+++ b/sim/firesim-lib/src/main/scala/util/Configs.scala
@@ -27,6 +27,9 @@ object BuildStrategies {
   object Timing extends IsBuildStrategy { val flowString = "TIMING" }
   object Explore extends IsBuildStrategy { val flowString = "EXPLORE" }
   object Congestion extends IsBuildStrategy { val flowString = "CONGESTION" }
+  // This is the same as the Timing strategy, except it removes the -retiming flag to avoid Vivado crashes
+  // during explicit retiming
+  object NoRetiming extends IsBuildStrategy { val flowString = "NORETIMING" }
   // This is the strategy AWS uses if you give it a bogus strategy string
   object Default extends IsBuildStrategy { val flowString = "DEFAULT" }
 }
@@ -70,3 +73,6 @@ class ILADepth16384 extends WithILADepth(16384)
 
 
 class Congestion extends WithBuildStategy(BuildStrategies.Congestion)
+class Timing extends WithBuildStategy(BuildStrategies.Timing)
+class Explore extends WithBuildStategy(BuildStrategies.Explore)
+class NoRetiming extends WithBuildStrategy(BuildStrategies.NoRetiming)
--- a/sim/firesim-lib/src/main/scala/util/Configs.scala
+++ b/sim/firesim-lib/src/main/scala/util/Configs.scala
@@ -35,7 +35,7 @@ object BuildStrategies {
 }
 
 // Overrides the AWS default strategy with a desired one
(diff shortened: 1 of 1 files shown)


davidbiancolin (approved):
LGTM. 

Though you might consider using a different base strategy if you're disabling retiming, since in many cases your fmax is going to be poor (e.g., basic, explore). Some of the other strats have significantly better runtime.

timsnyder-siv:
Is there something specific you're running into that you need to avoid the `-retiming` option?  I'm curious to know more.

jerryz123:
Our machine learning accelerator design utilizes many floating-point pipelines, which must be retimed. However, when the `-retiming` flag is enabled, Vivado will sometimes crash consistently with an internal error during the retiming passes, regardless of what target frequency I set.

Removing the retiming flag prevents the crash, and lets the design build successfully. 

David and I both suspect that updating to a newer version of Vivado might solve the issue, so this is an interim solution.

timsnyder-siv:
Ok.  I'm also interested in why `-retiming` isn't correctly retiming a block that I expect to be retimed in one of our cores (haven't seen the crash yet).  I was trying to look through the Vivado docs to learn more about the requirements for being able to retime.  I see messages in our log about some modules being retimed  and many others being skipped because their effective depth is much lower than the effective depth of the entire design but I see no mention of the module that ends up having the WNS path (that also happens to have gate depth equal to the one I was seeing in the retiming messages).

I was curious whether you guys have ever tried do anything with `rocketchip.util.RetimeModuleAnnotation` feeding into constraints for Vivado.

If newer Vivado doesn't fix the problem, we should definitely add this to the list of things to bring up with the Vivado devs and I'm happy to help push on that.

@davidbiancolin can we put this topic on the agenda for next Tuesday?

davidbiancolin:
> @davidbiancolin can we put this topic on the agenda for next Tuesday?

Sure thing. 

There's a long history of Vivado retiming not working for our designs, so i'm not altogether surprised.

jerryz123:
Not sure why this is failing CI

sim/firesim-lib/src/main/scala/util/Configs.scala
```diff
@@ -70,3 +73,6 @@ class ILADepth16384 extends WithILADepth(16384)
 
 
 class Congestion extends WithBuildStategy(BuildStrategies.Congestion)
+class Timing extends WithBuildStategy(BuildStrategies.Timing)
+class Explore extends WithBuildStategy(BuildStrategies.Explore)
+class NoRetiming extends WithBuildStrategy(BuildStrategies.NoRetiming)
```
abejgonzalez, line 78:
```suggestion
class Congestion extends WithBuildStrategy(BuildStrategies.Congestion)
class Timing extends WithBuildStrategy(BuildStrategies.Timing)
class Explore extends WithBuildStrategy(BuildStrategies.Explore)
class NoRetiming extends WithBuildStrategy(BuildStrategies.NoRetiming)
```

abejgonzalez: Additionally, the config. fragment needs to be changed from `WithBuildStategy` to `WithBuildStrategy`

abejgonzalez (changes requested):

abejgonzalez (approved):
Yay spelling!

jerryz123:
Can this be merged? Not sure why CI is failing

davidbiancolin:
Can you rebase it and run CI again?

Issue #592, opened 2020-07-01
synthesized_prints_t::print_format() missing '%c', '%n', '%N' from Chisel3.printf


timsnyder:
When annotating a printf in my closed-source design, I have need for a programmatic ASCII character in a field.  Looking at the [Chisel scaladoc for printf](https://github.com/freechipsproject/chisel3/blob/fbf5e6f1a0e8bf535d465b748ad554575fe62156/core/src/main/scala/chisel3/Printf.scala#L63-L69), I didn't see '%s' but I did see '%c', so I used it.  However, [synthesized_prints_t::print_format() doesn't have support for '%c'](https://github.com/firesim/firesim/tree/db1915b19b119fadeba15f862cc9a904d66797ff/sim/midas/src/main/cc/bridges/synthesized_prints.cc#L160-L173) (but I notice that it does support '%s').

I dug deeper into the actual code just to make sure that the scaladoc wasn't out of date and here's the places where the format specifiers are defined:
- https://github.com/freechipsproject/chisel3/blob/fbf5e6f1a0e8bf535d465b748ad554575fe62156/core/src/main/scala/chisel3/Printable.scala#L87-L90
- https://github.com/freechipsproject/chisel3/blob/fbf5e6f1a0e8bf535d465b748ad554575fe62156/core/src/main/scala/chisel3/Printable.scala#L150-L154

Note that synthesized_prints::print_format() looks like it will also trip over '%n' and '%N'.  It would be good to document any divergence from what `chisel3.printf` supports.  I can definitely see how it is useful to add in support for '%s'.

I'd like the printf to work when it isn't annotated for MIDAS and only have to wrap it in the annotation.  Should I add support for '%c' and submit a PR or am I missing something?

davidbiancolin:
I think adding support for missing format specifiers would be most welcome. It should be fairly straight forward to support %c. You could add a couple printfs with the new specifiers to one of the midasexample PrintfModules to sanity check everything works.

PR #735, opened 2021-03-24
[printf] Support %c format specifiers

davidbiancolin:
See: https://groups.google.com/g/firesim/c/jSDHYbiSpuQ and #592 

It seems %N and %n are handled by chisel and do not show up in emitted format strings.


timsnyder-siv (approved):
LGTM

PR #736, opened 2021-03-24
Test FMR of threaded midasexamples

albert-magyar:
This adds a test to make sure that a threaded midasexamples design has the expected FMR -- i.e., if it is as slow as would be expected for the degree of threading. This is a decent "smoke test" of whether threading is working as intended.

There is a TODO for the design that mixes threading and multi-cycle RAM models, as this seems to not be threading as intended in recent commits to dev.


davidbiancolin (approved):
LGTM. But could you add some (scala) doc for the Chisel modules, if there is none already, to explain what they test / demonstrate.  

Like what Tim did this PR #717. The comments need not be as detailed.

PR #738, opened 2021-03-28
Changes to allow limited use of FIRRTL 1.4 dedup to maintain support for composed optimizations

albert-magyar:
The new implementation of `DedupModules` in FIRRTL 1.4 can interact poorly with some FAME annotations. However, it generally performs better.

This PR uses three tweaks to allow this upstream dedup to be used sparingly to support threading in combination with other features such as debug tools or memory optimizations:
1) Only dedup the circuit after all models have been promoted to the top level. This limits the form of the path for the instance and reference targets that appear in the FAME annotations at the time of dedup. This is accomplished by introducing a new `EnableAndRunDedupOnce` transform that temporarily overrides `NoCircuitDedupAnnotation`s.
2) Protect extracted combinational memory models from being deduplicated, as their connection annotations are currently brittle. This could potentially be a straightforward fix in a follow-on PR.
3) Clean up some FAME annotations as soon as they are no longer used in the compiler; this helps reduce dedup complexity.

While this limited use of dedup fails to capture the full compile-time benefits of reduced FIRRTL code size, no dedup was performed at all in dev for a while up to #693, so this is not necessarily critical for adequate performance. This PR also adds a test that checks the FMR of a multi-optimized `midasexamples` unit test as a proxy to help provide evidence that optimizations are working as intended.

It is also possible to remove the in-tree `DedupModules` with these changes; I will look at investigating this further.


davidbiancolin (approved):
Could you remove the out of tree dedup and add an issue to track remaining annotation deduplication problems? It can be spartan, and just link here.

davidbiancolin:
@albert-magyar You can rebase on dev and your CI should run fine.

albert-magyar:
I will probably try running an instance-threaded BOOM simulation after that rebase just to double-check.

albert-magyar:
@davidbiancolin I was able to run Linux on `MTModels` + `MCRams` optimized multi-BOOM configs with this.

PR #742, opened 2021-03-30
FPGA Dev. AMI 1.10 Bump

abejgonzalez:
Related PRs: https://github.com/firesim/aws-fpga-firesim/pull/36

This PR updates AWS-FPGA so that FireSim can support the FPGA Developer AMI 1.10.X which contains Vivado 2020.2. To test this you need to use the FPGA Developer AMI 1.10.X (instead of 1.6.X) and you need to use the `machine-launch-script.sh` changes to properly check out all dependencies.

Things to check:
- [ ] All default AFI's work
- [ ] GG optimizations work (MCRams, MT)
- [ ] ILA works

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -11,7 +11,7 @@ from fabric.api import local, hide
 rootLogger = logging.getLogger()
 
 # this needs to be updated whenever the FPGA Dev AMI changes
-f1_ami_name = "FPGA Developer AMI - 1.6.1-40257ab5-6688-4c95-97d1-e251a40fd1fc-ami-01d5f32a3d517960b.4"
+f1_ami_name = "FPGA Developer AMI - 1.10.0-40257ab5-6688-4c95-97d1-e251a40fd1fc"
 
 def aws_resource_names():
     """ Get names for various aws resources the manager relies on. For example:
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -47,22 +47,23 @@ sudo yum -y install graphviz python-devel
 # used for CI
 sudo yum -y install expect
 
-python3 -m pip install -U pip setuptools
-python3 -m pip install fabric==1.14.0
-python3 -m pip install boto3==1.6.2
-python3 -m pip install colorama==0.3.7
-python3 -m pip install argcomplete==1.9.3
-python3 -m pip install graphviz==0.8.3
+# last working pip version before deprecation
+sudo pip2 install --upgrade pip==20.3.4
+sudo pip2 install fabric==1.14.0
+sudo pip2 install boto3==1.6.2
+sudo pip2 install colorama==0.3.7
+sudo pip2 install argcomplete==1.9.3
+sudo pip2 install graphviz==0.8.3
 # for some of our workload plotting scripts
-python3 -m pip install --upgrade --ignore-installed pyparsing
-python3 -m pip install numpy==1.16.6
-python3 -m pip install kiwisolver==1.1.0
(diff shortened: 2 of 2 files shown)


abejgonzalez:
Current issue:

```
[192.168.0.15] out: ERROR: [Place 30-838] The following clock nets need to use the same clock routing resource, as their clock buffer sources are locked to sites that use the same routing track. One or more loads of these clocks are locked to clock region(s) X2Y14 X3Y14 which causes the clock partitions for these clocks to overlap. This creates unresolvable contention on the clock routing resources. If the clock buffers need to be locked, we recommend users constrain them to a clock region and not to specific BUFGCE/BUFG_GT sites so they can use different routing resources. If clock sources should be locked to specific BUFGCE/BUFG_GT sites that share the same routing resources, make sure loads of such clocks are not constrained to the same region(s). Clock nets sharing routing resources:
[192.168.0.15] out: WRAPPER_INST/CL/firesim_top/top/sim/target/FireSim_/lazyModule/system/tile_prci_domain_2/tile_reset_domain_tile/ptw/valid_1_0[1023]_i_2__1_n_0
[192.168.0.15] out: WRAPPER_INST/SH/kernel_clks_i/clkwiz_kernel_clk0/inst/CLK_CORE_DRP_I/clk_inst/clk_out2
[192.168.0.15] out: ERROR: [Place 30-678] Failed to do clock region partitioning: failed to resolve clock partition contention for locked clock sources.
[192.168.0.15] out: Resolution: Try removing the area constraints set on the clocks (source and/or loads) that compete for resources in the same clock region(s). In case an area constraint on the clock loads is necessary, please either extend the area constraint to cover the clock source, or make sure that the clock source is constrained/placed in a clock region in the same row or column as one of the regions in the specified area constraint. Also try reducing the amount of clock resources in your design, by either combining some clock nets or by changing the placement of clock primitives to reduce the distance between the source and loads of each clock net involved in the area with higher clock routing utilization.
```

Steps to reproduce:
1.  Checkout Chipyard `dev` : `ed2bfa`.
2. `./scripts/init-submodules-no-riscv-tools.sh`.
3. Checkout this PR
4. `./scripts/build-toolchains.sh ec2fast`
5. `source ./env.sh && ./scripts/firesim-setup.sh --fast`
6. Classic FireSim setup steps (move `.pem` and `firesim managerinit`)
7. `firesim buildafi` with just the `firesim-rocket-quadcore-no-nic-l2-llc4mb-ddr3` config.

davidbiancolin:
Oof. That's not a good sign.

davidbiancolin:
Can you attach your build log?

abejgonzalez:
[2021-03-30--08-08-58-buildafi-AB05J75KSGMK8BTU.log](https://github.com/firesim/firesim/files/6269309/2021-03-30--08-08-58-buildafi-AB05J75KSGMK8BTU.log)

timsnyder-siv:
Hey @abejgonzalez , I see that the AWS guys replied with a followup question for [David's Forum Post about the [Place 30-687] error you mentioned above](https://forums.aws.amazon.com/thread.jspa?threadID=338735&tstart=0).  Do you know if David was still seeing the error after he 'tried the fix of adding dummy loads for good measure'?

I figured if he talked to you about that detail, you might be able to chime in and keep AWS making progress while David is out.  If not, no biggie.

davidbiancolin:
I do. although I had the loads flip between false and true instead of just
tying them off. I'm not sure that makes a difference

timsnyder-siv:
I forwarded @davidbiancolin's reply into the AWS forum post.

timsnyder-siv:
I think this is subsumed by #788

PR #750, opened 2021-04-06
CreateParametersInstancePhase now takes ConfigPackageAnnotation from AnnotationSeq

sequencer:
If a user use Scala directly invoke `GoldenGateStage`, rather than using cli:
```scala
  Seq(
    new chisel3.stage.ChiselStage,
    new midas.stage.GoldenGateStage
  ).foldLeft(
    AnnotationSeq(
      Seq(
        ChiselGeneratorAnnotation(() => new firesim.midasexamples.GCD()(new firesim.midasexamples.NoConfig)),
        ConfigParametersAnnotation((new midas.HostDebugFeatures).alter(new firesim.midasexamples.DefaultF1Config)),
        NoRunFirrtlCompilerAnnotation
      )
    )
  ) { case (anno, stage) => stage.transform(anno) }
```
This patch makes this snippet be able to work, before this patch `CreateParametersInstancePhase` will directly throw with errors.

### DIFF
--- a/sim/midas/src/main/scala/midas/stage/phases/CreateParametersInstancePhase.scala
+++ b/sim/midas/src/main/scala/midas/stage/phases/CreateParametersInstancePhase.scala
@@ -2,14 +2,14 @@
 
 package midas.stage.phases
 
+import firrtl.AnnotationSeq
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.options.{Phase, PreservesAll, TargetDirAnnotation, Unserializable}
+import freechips.rocketchip.config.{Config, Parameters}
 import midas.OutputDir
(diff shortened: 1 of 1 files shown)


davidbiancolin:
Hey Jiuyang,

Can you not just use the the GG phase instead of the stage, if you are invoking this from scala? My understanding is that a `stage` is really a phase intended for CLI use. 

I think it might be cleaner if this was split into two phases.
- A "GoldenGateChecks" phase that ensures exactly one instance of the Parameters instance annotation exists.
- The existing parameters instance phase

In your phase manager you could omit the second phase, since you are directly providing the annotation it creates.

davidbiancolin:
Closing due to inactivity.

PR #753, opened 2021-04-14
Arbiter Instead of DualQueue to Prevent Write Starvation

alonamid:
DualQueue can result in starvation of write request, and specifically cache eviction requests which can generate incredible tail latencies (12000 cycles in one case...).
Using a single queue and an arbiter should alleviate some of the complexity and edge cases of DualQueue.


davidbiancolin (approved):
LGTM. Pending CI

jerryz123 (approved):

davidbiancolin:
@alonamid Can you add a change log entry for this

PR #754, opened 2021-04-21
Check sns topic permission at start of buildafi and warn user

rahulb10:
Rather than exiting with a unhandled exception when SNS-related permissions failures happen at the end of `buildafi`, check whether the topic exists (or can be created) early in `buildafi` and warn the user that the script will be unable to send email (and log the details of the exception) but continue to finish `buildafi` because failure in sending the notification probably should not be a fatal error for the manager.

This PR also introduces `pytest` driven unit tests for the firesim manager.  The tests make use of `moto` to mock the backend of `boto` and prevent tests from actually calling out to AWS API's.  They also utilize `unittest.mock` and `botocore.stub.Stubber` to inject desired testing stimulus to the code under test.

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -359,9 +359,11 @@ def get_snsname_arn():
     except client.exceptions.ClientError as err:
         if 'AuthorizationError' in repr(err): 
             rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
-            rootLogger.critical(err)
+
         else:
             rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
+
+        rootLogger.critical(err)
         assert False
         
     return response['TopicArn']
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -344,8 +344,8 @@ def auto_create_bucket(userbucketname):
             rootLogger.critical(repr(exc))
             assert False
 
-def subscribe_to_firesim_topic(email):
-    """ Subscribe a user to their FireSim SNS topic for notifications. """
+def create_topic():
+    """ Send a FireSim SNS Email notification. """
     client = boto3.client('sns')
 
     aws_resource_names_dict = aws_resource_names()
@@ -355,8 +355,13 @@ def subscribe_to_firesim_topic(email):
     response = client.create_topic(
         Name=snsname
     )
-    arn = response['TopicArn']
+    return response
 
+def subscribe_to_firesim_topic(email):
+    """ Subscribe a user to their FireSim SNS topic for notifications. """
+
+    arn_dict = create_topic()
+    arn = arn_dict['TopicArn']
     response = client.subscribe(
         TopicArn=arn,
         Protocol='email',
@@ -371,18 +376,9 @@ receive any notifications until you click the confirmation link.""".format(email
     rootLogger.info(message)
 
 def send_firesim_notification(subject, body):
-    """ Send a FireSim SNS Email notification. """
-    client = boto3.client('sns')
 
-    aws_resource_names_dict = aws_resource_names()
-    snsname = aws_resource_names_dict['snsname']
-
-    # this will either create the topic, if it doesn't exist, or just get the arn
-    response = client.create_topic(
-        Name=snsname
-    )
-
-    arn = response['TopicArn']
+    arn_dict = create_topic()
+    arn = arn_dict['TopicArn']
 
     response = client.publish(
         TopicArn=arn,
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -344,24 +344,32 @@ def auto_create_bucket(userbucketname):
             rootLogger.critical(repr(exc))
             assert False
 
-def create_topic():
-    """ Send a FireSim SNS Email notification. """
+def get_snsname_arn():
+    """ If the Topic doesn't exist create it, send catch exceptions while creating. Or if it exists get arn """
     client = boto3.client('sns')
 
     aws_resource_names_dict = aws_resource_names()
     snsname = aws_resource_names_dict['snsname']
 
-    # this will either create the topic, if it doesn't exist, or just get the arn
-    response = client.create_topic(
-        Name=snsname
-    )
-    return response
+    response = None
+    try: # this will either create the topic, if it doesn't exist, or just get the arn
+        response = client.create_topic(
+            Name=snsname
+        )
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
+            rootLogger.critical(err)
+        else:
+            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
+        assert False
+        
+    return response['TopicArn']
 
 def subscribe_to_firesim_topic(email):
     """ Subscribe a user to their FireSim SNS topic for notifications. """
 
-    arn_dict = create_topic()
-    arn = arn_dict['TopicArn']
+    arn = get_snsname_arn()
     response = client.subscribe(
         TopicArn=arn,
         Protocol='email',
@@ -377,8 +385,7 @@ receive any notifications until you click the confirmation link.""".format(email
 
 def send_firesim_notification(subject, body):
 
-    arn_dict = create_topic()
-    arn = arn_dict['TopicArn']
+    arn = get_snsname_arn()
 
     response = client.publish(
         TopicArn=arn,
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -390,7 +390,6 @@ receive any notifications until you click the confirmation link.""".format(email
         else:
             rootLogger.warning("Unknown exception is encountered while trying subscribe notifications")
         rootLogger.warning(err)
-    rootLogger.info("Proceeding..")
 
 
 def send_firesim_notification(subject, body):
@@ -409,11 +408,10 @@ def send_firesim_notification(subject, body):
         )
     except client.exceptions.ClientError as err:
         if 'AuthorizationError' in repr(err): 
-            rootLogger.warning("You don't have permissions to subscribe to firesim notifications")
+            rootLogger.warning("You don't have permissions to publish to firesim notifications")
         else:
             rootLogger.warning("Unknown exception is encountered while trying publish notifications")
         rootLogger.warning(err)
-    rootLogger.info("Proceeding..")
 
 
 if __name__ == '__main__':
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -358,42 +358,63 @@ def get_snsname_arn():
         )
     except client.exceptions.ClientError as err:
         if 'AuthorizationError' in repr(err): 
-            rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
-
+            rootLogger.warning("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
         else:
-            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
-
-        rootLogger.critical(err)
-        assert False
+            rootLogger.warning("Unknown exception is encountered while trying to perform \"Topic Creation\"")
+        rootLogger.warning(err)
+        return None
         
     return response['TopicArn']
 
 def subscribe_to_firesim_topic(email):
     """ Subscribe a user to their FireSim SNS topic for notifications. """
 
+    client = boto3.client('sns')
     arn = get_snsname_arn()
-    response = client.subscribe(
-        TopicArn=arn,
-        Protocol='email',
-        Endpoint=email
-    )
-
-    message = """You should receive a message at
-{}
+    if not arn: 
+        return None
+    try:
+        response = client.subscribe(
+            TopicArn=arn,
+            Protocol='email',
+            Endpoint=email
+        )
+        message = """You should receive a message at {}
 asking to confirm your subscription to FireSim SNS Notifications. You will not
 receive any notifications until you click the confirmation link.""".format(email)
 
-    rootLogger.info(message)
+        rootLogger.info(message)
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.warning("You don't have permissions to subscribe to firesim notifications")
+        else:
+            rootLogger.warning("Unknown exception is encountered while trying subscribe notifications")
+        rootLogger.warning(err)
+    rootLogger.info("Proceeding..")
+
 
 def send_firesim_notification(subject, body):
 
+    client = boto3.client('sns')
     arn = get_snsname_arn()
 
-    response = client.publish(
-        TopicArn=arn,
-        Message=body,
-        Subject=subject
-    )
+    if not arn: 
+        return None
+
+    try:
+        response = client.publish(
+            TopicArn=arn,
+            Message=body,
+            Subject=subject
+        )
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.warning("You don't have permissions to subscribe to firesim notifications")
+        else:
+            rootLogger.warning("Unknown exception is encountered while trying publish notifications")
+        rootLogger.warning(err)
+    rootLogger.info("Proceeding..")
+
 
 if __name__ == '__main__':
     #""" Example usage """
--- a/deploy/firesim
+++ b/deploy/firesim
@@ -119,6 +119,16 @@ def buildafi(globalbuildconf):
 
     auto_create_bucket(globalbuildconf.s3_bucketname)
 
+    client = boto3.client('sns')
+
+    aws_resource_names_dict = aws_resource_names()
+    snsname = aws_resource_names_dict['snsname']
+
+    # this will either create the topic, if it doesn't exist, or just get the arn
+    client.create_topic(
+        Name=snsname
+    )
+
     def terminate_instances_handler(sig, frame):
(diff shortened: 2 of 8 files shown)


timsnyder-siv:
@rahulb10 , you need to edit your PR so that it is going into `firesim:dev`, then you will see only one commit in the PR that you expect.

deploy/firesim
```diff
@@ -119,6 +119,16 @@ def buildafi(globalbuildconf):
...
+    client = boto3.client('sns')
+
+    aws_resource_names_dict = aws_resource_names()
+    snsname = aws_resource_names_dict['snsname']
+
+    # this will either create the topic, if it doesn't exist, or just get the arn
+    client.create_topic(
+        Name=snsname
+    )
+
```
timsnyder-siv, line 131: 1. You should capture this in a new function in https://github.com/sifive/firesim/blob/Topic_permission_check/deploy/awstools/awstools.py and refactor the other two uses to also call the new function:
   * https://github.com/sifive/firesim/blob/fafb94de6926350aebe7cd1f4bea63fc840312a3/deploy/awstools/awstools.py#L349-L358
   * https://github.com/sifive/firesim/blob/fafb94de6926350aebe7cd1f4bea63fc840312a3/deploy/awstools/awstools.py#L375-L385
   
2. You should put this in a try/except block with a helpful error message sent to `rootLogger.critical` before an empty `raise` to rethrow the exception.

timsnyder-siv (changes requested):
In addition to refactoring that code into a single function please make sure that your git name and email address is setup correctly so that your commits don't say that they were done by "Cloud User".

deploy/firesim
rahulb10: > refactor the other two uses to also call the new function
I wrapped the topic creation in a function.
> You should put this in a try/except block with a helpful error message
Created the same and to exit after the error.

deploy/firesim
```diff
@@ -118,16 +118,12 @@ def buildafi(globalbuildconf):
...
-    snsname = aws_resource_names_dict['snsname']
-
-    # this will either create the topic, if it doesn't exist, or just get the arn
-    client.create_topic(
-        Name=snsname
-    )
+    #check to see email notifications can be subscribed
+    try:
+        create_topic() 
+    except :
```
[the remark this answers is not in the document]
timsnyder-siv: Also, I think it would be better to only have the `client.create_topic` call inside of our new function in the `try` block. It will more precisely catch what we're trying to catch and also handle the permissions errors in other different uses of the new function that we may not think about today.

timsnyder-siv (changes requested):
Getting better.  There are still some things I'd like to address.

deploy/firesim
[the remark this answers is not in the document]
rahulb10: > Also, I think it would be better to only have the `client.create_topic` call inside of our new function in the `try` block.

I didn't get this. We have to anyway get the 'snsname' right to execute 'client.create_topic' ?

deploy/awstools/awstools.py
```diff
@@ -344,19 +344,32 @@ def auto_create_bucket(userbucketname):
...
+        response = client.create_topic(
+            Name=snsname
+        )
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
+            rootLogger.critical(err)
+        else:
+            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
+        assert False
```
timsnyder-siv, line 365:
```suggestion
            rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
        else:
            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
        rootLogger.critical(err)
        assert False
```

Ugh, github's rendering of this patch is really ugly, should just be moving a line down with an indent change.  Anyway, move the `rootLogger.critical(err)` to just before the `assert False` so that it is logged either way.

rahulb10: Done

timsnyder-siv:
I think this is yet another place where it would be really good to mock boto3 and add a unit test that checks for this new behavior of the new function `awstools.get_snsname_arn`.  Talking offline with @rahulb10 about this and one of us will add a test.

timsnyder-siv:
@rahulb10 I pushed a couple of tests.  Figuring out all of the magic to properly mock boto and also prevent it from being able to reach out to AWS while testing took a bit longer than I though it would.

deploy/tests/awstools/test_awstools.py
```diff
@@ -0,0 +1,79 @@
...
+# In case you put any package-level tests, make sure they use the test credentials too
+pytestmark = pytest.mark.usefixtures("aws_test_credentials")
+
+@pytest.mark.usefixtures("aws_test_credentials")
+class TestSNS(object):
+    """Test functions in awstools that work with SNS"""
+
+    # When we're testing awstools.get_snsname_arn(), we don't also want to be
+    # testing awstools.aws_resource_names(). mainly because we're subjugating
+    # the boto3.client factory and also because we want to be able to run these
```
timsnyder-siv, line 24:
```suggestion
    # the boto3.client factory (in the second test below) and also because we want to be able to run these
```

deploy/awstools/awstools.py
```diff
@@ -344,19 +344,34 @@ def auto_create_bucket(userbucketname):
...
+        )
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.critical("You don't have permissions to perform \"Topic Creation \". Required to send you email notifications. Please contact your IT administrator")
+
+        else:
+            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\"")
+
+        rootLogger.critical(err)
+        assert False
```
timsnyder-siv, line 367:
```suggestion
    except client.exceptions.ClientError as err:
        if 'AuthorizationError' in repr(err): 
            rootLogger.warning("Will not be able to send you email notifications.  You do not have AWS IAM permission  \"sns:CreateTopic\". Please contact your administrator")
            rootLogger.debug(err)

        else:
            rootLogger.critical("Unknown exception is encountered while trying to perform \"Topic Creation\". Please notify the FireSim team.")
            rootLogger.critical(err)
            assert False
```

@rahulb10 I realized that somehow we've steered away from the original intent here.  It definitely shouldn't be a fatal error when we're unable to create the topic and we know it was for AuthError.  That would prevent people from being able to build without the notifications.  It's debatable whether the unknown exception should `assert False` but maybe it should to encourage folks to let us know about other failures we could handle more helpfully.

I'm realizing that we probably need to update the docs to describe this and also update the code used in `firesim managerinit` where it tries to create the topic and subscribe to it.  Subscription is a separate permission and could fail differently.

deploy/awstools/awstools.py
```diff
@@ -344,19 +344,34 @@ def auto_create_bucket(userbucketname):
...
+        
+    return response['TopicArn']
+
+def subscribe_to_firesim_topic(email):
+    """ Subscribe a user to their FireSim SNS topic for notifications. """
 
+    arn = get_snsname_arn()
     response = client.subscribe(
         TopicArn=arn,
         Protocol='email',
```
timsnyder-siv, line 377: @rahulb10 this is the the call I think we should put into a another try/except block with similar structure to what we've done in `get_snsname_arn` except we will want to say they are missing "sns:Subscribe" permission.

Looking down here, if we make the `get_snsname_arn()` AuthError failure non fatal, we need to either reraise an exception that can be caught by the callers or return `None` so the callers don't make further AWS requests with a bogus arn.

deploy/awstools/awstools.py
timsnyder-siv: If we change the AuthError to be non-fatal (like I think we should), we'll also need to update the test I added becuase it expects the AuthError to be fatal.

deploy/firesim
```diff
@@ -119,6 +119,9 @@ def buildafi(globalbuildconf):
 
     auto_create_bucket(globalbuildconf.s3_bucketname)
 
+    #check to see email notifications can be subscribed
+    get_snsname_arn()
```
timsnyder-siv, line 123: This is a great improvement but we could still have the case where the user has `sns:CreateTopic` but they are missing the `sns:Publish` permission and thus still can't send the message.  It's different than the failure case @rahulb10 and I hit but it we should put try/catch blocks around the other AWS API calls

deploy/tests/awstools/test_awstools.py
```diff
@@ -0,0 +1,79 @@
...
+            with pytest.raises(AssertionError):
+                get_snsname_arn()
+
+            # TODO we could mock rootLogger.critical to capture it's calls and args and validate that we're seeing the correct "nice" message
+
+            # make sure get_snsname_arn() actually called out to get a sns
+            # client, otherwise we aren't testing what we think we are
+            mock_session.assert_called_once_with('sns')
+
+        aws_res_mock.assert_called_once()
```
timsnyder-siv, line 79: I dislike this pattern of using `@patch` as a decorator because it is unclear that the `@patch` decorator is providing the parameter, especially when mixed with other decorators and given that pytest provides fixtures as parameters.  

There is a `pytest-mock` plugin to pytest that thinly wraps `mock` so that it's usage is a little cleaner in pytest.  if we use that plugin, this code would look like this
```python
    @mock_sns
    def test_get_snsname_arn_auth_exception_handling(self, mocker):
        """AuthorizationError is intercepted and re-raised as AssertionError"""
        # local imports of code-under-test ensure moto has mocks
        # registered before any possible calls out to AWS
        from awstools.awstools import get_snsname_arn
        
        mocker.patch('awstools.awstools.aws_resource_names', return_value={'snsname': 'Testing'})

        # create a mock SNS client that returns what we tell it to
        client = boto3.client('sns')
        stub = Stubber(client)
        stub.add_client_error('create_topic', service_error_code='AuthorizationError')
        stub.activate()


        # since firesim manager code doesn't take clients as method parameters
        # now we mock boto3.client to return our stubbed client
        mocked_client_factory = patch.object(boto3._get_default_session(), 'client', return_value=client)
        with pytest.raises(AssertionError):
              get_snsname_arn()

        # TODO we could mock rootLogger.critical to capture it's calls and args and validate that we're seeing the correct "nice" message

        # make sure get_snsname_arn() actually called out to get a sns
        # client, otherwise we aren't testing what we think we are
        mocked_client_factory.assert_called_once_with('sns')

        awstools.awstools.aws_resource_names.assert_called_once()
```

Or, we could do one of the alternatives discussed in https://github.com/pytest-dev/pytest-mock/blob/3a4b13b9413e9c67b91e55b972698da1241e3ee2/README.rst#why-bother-with-a-plugin 

I do think that the pytest-mock plugin makes the test code cleaner.  I didn't add it because:
* I was trying to limit the number of libraries I added 
* I was lazy and didn't want to spend time figuring out which version would work with the version of mock that we have installed.

deploy/tests/awstools/test_awstools.py
```diff
@@ -0,0 +1,79 @@
...
+    @mock_sns
+    def test_get_snsname_arn_auth_exception_handling(self, aws_res_mock):
+        """AuthorizationError is intercepted and re-raised as AssertionError"""
+        # local imports of code-under-test ensure moto has mocks
+        # registered before any possible calls out to AWS
+        from awstools.awstools import get_snsname_arn
+
+        # create a mock SNS client that returns what we tell it to
+        client = boto3.client('sns')
+        stub = Stubber(client)
```
timsnyder-siv, line 62: It's probably useful to add a comment here that says something like 
```
# we use botocore.stub.Stubber() here instead of mock.patch because boto dynamically
# creates classes at runtime for it's service-specific client objects.  mock.patch can only patch
# static classes that exist in the python symbol table.  botocore.stub.Stubber() is explicitly created for this
# see: https://github.com/boto/boto3/issues/2485
```

deploy/awstools/awstools.py
rahulb10: I have to still retain `assert False` in get_snsname_arn() if at all I use try/except block in `subscribe_to_firesim_topic()` . In that case for it to remain non-fatal I have to use try/except block in all the functions that call `get_snsname_arn()` including the check we are having at the beginning of buildafi.

deploy/firesim
timsnyder-siv: My concern here is addressed by the changes in b25cd16

deploy/awstools/awstools.py
```diff
@@ -344,51 +344,77 @@ def auto_create_bucket(userbucketname):
...
-    arn = response['TopicArn']
+    try:
+        response = client.publish(
+            TopicArn=arn,
+            Message=body,
+            Subject=subject
+        )
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.warning("You don't have permissions to subscribe to firesim notifications")
```
timsnyder-siv, line 412:
```suggestion
            rootLogger.warning("You don't have permissions to publish to firesim notifications")
```
You should say publish here instead of subscribe.

deploy/awstools/awstools.py
```diff
@@ -344,51 +344,77 @@ def auto_create_bucket(userbucketname):
...
+        else:
+            rootLogger.warning("Unknown exception is encountered while trying publish notifications")
+        rootLogger.warning(err)
+    rootLogger.info("Proceeding..")
```
timsnyder-siv, line 416:
```suggestion
```
I would remove this logging statement.  It may have been helpful while you were developing but I don't think it adds information during normal usage.

deploy/awstools/awstools.py
```diff
@@ -344,51 +344,77 @@ def auto_create_bucket(userbucketname):
...
 
-    rootLogger.info(message)
+        rootLogger.info(message)
+    except client.exceptions.ClientError as err:
+        if 'AuthorizationError' in repr(err): 
+            rootLogger.warning("You don't have permissions to subscribe to firesim notifications")
+        else:
+            rootLogger.warning("Unknown exception is encountered while trying subscribe notifications")
+        rootLogger.warning(err)
+    rootLogger.info("Proceeding..")
```
timsnyder-siv, line 393:
```suggestion
```
I would remove this logging statement.  It may have been helpful while you were developing but I don't think it adds information during normal usage.

deploy/awstools/awstools.py
```diff
@@ -344,51 +344,77 @@ def auto_create_bucket(userbucketname):
...
+            TopicArn=arn,
+            Protocol='email',
+            Endpoint=email
+        )
+        message = """You should receive a message at {}
 asking to confirm your subscription to FireSim SNS Notifications. You will not
 receive any notifications until you click the confirmation link.""".format(email)
 
-    rootLogger.info(message)
+        rootLogger.info(message)
```
timsnyder-siv, line 386: printing this message should move into an `else:` clause after the `except:` so that it is only printed if an exception is not raised when trying to do `client.subscribe()`

timsnyder-siv:
As we're adding all of these error handling try/except clauses around calls to `boto_client.<action>`, they look very similar and a future work might be good to either:
* create our own context-manager class that we can use with a simple `with HandleBotoClientErrors():` block around the action calls instead of having to repeat the except clauses many different places.
* create our own wrapped boto client class that we could put the error handling code in one place. The wrapper class might be more flexible.  Then when we want to handle errors in a standard way, we could do something like `client = ExceptionHandler(boto3.client(<service>))`.  

Like I said though, let's leave this for future work unless the BAR folks disagree.

deploy/awstools/awstools.py
timsnyder-siv: This is also resolved in b25cd16

deploy/awstools/awstools.py
rahulb10: If the message is in `try`:, it would only get printed if an exception is not raised right?

timsnyder-siv: Yes.  You're correct.  Since it is after the point where an exception would be raised it's correct the way it is.  The only advantage of moving it might be for a little clarity and so that there's no way it could raise an exception that would be made non-fatal.  I don't think that's a big enough deal to move it.

deploy/tests/awstools/test_awstools.py
```diff
@@ -67,8 +67,9 @@ def test_get_snsname_arn_auth_exception_handling(self, aws_res_mock):
         # since firesim manager code doesn't take clients as method parameters
         # now we mock boto3.client to return our stubbed client
         with patch.object(boto3._get_default_session(), 'client', return_value=client) as mock_session:
-            with pytest.raises(AssertionError):
-                get_snsname_arn()
+            get_snsname_arn()
```
timsnyder-siv, line 70: You should probably also assert that the return value is `None`.

timsnyder-siv: addressed in 2a29eb8

deploy/tests/awstools/test_awstools.py
```diff
@@ -0,0 +1,79 @@
...
+        stub = Stubber(client)
+        stub.add_client_error('create_topic', service_error_code='AuthorizationError')
+        stub.activate()
+
+
+        # since firesim manager code doesn't take clients as method parameters
+        # now we mock boto3.client to return our stubbed client
+        with patch.object(boto3._get_default_session(), 'client', return_value=client) as mock_session:
+            with pytest.raises(AssertionError):
+                get_snsname_arn()
```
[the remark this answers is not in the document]
timsnyder-siv: addressed in c24b334

abejgonzalez (approved):
LGTM. Though, I think it would be nice to add the tests to the CI before this is merged in.

timsnyder-siv:
> it would be nice to add the tests to the CI before this is merged in.
@davidbiancolin said he'd take care of adding them to the CI.  David, I'll leave it up to you whether you want to add them in this PR and merge them passing in CI or add them in a follow-on.

davidbiancolin:
Yeah i'll add them now.

davidbiancolin:
Don't forget to a changelog entry in the dev->master PR.

timsnyder-siv:
@rahulb10 unless you have more tests to add now for the changes we made to the other methods in awstools, I think we should go ahead to "squash and merge" this PR.

PR #759, opened 2021-04-27
[DO NOT MERGE] Local FireSim Mockup

abejgonzalez:
This is an initial sketch on how to run `buildafi` locally and have it do specific logic to one FPGA. This hopefully will transition into a later Local FireSim full PR.

Things to think about / ask:
- Why not just use `local` for the `replace_rtl` step? Why use `execute` with `run`?
- Right now I'm using `managerinit` + `--localrun` to only copy files over and not do any other AWS stuff... better way to do this?
- Clean up the class hierarchy / abstract more things away.
- Ctrl-C when running the `replace_rtl` step doesn't cancel the RTL build.

### DIFF
--- a/deploy/buildtools/buildafi.py
+++ b/deploy/buildtools/buildafi.py
@@ -6,8 +6,6 @@ import string
 import logging
 import os
 
-from fabric.api import *
-from fabric.contrib.console import confirm
 from fabric.contrib.project import rsync_project
 from awstools.afitools import *
 from awstools.awstools import send_firesim_notification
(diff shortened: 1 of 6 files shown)


abejgonzalez:
Want from Sagar, have local build support on AWS so that you don't have to spawn a buildfarm instance.

timsnyder-siv:
> Want from Sagar, have local build support on AWS so that you don't have to spawn a buildfarm instance.

How would you expect it to behave if you have multiple configurations?  The easy thing to do right now is run them all in parallel on the localhost.  Maybe it's also easy to make them sequential but it would take me more looking into how people have `@parallel` and `@sequential` variants of the same task with Fabric.

deploy/sample-backup-configs/sample_config_build.ini
```diff
@@ -1,6 +1,21 @@
 # BUILDTIME/AGFI management configuration for the FireSim Simulation Manager
 # See docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst for documentation of all of these params.
 
+[buildfarmconfig]
+# build platform (options are "aws" or "local")
+hostplatform=aws
```
timsnyder-siv, line 6: What if I want to run the `replace-rtl` and `aws-fpga` Vivdo build flow locally and then submit my checkpoint to S3 for an AMI to be created?  I've been asked whether this is possible.  We should think through what that would look like so that choices we make in the config file here don't end up being really confusing if/when that is added.

deploy/buildtools/buildglobal.py
```diff
@@ -0,0 +1,105 @@
...
+        self.hwdb = RuntimeHWDB(args.hwdbconfigfile)
+
+        # setup the particular host_platform
+        self.host_platform = global_build_configfile.get('buildfarmconfig', 'hostplatform')
+        if self.host_platform == "aws":
+            self.host_platform_config = GlobalAwsBuildConfig(args, global_build_configfile)
+        elif self.host_platform == "local":
+            self.host_platform_config = GlobalLocalBuildConfig(args, global_build_configfile, self.builds_list)
+        else:
+            sys.exit('Invalid host platform: {}'.format(host_platform))
```
timsnyder-siv, line 58: #763 would help this type of code too.  If we merge it, uses of sys.exit like this would get logged.  Right now, they wouldn't.

timsnyder-siv (commented):
@abejgonzalez , I rebased #651 forward today and was reminded that it shouldn't break if we need to switch to using `fabric.local()` instead of `fabric.execute()` for `replace_rtl`.

I have to run but I"m going to go ahead and post my initial thoughts on your code.

deploy/buildtools/buildlocal.py
```diff
@@ -0,0 +1,80 @@
...
+import logging
+import os
+import sys
+
+from fabric.api import *
+from util.streamlogger import StreamLogger, InfoStreamLogger
+
+rootLogger = logging.getLogger()
+
+class GlobalLocalBuildConfig:
```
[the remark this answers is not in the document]
abejgonzalez: Hahaha. I completely agree with the naming. I just quickly mocked this up. I think your names are definitely better.

deploy/firesim
```diff
@@ -51,19 +51,20 @@ def managerinit():
...
-    while not valid_creds:
-        # only run aws configure if we cannot already find valid creds
-        # this loops calling valid_aws_configure_creds until
-        rootLogger.info("Running aws configure. You must specify your AWS account info here to use the FireSim Manager.")
-        # DO NOT wrap this local call with StreamLogger, we don't want creds to get
-        # stored in the log
-        local("aws configure")
-
-        # check again
+    if not args.localrun:
```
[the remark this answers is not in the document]
abejgonzalez: I agree. I think that is cleaner, I just didn't want to introduce any new APIs (though... I slightly broke that assumption with the `--localrun` flag).

abejgonzalez:
> > Want from Sagar, have local build support on AWS so that you don't have to spawn a buildfarm instance.
> 
> How would you expect it to behave if you have multiple configurations? The easy thing to do right now is run them all in parallel on the localhost. Maybe it's also easy to make them sequential but it would take me more looking into how people have `@parallel` and `@sequential` variants of the same task with Fabric.

I haven't thought in-depth on this... Sagar complained about it to me offline so I wanted it to be documented somewhere. My assumption if that things are done locally things should be done sequentially (IIRC Vivado uses 8-ish threads... so I don't thing running 8+ builds in parallel is recommended).

deploy/sample-backup-configs/sample_config_build.ini
timsnyder-siv: > > > Want from Sagar, have local build support on AWS so that you don't have to spawn a buildfarm instance.
> > 
> > 
> > How would you expect it to behave if you have multiple configurations? The easy thing to do right now is run them all in parallel on the localhost. Maybe it's also easy to make them sequential but it would take me more looking into how people have `@parallel` and `@sequential` variants of the same task with Fabric.
> 
> I haven't thought in-depth on this... Sagar complained about it to me offline so I wanted it to be documented somewhere. My assumption if that things are done locally things should be done sequentially (IIRC Vivado uses 8-ish threads... so I don't thing running 8+ builds in parallel is recommended).

I made my comment before reading your code and I think we should maybe move this thread of conversation here so that the main unthreaded conversation in this PR doesn't become a ton of quoted replies that keep getting longer and longer... 

After reading through this part of the code and thinking about this more, it seems like we have two dimensions here:
1. buildhosts - aws, local (maybe there will need to be different kinds of local)
    * this could be as simple as aws provisions the buildcluster for you while local just uses a list of hostnames you provide
    * in either case, if you have fewer hostnames than configs you want to build, we could start by failing like you currently do. 
2. fpgaplatform - AWS.f1, other boards, is this not already provided by `PLATFORM_CONFIG` in the individual `build_config_recipes.ini`?  Would it be better if we just provided a mapping from values of scala `PLATFORM_CONFIG` to the task implementation class that needs to be used in python?  Perhaps longer term, the `PLATFORM_CONFIG` should be factored out into a JSON or YAML or something that both the Scala and Python could refer to for a single source of truth about the platform?  Or maybe the Scala PlatformConfig _generates_ Python or YAML or something that the firesim manager reads out of 

The existing firesim manager knows how to do `buildhosts=aws fpgaplatform=f1`.  In this PR, you're trying to teach it to do `buildhosts=local fpgaplatform=xilinx-u250`.  Sagar complained about wanting to be able to do `buildhosts=local fpgaplatform=f1`.

If we view it like this, the config file might be better like:
```
[buildhosts]
user@host1.mydatacenter
user@host2.mydatacenter
david@davidslaptop.net #because he wants to run my builds
somecustomthing
aws

[hostproviders]
aws=ProvisionEC2BuildFarm  #ProvisionEC2BuildFarm does the spinup and teardown of hosts that exists now
somecustomthing=NonBARPluginClass 
```


The way this would work is that the buildhosts is an ordered list that gets zipped with [builds] to create the mapping of  buildhost to build.  However, there's one fancy thing about iterating the [buildhosts], if you hit the name of a buildhost that is defined in [hostproviders], the provided class name acts like a factory that provisions a host for you and returns the IP address you use to access it.  I'm envisioning that we might also want to have the hostproviders be albe to return something that says "sorry bruh, can't give you any more hosts".  This would be useful in the case where you have a local cluster that can provide some number of hosts but then you have the aws hostprovider at the end and it can burst up to whatever your limit for EC2 hosts is (but even then it should probably learn that it's unable to provision more hosts and quit).  

Perhaps what I just described is that we want to zip [builds] with a flatMap of buildhosts where the hostproviders themselves are iterable.

I can see how this same idea could be applied at runtime (provided that we change things a little in config_hwdb.ini so that we could specify the target we want to simulate and the firesim manager learns how to pick the appropriate PLATFORM_CONFIG based on the host it is given to run on), maybe PLATFORM_CONFIG becomes a dict of configs to bitstream identifiers or something.

These are just thoughts.

davidbiancolin:
> > Want from Sagar, have local build support on AWS so that you don't have to spawn a buildfarm instance.
> 
> How would you expect it to behave if you have multiple configurations? The easy thing to do right now is run them all in parallel on the localhost. Maybe it's also easy to make them sequential but it would take me more looking into how people have `@parallel` and `@sequential` variants of the same task with Fabric.

FWIW, the stupid intermediate step of allowing only a single build run on localhost per buildafi invocation seems worthwhile.

deploy/sample-backup-configs/sample_config_build.ini
davidbiancolin: > What if I want to run the `replace-rtl` and `aws-fpga` Vivdo build flow locally and then submit my checkpoint to S3 for an AMI to be created? I've been asked whether this is possible. We should think through what that would look like so that choices we make in the config file here don't end up being really confusing if/when that is added.

I think this should just work if you have the AWS CLI set up? At the very least, it shouldn't be too hard.

davidbiancolin: I like your orthogonalization of concerns. We should move it to a separate issue though. Run & BuildFarm provisioning does seem similar, we could probably pull the hostproviders stuff + fixed hosts in a single `host_config.ini`. If we factored it nicely it would be pretty easy to write a slurm / qsub module that could dynamically provide hosts as we do in AWS.

timsnyder-siv: > We should move it to a separate issue though.
Sounds good to me.

deploy/buildtools/buildafi.py
```diff
@@ -6,252 +6,296 @@
...
+            rootLogger.debug(rsync_cap.stderr)
+            rsync_cap = rsync_project(local_dir=ddir + "/../platforms/f1/aws-fpga/{}/*".format(fpgabuilddir),
+                        remote_dir='/home/centos/firesim-build/platforms/f1/aws-fpga/' + remotefpgabuilddir,
+                        exclude='build/checkpoints',
+                        ssh_opts="-o StrictHostKeyChecking=no",
+                        extra_opts="-l", capture=True)
+            rootLogger.debug(rsync_cap)
+            rootLogger.debug(rsync_cap.stderr)
+
+        # run the Vivado build
```
abejgonzalez, line 156: Move this into some script that is called at this point.

abejgonzalez: Bash script. Args (host fpga freq, build strategy)

abejgonzalez: config_hwdb.ini

```
[name]
platform = blah
schema
```

abejgonzalez: figure out how to get a hwdb from any platform from local/remote

abejgonzalez: how to do the build vs how to allocate the machines to do the builds

PR #768, opened 2021-05-18
[simif] Hide SEED output unless explicitly set

davidbiancolin:
The SEED is not used anywhere that will affect the target, so mask it's output until we start using it / reimplement it in a safer way. simif_t::rand_next is used in a few places, so i'll remove it and simif_t::gen() when i factor out the peek-poke widget from simif as well.

### DIFF
--- a/sim/midas/src/main/cc/simif.cc
+++ b/sim/midas/src/main/cc/simif.cc
@@ -18,7 +18,6 @@ simif_t::simif_t() {
   pass = true;
   t = 0;
   fail_t = 0;
-  seed = time(NULL); // FIXME: better initail seed?
   SIMULATIONMASTER_0_substruct_create;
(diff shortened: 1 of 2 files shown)


timsnyder-siv:
Is there an urgency associated with removing the printout of SEED?  It seems like a good idea to just do this along with however you're planning to remove the seed and `simif_t::gen()`.  It gives me heartburn to just hide the fact that there is a random number generator being used somehow in the driver that I don't see documented anywhere. 

Finally, I think it's important to call out that the real change done by this PR is "use 0 to seed simif_t::gen() unless otherwise specificd by the cmdline option `+seed=`" when you put something into the dev pr for this.

davidbiancolin:
Yeah, that makes sense.  I'll jsut fold this into the next PR.

davidbiancolin:
This will be subsumed by a future PR.

PR #769, opened 2021-05-26
Remove Coremark and SPEC2017 from deploy/workloads

davidbiancolin:
... as support for these is provided through FireMarshal and workload definitions in Chipyard.


NathanTP (approved):
sounds good to me!

PR #776, opened 2021-06-03
[TracerV] Support Heterogenous Mixes of Cores 

### LINKED ISSUE #729: TracerVBridge issue with HeteroConfig
When I tried to build a FPGA image with HeteroConfig, I faced TracerV issue that "ports of extmodule TracerVBridge with defname TracerVBridge are different for an extmodule with the same defname"
For now, I commented out TracerVBridge at BridgeBinder.scala, and it was built. (Though it wouldn't have autocounter features)
image

davidbiancolin:
Previously, FIRRTL CheckHighForm would croak because different tracerV target black boxes would have different port definitions. This works around that problem by giving a parameterized name to each black box based on the widths of it's IO. 

Providing widths as parameters in the black box does not suffice to fix this problem, because the _length_ of the Vec of traced instruction can vary from bridge to bridge. 

Also adds some scaladoc. 

Resolves #729.

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -30,11 +30,19 @@ class TracerVTargetIO(insnWidths: TracedInstructionWidths, numInsns: Int) extend
   val triggerCredit = Output(Bool())
   val triggerDebit = Output(Bool())
 }
-
-// Warning: If you're not going to use the companion object to instantiate this
-// bridge you must call generate trigger annotations _in the parent module_.
-//
-// TODO: Generalize a mechanism to promote annotations from extracted bridges...
+/**
+  * Target-side module for the TracerV Bridge.
+  *
+  * @param insnWidths A case class containing the widths of configurable-length
+  * fields in the trace interface.
+  *
+  * @param numInsns The number of instructions captured in a single a cycle
+  * (generally, the commit width of the pipeline)
+  *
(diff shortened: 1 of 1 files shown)


sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -45,6 +53,20 @@ class TracerVBridge(insnWidths: TracedInstructionWidths, numInsns: Int) extends
...
+  // bridge's instruction count as a string suffix. This ensures that TracerV
+  // blackboxes with different instruction counts will have different defnames,
+  // preventing FIRRTL CheckHighForm failure when using a chipyard "Hetero"
+  // config. While a black box parameter relaxes the check on leaf field
+  // widths, CheckHighForm does not permit parameterizations of the length of a
+  // Vec enclosing those fields (as is the case here), since the Vec is lost in
+  // a lowered verilog module.
+  //
+  // See https://github.com/firesim/firesim/issues/729.
+  val defnameSuffix = s"_${numInsns}Wide_" + insnWidths.toString.replaceAll("[(),]", "_")
```
jackkoenig, line 67:
```suggestion
  def defnameSuffix = s"_${numInsns}Wide_" + insnWidths.toString.replaceAll("[(),]", "_")
```
Doesn't really need to be a `val`, not that the extra memory use really matters in prctice.

alonamid (approved):
LGTM in terms of functionality (l'll defer to Jack in terms of Scala cleanliness)

PR #782, opened 2021-06-08
Introduce a ResetPulseBridge to Remove An Early Deadlock Condition

davidbiancolin:
Currently, reset in the target is implemented using the peek-poke widget, via a blocking call to `simif_t::target_reset`. Since other bridges are not being ticked while this function executes, the simulator can deadlock near time zero if any bridge stalls waiting on an invocation of `tick` (e.g., a synthesized assertion fires). 

This PR introduces a dedicated bridge that can produce a runtime-configurable reset pulse to drive the target. This will subsume target_reset(), allowing the simulator to enter it's main loop immediately upon simulation commence. 

Before i can remove target_reset from firesim_top, i'll need to instantiate the bridge in Chipyard. I will open separate PRs to do that after the release has been handled, and this PR has been merged. 

This is based off the hostIF PR (#778); i'll retarget to dev once it is merged.

### DIFF
--- a/sim/midas/src/main/cc/bridges/reset_pulse.cc
+++ b/sim/midas/src/main/cc/bridges/reset_pulse.cc
@@ -1,0 +1,39 @@
+#ifdef RESETPULSEBRIDGEMODULE_struct_guard
+
+#include "reset_pulse.h"
+
+reset_pulse_t::reset_pulse_t(
+  simif_t* sim,
+  std::vector<std::string> &args,
+  RESETPULSEBRIDGEMODULE_struct * mmio_addrs,
+  unsigned int max_pulse_length,
+  unsigned int default_pulse_length,
+  int reset_index):
+    bridge_driver_t(sim),
+    mmio_addrs(mmio_addrs),
+    max_pulse_length(max_pulse_length),
+    default_pulse_length(default_pulse_length),
+    reset_index(reset_index) {
+
+  std::string num_equals = std::to_string(reset_index) + std::string("=");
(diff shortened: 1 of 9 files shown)


sim/midas/src/main/cc/bridges/reset_pulse.cc
```diff
@@ -0,0 +1,39 @@
+#ifdef RESETPULSEBRIDGEMODULE_struct_guard
```
timsnyder-siv, line 1: What is this `RESETPULSEBRIDGEMODULE_struct_guard` macro and how does it get defined?  I'm not seeing that in this pr, is it just letting you have this code in your branch but not use it in the driver right now?  Is this supposed to be taken out before merging?

davidbiancolin: This is a necessary (evil) component of defining a bridge since one of the constructor arguments is defined for the first time in the generated header. If no bridges are instantiated, the struct is not defined, and this code cannot compile...

timsnyder-siv (approved):

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -564,9 +568,6 @@ void firesim_top_t::run() {
     fprintf(stderr, "Commencing simulation.\n");
     record_start_times();
 
-    // Assert reset T=0 -> 50
-    target_reset(50);
-
```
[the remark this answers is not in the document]
davidbiancolin: I'm going to hold off on removing these two lines until we've bumped CY.

PR #791, opened 2021-06-17
Adds a Global Reset Condition to Mask Off Events During Reset

davidbiancolin:
In more complex reset schemes, leaf resets may be pipelined.  As a result, assertions, printfs, and autocounters in these reset domains may start counting at time zero --- before these regions of the target have been reset --- leading to spurious (really, undesired) assertions. 

IN RC + CY we work around this problem by using a global `PRINTF_COND` and `RESET_COND`, which point at a bool that will be asserted over the region in which local resets may not be. This PR essentially synthesizes that same logic, by wiring an annotated Bool, the `GlobalResetCondition`, to every instance of an instrumentation bridge _in the target_. 

Note, this violates the rule that sinks outputs must be driven by registers in their clock domain, which breaks assumptions made by PromotePassthroughPaths. The hub-model can simulate the paths fine, however.  To work around this, i introduce a utility `PassthroughBlocker` which can be used to prevent the PromotePassthroughPaths from hoisting out these multi-clock-sink paththroughs. In a separate PR i will add an analysis to better check that channels cannot fanout to multiple clock domains. 

#### Related PRs / Issues
Builds off #778 #782. 

#### UI / API Impact

Added `GlobalResetCondition` to targetutils. Assertions / Printfs / AutoCoutner will have different semantics when this is deployed. 

No material change until Chipyard is bumped to use this feature. 

### Contributor Checklist

- [x] (If applicable) Did you mark the PR as "Please Backport"?


sim/midas/src/main/scala/midas/passes/AutoCounterTransform.scala
```diff
@@ -166,32 +166,39 @@ class AutoCounterTransform extends Transform with AutoCounterConsts {
...
         DefNode(NoInfo, triggerPortNode, one),
-        Connect(NoInfo, WRef(triggerPortName), WRef(triggerPortNode)))
-      val triggerPortRT  = topMT.ref(triggerPortName)
-      val triggerFcca = FAMEChannelConnectionAnnotation.source(
-        triggerPortName,
-        WireChannel,
-        Some(sinkClockRT),
-        Seq(triggerPortRT))
+        Connect(NoInfo, WRef(triggerPortName), WRef(triggerPortNode)),
+        Connect(NoInfo, WRef(globalResetName), zero),
```
alonamid, line 183:
```suggestion
        // In the event no GlobalResetCondition is provided, tie this off
        Connect(NoInfo, WRef(globalResetName), zero),
```

alonamid (approved):
Looks OK as far as I can tell.
The only part I'm not sure I was able to find was where do the global reset signal get applied/sourced, and whether there is an option to enable or disable it.
Also, it would be nice to add a line in the debugging sections of the docs mentioning what would be the default semantic of the instrumentation features (with reset or without reset).

PR #800, opened 2021-06-25
Add XDC Emission Support + Multicycle Setup constraints

davidbiancolin:
This PR implements  does two things:
1) It adds runtime generation of XDC files. Which we'll use in implementation.
2) It implements an optimization I wrote about in my dissertation, wherein target clocks that do not fire in back to back host cycles are given a multi-cycle setup constraint equal to some lower bound > 1. This can improve f_max by up to 50% in some designs. 

On the XDC generation:

I think there's lots of room to make this more powerful. There's a comment explaining the limitations. 

On the MFMR bound:

This lower bound can be derived statically from the clock token stream. Here I do the stupid thing, and divide every slower clock by the period of the fastest, as there will always be at least one clock token firing the faster clock between tokens firing the slower one. This establishes a minimum FMR (MFMR) of ceil(slow_period / fast_period) in the slower clock domain.  We could do better still by counting the minimum number of clock tokens between active edges in each domain. When there are many clocks that are not divisions of the clock, this may substantially improve the MFMR bound. 

Other random thoughts:
- We emit a ton of different files now. We probably what to harmonize how emission is done:
   - use a common prefix for all generated files
   - clean up replace_rtl 
   - specify prefix on CML
   - document all output files in docs somewhere 

Additional TODOS;
- [ ] Use FIRRTLFile Emission trait
- [ ] Add tests.

### Contributor Checklist

### DIFF
--- a/deploy/sample-backup-configs/sample_config_hwdb.ini
+++ b/deploy/sample-backup-configs/sample_config_hwdb.ini
@@ -10,27 +10,27 @@
 # own images.
 
 [firesim-boom-singlecore-nic-l2-llc4mb-ddr3]
-agfi=agfi-03ed6ca96115e8928
+agfi=agfi-020f55d5f8ff00dd4
 deploytripletoverride=None
 customruntimeconfig=None
 
 [firesim-boom-singlecore-no-nic-l2-llc4mb-ddr3]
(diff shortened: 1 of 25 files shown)


davidbiancolin:
Subsumed by #825, #834

PR #804, opened 2021-07-07
Install Boost in the Machine Launch Script 

davidbiancolin:
It's not actually clear to me where we are using it.

#### Related PRs / Issues

None

#### UI / API Impact

Users will need to install boost on existing managers when upgrading, _if_ it's actually needed. 

None

### Contributor Checklist
- [] Did you add at least one test demonstrating the PR?

- [x] (If applicable) Did you regenerate and publicly share default AGFIs?

### DIFF
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -79,7 +79,7 @@ boost_minor_v=69
 wget https://sourceforge.net/projects/boost/files/boost/1.${boost_minor_v}.0/boost_1_${boost_minor_v}_0.tar.bz2
 tar --bzip2 -xf boost_1_${boost_minor_v}_0.tar.bz2
 cd boost_1_${boost_minor_v}_0
-sudo yum install bzip2-devel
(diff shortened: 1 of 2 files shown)


timsnyder-siv:
> It's not actually clear to me where we are using it.

When I look, it isn't obvious to me where it's being used either.  Can we remove it from the include paths in the build and see what breaks in CI?  Are you doing builds of the driver in CI now?

timsnyder-siv:
I realized that I was also meaning that we'd remove it from the makefrag but not install it in the default include paths so that we can get compiler errors when we hit usages.  We definitely also needed to remove the path that we're not installing inside the clone too.  

I'm also remembering that I need to spin up a new manager with the new ami to be able to test this.  le sigh.  looking.

timsnyder-siv:
Since #805 showed that Boost isn't currently used by anything in FireSim, can we just remove it completely?  @sagark ?

davidbiancolin:
Subsumed by #806

PR #811, opened 2021-07-13
Remove deadcode in midas.passes.Utils

davidbiancolin:
Here i remove some legacy strober utilities that either:
- are unused
- have first-class equivalents in FIRRTL
- or are subsumed by utilities in midas.passes.fame

#### Related PRs / Issues

#807 

#### UI / API Impact

I'm removing public APIs, but these were never intended to be used outside of midas.

#### Verilog / AGFI Compatibility

None

### Contributor Checklist
- [x] Did you set dev as the base branch?
- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous prints/debugging code?
- [x] Did you state the UI / API impact?
- [x] Did you specify the Verilog / AGFI compatibility impact?

- [x] (If applicable) Did you regenerate and publicly share default AGFIs?

- [ ] (If applicable) Did you mark the PR as "Please Backport"?
- [ ] (On merge) Did you update release notes in the dev-to-master PR ?

### Reviewer Checklist (only modified by reviewer)
- [x] Did you mark the proper release milestone?
- [x] Did you check whether all relevant Contributor checkboxes have been checked?

Issue #807, opened 2021-07-10
Remove Known Dead Code 


davidbiancolin:
I'm keeping a list here for future clean up.
* midasexamples PointerChaser
* Many utilities in Lib
* Legacy strober utilities 
* Legacy FAME1 transform; use gated clock conversion instead.

PR #812, opened 2021-07-13
Remove Unused Chisel Utilities + GeneratorUtils + PlusArgReader Pass

davidbiancolin:
- Plus arg reader removal was replaced with simply passing `SYNTHESIS a few releases ago, but we didn't delete the pass then.
- Config lookup code was simply duplicated into the phase that does the same thing.
- Rocket's GeneratorUtils have been replaced with the Stage. 
- Also delete misc. Chisel utilities once used by FASED 
#### Related PRs / Issues

#807 

#### UI / API Impact

As in #811, these the Chisel utilities are public, but not likely used outside of FASED. 

### Contributor Checklist

- [x] (If applicable) Did you regenerate and publicly share default AGFIs?


alonamid (approved):
Looks ok to me. If I understand this correctly, much of this code in the past assumed the FireSim would operate independently as a top level generator on arbitrary chisel designs (with it's own generator, etc.), but the current removal of this code is since we are now assuming the target design always has a generator associated with it?

davidbiancolin:
> Looks ok to me. If I understand this correctly, much of this code in the past assumed the FireSim would operate independently as a top level generator on arbitrary chisel designs (with it's own generator, etc.), but the current removal of this code is since we are now assuming the target design always has a generator associated with it?

That's right.  Specifically, all of the generator makefrags call out to Chipyard's Stage now.

PR #826, opened 2021-09-10
Limit builddir name length to 255 chars

jerryz123:
#### Related PRs / Issues

Will avoid some errors when user-requested build config string is very long

#### UI / API Impact

No impact.

No effect

### Contributor Checklist

### DIFF
--- a/deploy/buildtools/buildconfig.py
+++ b/deploy/buildtools/buildconfig.py
@@ -66,8 +66,7 @@ class BuildConfig:
 
     def get_build_dir_name(self):
         """" Get the name of the local build directory. """
-        return """{}-{}-{}""".format(self.launch_time,
-                                     self.get_chisel_triplet(), self.name)
+        return """{}-{}""".format(self.launch_time, self.name)
 
     # Builds up a string for a make invocation using the tuple variables
     def make_recipe(self, recipe):


deploy/buildtools/buildconfig.py
```diff
@@ -66,8 +66,7 @@ def terminate_build_instance(self):
 
     def get_build_dir_name(self):
         """" Get the name of the local build directory. """
-        return """{}-{}-{}""".format(self.launch_time,
-                                     self.get_chisel_triplet(), self.name)
+        return """{}-{}""".format(self.launch_time, self.name)[:255]
```
davidbiancolin, line 69: I don't think truncation is the play here. I'd prefer a warning. 

What exactly is struggling with the long filenames?

davidbiancolin: If we do decide that these directories must be kept within 255 characters we should just throw an error on recipe names that are too long when they are pulled in. 

Also the directories under sim/generated-src + sim/output will suffer of this file length problem.

jerryz123: Max filename size for most filesystems is 255 bytes.
I'll remove the truncation. Its actually pretty unlikely this will cause further errors with the builddir name if the chisel-triplet is no longer included.

The directories under sim/generated-src and sim/output will suffer the file length problem only if the chisel-triplet itself is > 255 chars.

davidbiancolin: > The directories under sim/generated-src and sim/output will suffer the file length problem only if the chisel-triplet itself is > 255 chars.

Yeah i guess what i'm saying is that is all but assured if we continue using string concatenation + non-abbreviated class names.

davidbiancolin:
@jerryz123, could you remove the truncation and rebase? CI should be more stable.

davidbiancolin (approved):
LGTM

PR #825, opened 2021-09-08
Add Support for XDC Emission that is Hierarchy-Mutation Robust. 

davidbiancolin:
This subsumes one half of #800. The multicycle constraints are to follow in a separate PR.

In short, this PR permits adding XDC snippets that refer to chisel/firrtl nodes in target-generator code or in FireSim sources, such that their references to hardware will be resolved to correct absolute paths after Golden Gate runs. This makes the XDC generation more robust against changes in module hierarchy (you don't have to rely on regexp or hierarchal matches) and changes in node names which could be introduced by the compiler. 

#### Related PRs / Issues

#800

#### UI / API Impact

Expands targetutils. 

#### Verilog / AGFI Compatibility

No. 

### Contributor Checklist
- [x] Did you set dev as the base branch?
- [ ] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous prints/debugging code?
- [x] Did you state the UI / API impact?
- [x] Did you specify the Verilog / AGFI compatibility impact?

- [ ] (If applicable) Did you regenerate and publicly share default AGFIs?

- [ ] (If applicable) Did you mark the PR as "Please Backport"?
- [ ] (On merge) Did you update release notes in the dev-to-master PR ?

### Reviewer Checklist (only modified by reviewer)
- [x] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #834, opened 2021-09-15
Emit multi-cycle clock constraints on target domains

davidbiancolin:
Using the XDC emission utilities, now emit the multi-cycle constraints. Quoting the previous PR:

This implements an optimization I wrote about in my dissertation, wherein target clocks that do not fire in back to back host cycles are given a multi-cycle setup constraint equal to some lower bound > 1. This can improve f_max by up to 50% in some designs.

This lower bound can be derived statically from the clock token stream. Here I do the stupid thing, and divide every slower clock by the period of the fastest, as there will always be at least one clock token firing the faster clock between tokens firing the slower one. This establishes a minimum FMR (MFMR) of ceil(slow_period / fast_period) in the slower clock domain. We could do better still by counting the minimum number of clock tokens between active edges in each domain. When there are many clocks that are not divisions of the clock, this may substantially improve the MFMR bound.

#### Related PRs / Issues

#800 #825

#### UI / API Impact

None.

Could push the timing on the default targets.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
+++ b/sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
@@ -15,6 +15,8 @@ import scala.collection.mutable
 import mutable.{LinkedHashSet, LinkedHashMap}
 
 import midas.passes._
+import midas.targetutils.xdc.{XDCFiles, XDCAnnotation}
+import midas.widgets.{RationalClock}
 
 /**************
  PRECONDITIONS:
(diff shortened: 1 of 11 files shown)


sim/midas/src/main/scala/midas/passes/fame/FAMETransform.scala
```diff
@@ -178,6 +191,44 @@ object FAMEModuleTransformer {
...
+        // clk 3  : 1 0 1 0 1 0
+        // clk 4  : 1 0 0 1 0 1
+        // t_time : 0 2 3 4 6 8
+        //
+        // While the latter two clocks have MFMRs of 2, they sometimes fire in
+        // back-to-back host-cycles (tokens 2->3).
+        //
+        // Note: "host_clock" is defined statically in the shell XDC.
+        val xdcAnno = XDCAnnotation(
+          XDCFiles.Implementation,
```
timsnyder-siv, line 223: Does this mean that only the Implementation XDC will have these constraints?  If so, I'm concerned that the retiming that happens during CL synthesis will continue to spend resources retiming paths in the composible cache that don't need to be retimed and then not have resources available for retiming.  Why wouldn't you use 
```suggestion
          XDCFiles.allFiles,
```

davidbiancolin: So, this is a good question. 

I can put them in synthesis, and they should carry though. Or yeah, i could duplicate them. 

I picked just implementation because I had some trouble getting the synthesis constraints to apply correctly in synthesis, and most of the benefit is materialized in implementation anyways -- frankly, i was feeling lazy. 

I'll try again now.

timsnyder-siv: Ok.  I wasn't sure whether applying them in synthesis would mean they would come through the checkpoint and linking to be included in implementation.  I'll pick this up and also try it out.

davidbiancolin: Yeah, you can choose the behavior you want.

PR #842, opened 2021-10-07
Expose `awstools.py` cmdline to launch/terminate instances

abejgonzalez:
Changes:
- Exposes `awstools.py`'s `launch_instances` function to a command line.
- Instead of doing instance liveness using `uname -a` (in `firesim`) to check if an instance is ready to ssh/execute commands on, use the `instance_status_ok` waiter (previously used in CI) under the hood when an instance is launched. This removes the liveness check and has it bundled with the `wait_on_instance_launches` function. (Small downside is that this serializes the check (in `firesim`)... but shouldn't be too much of an issue if the instances are of the same type)

Example Help:
image

#### Related PRs / Issues

N/A

#### UI / API Impact

This doesn't affect any current FireSim APIs. This just adds an extra one if using `awstool.py` as the main function.

N/A

### Contributor Checklist

### DIFF
--- a/.circleci/cull-old-ci-instances.py
+++ b/.circleci/cull-old-ci-instances.py
@@ -7,7 +7,12 @@ import datetime
 import pytz
 import boto3
 
-from common import get_instances_with_filter, unique_tag_key
+from common import unique_tag_key
+
+# Reuse manager utilities
+from ci_variables import ci_workdir
+sys.path.append(ci_workdir + "/deploy/awstools")
+from awstools import get_instances_with_filter
 
 # The number of hours an instance may exist since its initial launch time
 INSTANCE_LIFETIME_LIMIT_HOURS = 24


abejgonzalez:
@davidbiancolin I don't think CI needs to change because within `launch-manager-instance.py` it directly calls the functions within `awstools.py`. Did you want a specific change?

davidbiancolin:
the changes themselves seem fine. 

What i'm asking is that you throw out this entire script (here https://github.com/firesim/firesim/blob/dev/.circleci/launch-manager-instance.py) to instead have CI directly call out to the aws-tools main. Do you expose enough knobs to make this possible?

abejgonzalez:
> the changes themselves seem fine.
> 
> What i'm asking is that you throw out this entire script (here https://github.com/firesim/firesim/blob/dev/.circleci/launch-manager-instance.py) to instead have CI directly call out to the aws-tools main. Do you expose enough knobs to make this possible?

Not exactly... I still need to expose things like `blockdevices`. I'll get to it eventually.

abejgonzalez:
> > the changes themselves seem fine.
> > What i'm asking is that you throw out this entire script (here https://github.com/firesim/firesim/blob/dev/.circleci/launch-manager-instance.py) to instead have CI directly call out to the aws-tools main. Do you expose enough knobs to make this possible?
> 
> Not exactly... I still need to expose things like `blockdevices`. I'll get to it eventually.

This should be addressed now.

PR #864, opened 2021-10-28
Factor peek/poke out of simif_t

davidbiancolin:
I'd like to get to a point where we can build most of the driver source as a single library -- this as a stepping stone towards building a single driver. My plan is check in mmio address struct definitions alongside their bridges so that we can remove the macro guards, and eventually stop including the generated header in all compilation units.

This biggest blocker of this are derived simif classes that use the peek-poke interface, since:
- the mmio struct definition changes from target to target
- there's a lot of code that uses this IF 

We've long wanted to remove peek-poke from simif. However, peek poke is still useful for testing (but not for user's deployment) so here i pull out all peek-poke related methods out of simif into a new derived class used only in midasexamples/fasedtests. In a future PR i will set about removing the requirement that peek-poke bridge must be used in the target. 

#### UI / API Impact

Further slims down the god class `simif_t`. 
- Changes the semantics of finish to avoid diamond-pattern challenges with multiple inheritance
  - This method should be defined by class implementing the host abstraction layer (simif_f1/simif_emul). 

Most people don't write custom derived classes of simif_t that use peek poke unless they are writing tests. 

No change.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/simif.cc
+++ b/sim/midas/src/main/cc/simif.cc
@@ -24,9 +24,10 @@ simif_t::simif_t() {
   this->clock_bridge_mmio_addrs = CLOCKBRIDGEMODULE_0_substruct;
 }
 
-void simif_t::init(int argc, char** argv, bool log) {
+void simif_t::init(int argc, char** argv) {
+  // Do any post-constructor initialization required before requesting MMIO
+  this->host_init(argc, argv);
   while(!read(this->master_mmio_addrs->INIT_DONE));
-  this->log = log;
   std::vector<std::string> args(argv + 1, argv + argc);
   std::string loadmem;
   bool fastloadmem = false;
--- a/sim/midas/src/main/cc/simif.cc
+++ b/sim/midas/src/main/cc/simif.cc
@@ -15,16 +15,11 @@ double diff_secs(midas_time_t end, midas_time_t start) {
 }
 
 simif_t::simif_t() {
-  pass = true;
-  t = 0;
-  fail_t = 0;
   seed = time(NULL); // FIXME: better initail seed?
   SIMULATIONMASTER_0_substruct_create;
   this->master_mmio_addrs = SIMULATIONMASTER_0_substruct;
   LOADMEMWIDGET_0_substruct_create;
   this->loadmem_mmio_addrs = LOADMEMWIDGET_0_substruct;
-  PEEKPOKEBRIDGEMODULE_0_substruct_create;
-  this->defaultiowidget_mmio_addrs = PEEKPOKEBRIDGEMODULE_0_substruct;
   CLOCKBRIDGEMODULE_0_substruct_create;
   this->clock_bridge_mmio_addrs = CLOCKBRIDGEMODULE_0_substruct;
 }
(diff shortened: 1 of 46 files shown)


sim/midas/src/main/cc/simif.cc
```diff
@@ -15,23 +15,19 @@ double diff_secs(midas_time_t end, midas_time_t start) {
...
   CLOCKBRIDGEMODULE_0_substruct_create;
   this->clock_bridge_mmio_addrs = CLOCKBRIDGEMODULE_0_substruct;
 }
 
-void simif_t::init(int argc, char** argv, bool log) {
+void simif_t::init(int argc, char** argv) {
+  // Do any post-constructor initialization required before requesting MMIO
+  this->host_init(argc, argv);
   while(!read(this->master_mmio_addrs->INIT_DONE));
-  this->log = log;
```
abejgonzalez, line 34: How does this `log` boolean get enabled? This seems to be only used in the `simif_peek_poke` classes (at least from the diff)... but its a `private` member and thus this `log` isn't used?

sim/src/main/cc/fasedtests/fasedtests_top.cc
```diff
@@ -111,11 +115,7 @@ void fasedtests_top_t::run() {
         zero_out_dram();
     }
     fprintf(stderr, "Commencing simulation.\n");
-    uint64_t start_hcycle = hcycle();
-    uint64_t start_time = timestamp();
-
-    // Assert reset T=0 -> 50
```
abejgonzalez, line 117: This isn't needed because of the reset pulse bridge right?

sim/src/main/cc/fasedtests/test_harness_bridge.cc
```diff
@@ -24,6 +24,10 @@ test_harness_bridge_t::test_harness_bridge_t(
 // it then reads uarch event counts from the FASED instance and compares them against
 // expected values
 void test_harness_bridge_t::tick(){
+  // Wait for reset to complete.
+  if (sim->actual_tcycle() < 100)
```
abejgonzalez, line 28: Why is this needed?

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -570,7 +570,7 @@ void firesim_top_t::run() {
 
     while (!simulation_complete() && !has_timed_out()) {
         run_scheduled_tasks();
-        step(get_largest_stepsize(), false);
+        take_steps(get_largest_stepsize(), false);
```
abejgonzalez, line 573: So it looks like `step` was deleted and thus you replaced it with this. Can you elaborate a bit on what `step` did? I think my confusion is around what `t` means/meant (I know it moved to the `peek_poke` class).

abejgonzalez (commented):
Most of this LGTM. I would just like to understand a couple points before the full stamp.

sim/src/main/cc/fasedtests/test_harness_bridge.cc
davidbiancolin: There's no longer a blocking call to target_reset. This little hack does the job for these tests.

sim/src/main/cc/firesim/firesim_top.cc
davidbiancolin: t was essentially a cycle count, vis-a-vis steps taken by peek poke. If you used non-blocking peeks / pokes (rare) this would be pretty non-sensical. 

What `step` and it's underlying implementation `take_steps` does is write (indirectly) to the peek-poke widget some number of credits. When the credits are exhausted it stops producing tokens and the simulator eventually stalls. So "steps" here is in terms of cycles in whatever clock domain the peek-poke widget is in. 
We should probably change how time-control is handled to make more sense in a multi-clock, peek-poke-less simulator (in my dissertation branch the hub model will not advance past a certain time thats been plumbed out to MMIO), and rename this.

abejgonzalez (approved):
SGTM

PR #870, opened 2021-11-05
 Option to Compress Workload Results

### LINKED ISSUE #869: Feature Request?: Compress Workload Results on Completion
In the case, that simulation outputs are very large, have the manager (or `config_runtime.ini`) have an option to store workload results (found in `results_workload`) in a compressed format. Then it would be up to the user to unzip the results when ready to analyze. This is useful for when many simulations are run in parallel.

If done as a manager option... `manager runworkload --compressed`?
If done as a `config_runtime.ini` option place under `[workload]` with something like `compress=yes` but default to `compress=no`?

**Example Use Case:**
Imagine running N simulations where each simulation copies back things like the TracerV log and AutoCounter logs. In some cases, those logs are extremely large (depending on the length of the workload). Therefore, you can quickly fill up disk space on the manager instance. Instead, bundle up simulation results into a zip or gz file (or any equivalent) and copy back that file.

abejgonzalez:
#### Related PRs / Issues

Fixes #869 (closes #869)

#### UI / API Impact

Adds new `compressresults=` option to `config_runtime.ini`'s `[workload]` section to compress outputs coming from simulations into the `results-workload` directory (the compressed file is stored per job and is named after the job). The default behavior is still to not compress workload results. This is using `tar` to create the archive.

N/a

### Contributor Checklist


deploy/runtools/firesim_topology_elements.py
```diff
@@ -293,27 +288,66 @@ def copy_back_job_results_from_run(self, slotno):
                     run("""sudo chattr -i {}/etc/sysconfig/nfs""".format(mountpoint))
                 run("""sudo chown -R centos {}""".format(mountpoint))
 
-            ## copy back files from inside the rootfs
-            with warn_only(), StreamLogger('stdout'), StreamLogger('stderr'):
+        if not self.get_job().parent_workload.compress_results:
```
davidbiancolin, line 291: Please reduce the code duplication across both sides of the conditional. Perhaps it makes sense to always compress before transporting. Alternatively, build up a filelist first before either compressing or iterating through them with gets, and then do the common clean up tasks.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -573,8 +573,8 @@ def loop_logger(instancestates, terminateoncompletion):
 
             def get_jobs_completed_local_info():
                 # this is a list of jobs completed, since any completed job will have
-                # a directory within this directory.
-                jobscompleted = os.listdir(self.workload.job_results_dir)
+                # a directory (or .tar.gz file) within this directory.
+                jobscompleted = [os.path.splitext(p)[0] for p in os.listdir(self.workload.job_results_dir)]
```
abejgonzalez, line 577:
```suggestion
                jobscompleted = [p.split(".")[0] for p in os.listdir(self.workload.job_results_dir)]
```

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -234,6 +234,13 @@ without renaming the entire workload. For example, specifying
 in a workload results directory named
 ``results-workload/DATE--TIME-super-application-test-v1/``.
 
+``compressresults``
+""""""""""""""""""""""""""
+
+Set this to ``yes`` if you want workload results to be compressed when copied back
```
a0u, line 240: Consider making `compressresults` instead accept a compression algorithm as an argument (e.g.,  `compressresults=gzip`), which leaves open the option of supporting alternatives such as lz4 and zstd in the future.

Also, I'm curious about how the compression ratio compares to on-the-fly filesystem compression with zfs or btrfs, as opposed to ad-hoc support in FireSim.

abejgonzalez:
Closing since the preferred method is to compress files with a new volume + filesystem (like ZFS). A future PR will add this tip.

PR #876, opened 2021-11-17
Implement PerfCounter.identity + CSV Output Format

davidbiancolin:
This PR has three major components. 

1) Implements PerfCounter.identity()
-  Multiple users have requested the ability to pass out a value through AutoCounter (i.e., avoid accumulation) that is sampled alongside events that are traditionally accumulated. This permits having user defined functions in the target to do more complex operations beyond accumulate, and can serve as means to capture additional information about the target in the autocounter output files (avoiding the need to use a printf or a secondary sideband).

2) Changes the AC output format to CSV. The CSV contains all of the metadata required to process the file. The format is given here, but will be put in the docs. [See this sketch.](https://docs.google.com/spreadsheets/d/16VbVYnN742MLtK7WGNqyxmCp4eWlVmVto2pegooo0qo/edit?usp=sharing)

3) Overhauls the AC test utilities. The old tests tediously mirrored the output format with brittle printfs. This introduces a system to track PerfCounters as they are instantiated, and pushes the printf validation to the end of elaboration. Printf generation has been DRYed out such that updating the format does not require changing literally every test module.
- Warts: instance path naming could be better. 

#### UI / API Impact

- Perfcounter object has new identity method. 
- Output format changed to CSV. This is less human readable, but much easier to parse. 

No

### Contributor Checklist

### DIFF
--- a/docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
+++ b/docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
@@ -16,28 +16,45 @@ simulation (unlike target-level performance counters), these counters do not
 affect the behavior of the simulated machine, no matter how often they are
 sampled.
 
-Ad-hoc Performance Counters
-------------------------------
+Chisel Interface
+----------------
 
 AutoCounter enables the addition of ad-hoc counters using the ``PerfCounter``
-function.  The ``PerfCounter`` function takes 3 arguments: A UInt signal encoding an event increment, a counter label, and the counter description. The increment indicates the number of times the event of occurs in a single cycle and will be added on every simulated cycle to the generated performance counter. An increment can be boolean, with true implying the event occured (once) in the cycle, or multibit, to capture events that occur more than once in a cycle. We show two examples
-of using the ``PerfCounter`` function below.
+object in the `midas.targetutils` package. PerfCounters counters can be added in one of two modes:
+
+# `Accumulate`, using the standard ``PerfCounter.apply`` method. Here the annotated UInt (1 or
+  more bits) is added to a 64b accumulation register: the target is treated as
+  representing an N-bit UInt and will increment the counter by a value between [0, 2^n - 1] per cycle.
+# `Identity`, using the ``PerfCounter.identity`` method. Here the annotated UInt is sampled directly. This can be used
+  to annotate a sample with values are not accumulator-like (e.g., a PC),
+  and permits the user to define more complex instrumentation logic in the target itself.
+
+We give examples of using PerfCounter below:
 
 .. code-block:: scala
 
+    // A standard boolean event. Increments by 1 or 0 every local clock cycle.
     midas.targetutils.PerfCounter(en_clock, "gate_clock", "Core clock gated")
-    // A multibit example
+
+    // A multibit example. If the core can retire three isntructions per cycle,
+    // encode this as a two-bit unit. Extra-width is OK but the encoding to the UInt
+    // (e.g., doing a pop count), must be done by the user.
     midas.targetutils.PerfCounter(insns_ret, "iret", "Instructions retired")
 
+    // An identity value. Note: the pc here must be <= 64b wide.
+    midas.targetutils.PerfCounter.identity(pc, "pc", "The value of the program counter at the time of a sample")
 
 
-Building a Design with AutoCounter
+See the `PerfCounter Scala API docs
+<https://fires.im/firesim/latest/api/midas/targetutils/PerfCounter$.html>`_ for more detail about the Chisel-side interface.
+
+
+Enabling AutoCounter in Golden Gate
 -------------------------------------
-
-To enable AutoCounter when building a design, prepend the ``WithAutoCounter``
-config to your ``PLATFORM_CONFIG``. During compilation, FireSim will print the
-signals it is generating counters for. If AutoCounter has been enabled, the
-``autocounter_t`` bridge driver will also be automatically instantiated.
+By default, annotated events are not synthesized into AutoCounters.  To enable
+AutoCounter when compiling a design, prepend the ``WithAutoCounter`` config to
+your ``PLATFORM_CONFIG``. During compilation, Golden Gate will print the
+signals it is generating counters for.
 
 
 Rocket Chip Cover Functions
@@ -95,18 +112,41 @@ By default, the read-rate is set to 0 cycles, which disables AutoCounter.
    readrate=100
 
 
-Upon setting this value, when you run a workload, an AutoCounter output file
-will be placed in the ``sim_slot_<slot #>`` directory on the F1 instance under
-the name ``AUTOCOUNTERFILE<N>``, with one file generated per clock domain
-containing an AutoCounter event. The header of each output file indicates the
-associated clock domain and its frequency relative to the base clock.
-
 .. Note:: AutoCounter is designed as a coarse-grained observability mechanism, as sampling 
       each counter requires two (blocking) MMIO reads (each read takes O(100) ns on EC2 F1).
       As a result sampling at intervals less than O(10000) cycles may adversely affect
       simulation performance for large numbers of counters.
-      If you intend on reading counters at a finer granularity, please consider using
+      If you intend on reading counters at a finer granularity, consider using
       synthesizable printfs.
+
+AutoCounter CSV Output Format
+---------------------------------
+AutoCounter output files are CSVs generated in the working directory where the
+simulator was invoked (this applies to metasimulators too), with the default
+names ``AUTOCOUNTERFILE<i>.csv``, one per clock domain. The CSV output format is
+depicted below, assuming a sampling period of ``N`` base clock cycles.
+
+.. csv-table:: AutoCounter CSV Format
+    :file: autocounter-csv-format.csv
+
+Column Notes:
+
+#. Each column beyond the first two corresponds to a PerfCounter instance in the clock domain.
+#. Column 0 past the header corresponds to the base clock cycle of the sample.
+#. The local_cycle counter (column 1) is implemented as an always enabled
+   single-bit event, and increments even when the target is under reset.
+
+Row Notes:
+
+#. Header row 0: autocounter csv format version, an integer.
+#. Header row 1: clock domain information.
+#. Header row 2: the label parameter provided to PerfCounter suffixed with the instance path.
+#. Header row 3: the description parameter provided to PerfCounter. Quoted.
+#. Header row 4: the width of the field annotated in the target.
+#. Header row 5: the width of the accumulation register. Not configurable, but makes it clear when to expect rollover.
+#. Header row 6: indicates the accumulation scheme. Can be "Identity" or "Accumulate".
+#. Sample row 0: sampled values at the bitwidth of the accumulation register.
+#. Sample row k: ditto above, k * N base cycles later
 
 Using TracerV Trigger with AutoCounter
 -----------------------------------------
@@ -117,7 +157,7 @@ triggers. See the :ref:`tracerv-trigger` section for more information.
 
 AutoCounter using Synthesizable Printfs
 ------------------------------------------------
-The AutoCounter transformation in the Golden Gate compiler includes an event-driven
+The AutoCounter transformation in Golden Gate includes an event-driven
 mode that uses Synthesizable Printfs (see
 :ref:`printf-synthesis`) to export counter results `as they are updated` rather than sampling them
 periodically with a dedicated Bridge. This mode can be enabled by prepending the
@@ -130,3 +170,15 @@ printed to the same output stream as other synthesizable printfs.  This mode
 uses considerably more FPGA resources per counter, and may consume considerable
 amounts of DMA bandwidth (since it prints every cycle a counter
 increments), which may adversly affect simulation performance (increased FMR).
+
+Reset & Timing Considerations
+------------------------------
+* Events and identity values provided while under local reset, or while the
+  ``GlobalResetCondition``  asserted, are zero-ed out. Similarly, printfs that
+  might otherwise be active under a reset are masked out.
+* The sampling period in slower clock domains is currently calculated using a truncating
+  division of the period in the base clock domain. Thus, when the base clock
+  period can not be cleanly divided, samples in the slower clock domain will
+  gradually fall out of phase with samples in the base clock domain. In all
+  cases, the "local_cycle" column is most accurate measure of sample time.
+
--- a/docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
+++ b/docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
@@ -162,14 +162,21 @@ mode that uses Synthesizable Printfs (see
 :ref:`printf-synthesis`) to export counter results `as they are updated` rather than sampling them
 periodically with a dedicated Bridge. This mode can be enabled by prepending the
 ``WithAutoCounterCoverPrintf`` config to your ``PLATFORM_CONFIG`` instead of
-``WithAutoCounterCover``. In this mode, the counter values and the local cycle count will be printed
-every time the counter is incremented using a synthesized printf (hence, you
-will observe a series of printfs incrementing by 1). This mode may
-be useful for fine-grained observation of counters.  The counter values will be
-printed to the same output stream as other synthesizable printfs.  This mode
-uses considerably more FPGA resources per counter, and may consume considerable
-amounts of DMA bandwidth (since it prints every cycle a counter
-increments), which may adversly affect simulation performance (increased FMR).
+``WithAutoCounterCover``. Based on the selected event mode the printfs will have the following runtime behavior:
+
+* `Accumulate`: On a non-zero increment, the local cycle count and the new
+  counter value are printed. This produces a series of prints with
+  monotonically increasingly values.
+* `Identity`: On a transition of the annotated target, the local cycle count and
+  the new value are printed. Thus a target that transitions every cycle will
+  produce printf traffic every cycle.
+
+This mode may be useful for temporally fine-grained observation of counters.
+The counter values will be printed to the same output stream as other
+synthesizable printfs.  This mode uses considerably more FPGA resources per
+counter, and may consume considerable amounts of DMA bandwidth (since it prints
+every cycle a counter increments), which may adversly affect simulation
+performance (increased FMR).
 
 Reset & Timing Considerations
 ------------------------------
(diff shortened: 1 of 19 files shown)


abejgonzalez:
So both the `identity` AutoCounters and the normal accumulated ones would exist in the same file? Then its up to the person parsing the AutoCounter file to separate out accumulated vs identity counters? Is there a nice way to have these be separate files? Or maybe have some way to identify which "counters" are accumulating vs non (though I think we talked about a potential solution to this which is to have a "summary file" and a data file (that is easier to parse)).

davidbiancolin:
Can we just change the default output to be CSV like we talked about, and then stick metadata on successive rows? That was one of the proposals.

abejgonzalez:
Overall, the changes seem good to me. I just want to see the docs updated on this. Also I think that the `message` field is probably better named as `description`. Finally, the accumulator width is really only 64b right now right? In the future you plan to make this configurable?

davidbiancolin:
> Overall, the changes seem good to me. I just want to see the docs updated on this. Also I think that the `message` field is probably better named as `description`. 
- I agree, i can change it though i think it might've been borrowed from cover? 

> Finally, the accumulator width is really only 64b right now right? In the future you plan to make this configurable?
- Yes. When needed though.

davidbiancolin:
Also to your message point, i'd like to decide on what the various names mean for like PerfCounter / AutoCounter / Event. So i can clean up the names throughout the code.

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -16,28 +16,38 @@ simulation (unlike target-level performance counters), these counters do not
...
+# `Identity`, using``PerfCounter.identity``. Here the user provided is sampled directly. This can be used
+  to annotate a sample with values are not accumulator-like (e.g., a PC),
+  and permits the user to define more complex instrumentation logic in the target itself.
+
+We give examples of using PerfCounter below:
 
 .. code-block:: scala
 
     midas.targetutils.PerfCounter(en_clock, "gate_clock", "Core clock gated")
     // A multibit example
```
abejgonzalez, line 35: Nit: In the past I got confused by what it means to add a multi-bit signal (I had to dig into the xform for the solution). For example, if `insns_ret == b0100` did it mean that `b0100` interpreted as an `UInt` would be added to the accumulation register or would a "popcount"ed version be added (i.e. `b0100` converts to "add 1 to accumulation register"). Maybe some clarity on this would be nice (maybe just stating this example in the comment of the multibit example).

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -95,19 +105,41 @@ By default, the read-rate is set to 0 cycles, which disables AutoCounter.
...
       simulation performance for large numbers of counters.
-      If you intend on reading counters at a finer granularity, please consider using
+      If you intend on reading counters at a finer granularity, consider using
       synthesizable printfs.
 
+AutoCounter CSV Output Format
+---------------------------------
+AutoCounter output files are CSVs generated in the working directory where the
+simulator was invoked (this applies to metasimulators too), with the default
+names ``AUTOCOUNTERFILE<N>``, one per clock domain. The CSV output format is
```
abejgonzalez, line 119: Shouldn't the name have `.csv` attached at the end?

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -95,19 +105,41 @@ By default, the read-rate is set to 0 cycles, which disables AutoCounter.
...
+depicted below, assuming a sampling period of ``N`` base clock cycles.
+
+.. csv-table:: AutoCounter CSV Format
+    :file: autocounter-csv-format.csv
```
abejgonzalez, line 123: Did you forget to upload this file?

sim/midas/src/main/scala/midas/widgets/AutoCounterBridge.scala
```diff
@@ -129,6 +162,14 @@ class AutoCounterBridgeModule(
       crRegistry.genHeader(headerWidgetName, base, sb, lowCountAddrs ++ highCountAddrs)
       crRegistry.genArrayHeader(headerWidgetName, base, sb)
       emitClockDomainInfo(headerWidgetName, sb)
+      sb.append(genConstStatic(s"${headerWidgetName}_event_count",  UInt32(allEventMetadata.size)))
+      sb.append(genArray(s"${headerWidgetName}_event_types",  allEventMetadata.map { m => CStrLit(m.opType.toString) }))
+      sb.append(genArray(s"${headerWidgetName}_event_labels", allEventMetadata.map { m => CStrLit(m.label) } ))
+      sb.append(genArray(s"${headerWidgetName}_event_messages", allEventMetadata.map { m => CStrLit(m.message) } ))
```
abejgonzalez, line 168: Didn't we settle on `description` instead of `message(s)`?

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
davidbiancolin: I can append a csv, but I wasn't sure that was what people wanted. 

I'd be down if we renamed this plus arg: https://github.com/firesim/firesim/blob/master/sim/midas/src/main/cc/bridges/autocounter.cc#L54
to +autocounter-base-filename?

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -16,28 +16,45 @@ simulation (unlike target-level performance counters), these counters do not
...
+We give examples of using PerfCounter below:
 
 .. code-block:: scala
 
+    // A standard boolean event.
     midas.targetutils.PerfCounter(en_clock, "gate_clock", "Core clock gated")
-    // A multibit example
+
+    // A multibit example. If the core can retire three isntructions per cycle,
+    // encode this as a two-bit unit. (Extra-width is OK but encoding to the UInt
```
abejgonzalez, line 40:
```suggestion
    // encode this as a two-bit unit. Extra-width is OK but encoding to the UInt
```

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
abejgonzalez: I think appending `.csv` makes it extremely clear how to read the file and makes it nice if people start opening the file in other tools. I'm fine with renaming the plusarg.

abejgonzalez (approved):
I think this looks good now. Small nits but otherwise great!

abejgonzalez (approved):
LGTM. Ship it! On a side note, I wonder if there is a grammar RST bot that can give PR comments whenever docs are changed...

sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -122,6 +123,41 @@ abstract class TutorialSuite(
     }
   }
 
+  /**
+    * Compares an AutoCounter output CSV against a reference generated using in-circuit printfs.
+    */
+  def checkAutoCounterCSV(filename: String, stdoutPrefix: String) {
+    it should s"produce a csv file (${filename}) with the correct content" in {
```
timsnyder-siv, line 130:
```suggestion
    it should s"produce a csv file (${filename}) that matches in-circuit printf output" in {
```
don't hide what you mean by correctness.

timsnyder-siv (approved):
This looks good enough to merge.  I do have some questions and a few small typo fixups.  I suppose the biggest question I have is "What happens if someone puts a comma in one of their description strings?"...

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -95,19 +112,41 @@ By default, the read-rate is set to 0 cycles, which disables AutoCounter.
...
-      If you intend on reading counters at a finer granularity, please consider using
+      If you intend on reading counters at a finer granularity, consider using
       synthesizable printfs.
 
+AutoCounter CSV Output Format
+---------------------------------
+AutoCounter output files are CSVs generated in the working directory where the
+simulator was invoked (this applies to metasimulators too), with the default
+names ``AUTOCOUNTERFILE<N>.csv``, one per clock domain. The CSV output format is
+depicted below, assuming a sampling period of ``N`` base clock cycles.
```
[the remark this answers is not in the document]
davidbiancolin: I'll use a different symbol; it's not related to the sample rate.

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
```diff
@@ -95,19 +112,41 @@ By default, the read-rate is set to 0 cycles, which disables AutoCounter.
...
+    :file: autocounter-csv-format.csv
+
+Column Notes:
+
+#. Each column beyond the first two corresponds to a PerfCounter instance in the clock domain.
+#. Column 0 past the header corresponds to the base clock cycle of the sample.
+#. The local cycle counter (column 1) is implemented as an always enabled
+   single-bit event, and increments even when the target is under reset.
+
+Row Notes:
```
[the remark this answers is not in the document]
davidbiancolin: Perhaps a better call is to instead version the csv, and put a version string in the first line? I don't mind putting the header length at the top.

sim/midas/src/main/cc/bridges/autocounter.cc
```diff
@@ -19,86 +19,117 @@ autocounter_t::autocounter_t(
...
     AUTOCOUNTERBRIDGEMODULE_struct * mmio_addrs,
     AddressMap addr_map,
+    const uint32_t event_count,
+    const char* const* event_types,
+    const uint32_t* event_widths,
+    const uint32_t* accumulator_widths,
+    const uint32_t* event_addr_hi,
+    const uint32_t* event_addr_lo,
+    const char* const* event_msgs,
+    const char* const* event_labels,
```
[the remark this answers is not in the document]
davidbiancolin: Emission from Golden Gate still is heavily C oriented, and i rather not write new serialization machinery if we're going to switch to a completely new protobuf-based format soon. In the past i've struggled to include c++ constructs in the header, i think because it's being included in C compilation units with a global `-include`. 

I do think this should be changed though -- I can convert these arrays to c++ containers when the bridge driver is constructed for now.

docs/Advanced-Usage/Debugging-and-Profiling-on-FPGA/AutoCounter.rst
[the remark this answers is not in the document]
timsnyder-siv: Yeah.  Versioning it is maybe a good idea but it seemed a bit heavyweight.  However, if you keep good changelog for the format between versions it will make it easier to handle reading them later.

sim/midas/targetutils/src/main/scala/midas/annotations.scala
```diff
@@ -220,31 +256,44 @@ object PerfCounter {
     *
     * @param label A verilog-friendly identifier for the event signal
     *
-    * @param message A description of the event.
+    * @param description A description of the event.
```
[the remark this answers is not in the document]
davidbiancolin: Yeah, I was being lazy here, I'll quote the description strings.

sim/src/main/scala/midasexamples/AutoCounterValidator.scala
```diff
@@ -0,0 +1,222 @@
...
+    domainName: String = "BaseClock",
+    printfPrefix: String = "AUTOCOUNTER_PRINT ",
+    autoCounterPrintfMode: Boolean = false,
+    clockDivision: Int = 1,
+    clockMultiplication: Int = 1,
+    samplePeriodBase: Int = 1000
+  ) extends InstanceNameHelper with AutoCounterConsts {
+
+  // Auto Counter currently uses the top module name as the final substring in
+  // the label path Se this top-module name upon instantiation, since we must
```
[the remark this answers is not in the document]
davidbiancolin: Will fix.

sim/src/main/scala/midasexamples/AutoCounterValidator.scala
```diff
@@ -0,0 +1,222 @@
...
+      print_header_row("accumulator width", {_ => counterWidth.toString } );
+    }
+
+    when ((localCycles >= (samplePeriod - 1).U) && (localCycles % samplePeriod.U === 0.U)) {
+      prefixed_printf(Seq.fill(counters.size)("%d").mkString(","), counters:_*)
+      sampleCount := sampleCount + 1.U
+    }
+  }
+
+  // TODO: define what an identity AC should do under this mode.
```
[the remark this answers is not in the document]
davidbiancolin: The latter makes sense.

sim/midas/targetutils/src/main/scala/midas/annotations.scala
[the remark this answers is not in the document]
timsnyder-siv: I think it is fine to be lazy and blow up if someone wants to have a comma in their description string... it would just be annoying if you allow commas to be there and not be quoted.

sim/midas/src/main/cc/bridges/autocounter.cc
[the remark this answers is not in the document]
timsnyder-siv: Thanks for the explanation.  Whatever you want to do that doesn't create a bunch of work, knowing that this will be resolved by the header reworking is helpful.

sim/src/main/scala/midasexamples/AutoCounterValidator.scala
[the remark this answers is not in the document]
davidbiancolin: Implemented.

sim/midas/src/main/cc/bridges/autocounter.cc
```diff
@@ -92,25 +93,49 @@ void autocounter_t::init() {
...
+    size_t start_pos = 0;
+    while((start_pos = str.find(from, start_pos)) != std::string::npos) {
+        str.replace(start_pos, from.length(), to);
+        start_pos += to.length();
+    }
+    return str;
+}
+
+std::string quote_csv_element(std::string str) {
+    std::string quoted = replace_all(str, "\"", "\"\"");
```
timsnyder-siv, line 106: Is this really what you want?  Is it going to replace double quotes in the description with two sets of double quotes?  Should it be adding an escape backslash before the double-quote in the description?

sim/src/main/scala/midasexamples/AutoCounterModule.scala
```diff
@@ -36,7 +36,7 @@ class AutoCounterModuleDUT(val instName: String = "dut")(implicit val v: AutoCou
   // Multibit event
   val count = RegInit(0.U(4.W))
   count := count + 1.U
-  PerfCounter(count, "MULTIBIT_EVENT", "A multibit event")
+  PerfCounter(count, "MULTIBIT_EVENT", "A multibit event. Here's a quote: \" and a comma: , to check description serialization.")
```
timsnyder-siv, line 40: Did you just visually inspect the CSV for correctness here?  How is this test checked, seems different than checking the count.

sim/src/main/scala/midasexamples/AutoCounterValidator.scala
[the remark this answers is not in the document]
timsnyder-siv: I think you still need to document the behavior in the RST, right?

sim/midas/src/main/cc/bridges/autocounter.cc
timsnyder-siv: In other words, I'm asking if you really mean 
```suggestion
    std::string quoted = replace_all(str, "\"", "\\\"");
```

I'm going to go try to run your Scala test for AutoCounterModule and look at the CSV that is generated to see if it makes sense.

This junk is why I was alright with you blowing up if someone passes a description that contains a comma.  It's unlikely to be a real issue and doesn't add the complexity of having to correctly handle quoting.

timsnyder-siv: I can't explain how happy it makes me that github allows you to to put poo emojis in code blocks...

```
// you should replace your code with
💩 💩 💩 
```
🤣

davidbiancolin: My understanding is that you escape double quotes in csv by doubing them, not by using a backslash. 
https://stackoverflow.com/questions/17808511/properly-escape-a-double-quote-in-csv

Quotes shouldn't be escaped in the description in a csv-consistent way, since the description should not be assumed to be bound for CSV. Worst case, the user gets an extra pair of double quotes -- i think that's alright.

sim/src/main/scala/midasexamples/AutoCounterModule.scala
davidbiancolin: It's different because i can't simply split on ',' and unswizzle the columns, since there would be more columns. Yes, inspected the CSV for correctness, though perhaps i should've tried uploading it to google sheets / excel as well.

davidbiancolin: I thought about folding in the csv library from json4s but it seemed a little overkill at the time, perhaps i was mistaken.

davidbiancolin: Google sheets appears to do the right thing

sim/midas/src/main/cc/bridges/autocounter.cc
timsnyder-siv: I think that I want to run
```
testonly firesim.midsasexamples.AutoCounterModule
```

to run only the AutoCounterModule test.

davidbiancolin: You want this: 

```testOnly firesim.midasexamples.AutoCounterModuleF1Test ```

You need to be in the firesim subproject to make that work though.

timsnyder-siv: > My understanding is that you escape double quotes in csv by doubing them, not by using a backslash.

Thanks for the pointer.  I hadn't seen that before.  It would be good to include a pointer to that SO question and then ignore my escaping with backslash.

I looked and `pandas.read_csv(doublequote=True)` (the default) follows this behavior as well:
image

sim/src/main/scala/midasexamples/AutoCounterModule.scala
timsnyder-siv: Thanks for checking it again and for the pointer to SO above.

PR #889, opened 2021-12-01
Bake-in FASED default runtime configuration into hardware

davidbiancolin:
This adds a mode that makes it easier to reuse runtime configuration files across different FASED instances. At SiFive,
we want to use the same latency for the LBP across different designs, but we also want to model to accept as many outstanding requests as possible. Since the latter is target dependent, we want to have runtime confs that only specify latency and let the hardware decide how to handle the rest. With this feature you can do that the following conf:

+mm_useHardwareDefaultRuntimeSettings_0
+mm_readLatency_0=<lat_read>
+mm_writeLatency_0=<lat_write>

The bridge driver will report the default values (read from the MMIO registers) if no plus arg is provided.

This PR touches a bunch of fasedtests code to make it easier to test this feature. I should probably try to better unify/clean up the midasexamples and fasedtests makefrags and scalatest suites in a future PR. 

Cherry-picks a fix from 887 to get CI working.

#### UI / API Impact

Adds a plus arg for FASED timing models, that changes runtime configuration semantics. The default behavior is unchanged.

Changes the generated verilog, but not the simulator memory map.

### Contributor Checklist


sim/src/test/scala/fasedtests/FASEDTestSuite.scala
```diff
@@ -33,24 +64,66 @@ abstract class FASEDTest(
...
       }
     }
   }
-  def runTests() {
-    runTest("verilator", false)
-    //runTest("vcs", true)
-  }
 
-  behavior of s"FASED Instance configured with ${platformConfigs} driven by target: ${topModuleClass}"
+  compileMlSimulator("verilator", false)
```
abejgonzalez, line 97: Why do you need to call compile before running? Shouldn't the make system handle this?

abejgonzalez (approved):
Overall seems fine to me, just have a single question on the test.

sim/src/test/scala/fasedtests/FASEDTestSuite.scala
davidbiancolin: It provides a slightly more specific failure report in scalatest, but yeah, the make system could handle this fine.

PR #913, opened 2022-01-25
Remove WithAutoILA from default recipes

### LINKED ISSUE #695: Default FPGADebug Annotations & ILA Platform Config Switch

davidbiancolin:
Users should explicitly mix this in like other debug features. This makes it easier to check in default FPGADebugAnnotations and makes AutoILA consistent with other debug features.

#### Related PRs / Issues

Resolves #695.

#### UI / API Impact

Users must mix in "WithAutoILA_" to their platform configs.

None in practice, since we don't tend to include FPGADebugAnnotations by default.

### Contributor Checklist


abejgonzalez (approved):
LGTM.

PR #932, opened 2022-02-11
Update machine launch script for Ubuntu/Centos

abejgonzalez:
Support both Ubuntu and Centos with the machine launch script.

#### UI / API Impact

Adds `-help` and `-launch-status-path` to `machine-launch-script.sh`. By default both are disabled matching the previous API.

### Contributor Checklist


scripts/machine-launch-script.sh
```diff
@@ -1,57 +1,156 @@
 #!/bin/bash
 
+LAUNCH_STATUS_PATH="/home/centos"
+
+usage()
+{
+    echo "Usage: $0 [options]"
+    echo
+    echo "[-help]                  List this help"
```
davidbiancolin, line 9: Can we conform to the -- / - style used elsewhere in our bash scripts.

scripts/machine-launch-script.sh
```diff
@@ -1,57 +1,156 @@
...
+    esac
+done
+
 set -ex
 set -o pipefail
 
-echo "machine launch script started" > /home/centos/machine-launchstatus
-sudo chgrp centos /home/centos/machine-launchstatus
-sudo chown centos /home/centos/machine-launchstatus
+FLAVOR=`grep '^ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
```
davidbiancolin, line 35: nit: favor `$( )` over backticks

scripts/machine-launch-script.sh
```diff
@@ -1,57 +1,156 @@
 #!/bin/bash
 
+LAUNCH_STATUS_PATH="/home/centos"
```
davidbiancolin, line 3: Can we just drop it in the the working directory  instead of using this absolute path which may not exist on some systems?

davidbiancolin (commented):
This really seems like hte sort of thing @a0u would have good opinions about. It would be nice to list packages in a disto-agnostic way instead of having to enumerate two separate lists. 

If both branches get tested in CI i'm ok with it for now.

scripts/machine-launch-script.sh
```diff
@@ -1,57 +1,156 @@
...
-then
-    git clone http://git.veripool.org/git/verilator
-    cd verilator/
-    git checkout v4.034
-    autoconf && ./configure && make -j4 && sudo make install
-    cd ..
+echo "machine launch script started" > $LAUNCH_STATUS_PATH/machine-launchstatus
+if [ $(whoami) == "centos" ]; then
+    sudo chgrp centos $LAUNCH_STATUS_PATH/machine-launchstatus
+    sudo chown centos $LAUNCH_STATUS_PATH/machine-launchstatus
```
timsnyder-siv, line 48: Abe, are you sure that when the script is copy-pasted into userdata that it gets run as centos user?  I think that it gets run as root and that's the reason why they chgrp and chown it.  Probably makes more sense to just `chmod ugo+r` the log.

scripts/machine-launch-script.sh
timsnyder-siv: I was curious where this would go if put in userdata without the full path.  On an Ubuntu 18.04 server, it ended up in /usr/share/doc/cloud-init/userdata.txt

scripts/machine-launch-script.sh
abejgonzalez: It does run as root. The lines added were to make sure the `centos` user had access to the log. I'm happy to use `ugo+r` instead.

scripts/machine-launch-script.sh
timsnyder-siv: On Ubuntu 18.04, it will go into the root of the filesystem `/machine-launchstatus`.  I suspect that it will be the same for anything else using AWS userdata because that all happens using cloud-init.

scripts/machine-launch-script.sh
```diff
@@ -76,8 +175,11 @@ sudo python3 -m pip install pyyaml==5.4.1
 # setup argcomplete
 activate-global-python-argcomplete
 
-} 2>&1 | tee /home/centos/machine-launchstatus.log
+} 2>&1 | tee $LAUNCH_STATUS_PATH/machine-launchstatus.log
+
+if [ $FLAVOR == "ubuntu" ]; then
+    # default /bin/sh to bash (not dash)
+    sudo dpkg-reconfigure -p critical dash
+fi
```
timsnyder-siv, line 183: As a matter of personal preference, I like this and always do it on Ubuntu machines that I admin.  However, I don't think you guys should _require_ this.  Anywhere that you are using `/bin/sh` and really want it to be BASH, you should really use `/bin/bash`.

abejgonzalez:
> From an offline conversation:
> 
> Have some file (like `.json`) that has English names of packages and the distro packages associated with it (in however many distros supported). This is passed to some script (Python?) that takes this file and properly fills out the `machine-launch-script.sh` (from some sort of template). Note: When this generation is done is still unknown (in CI commit the auto-genned script? By someone modifying the `.json` at PR time? etc).
> 
> Overall documentation should be added that talks about the following:
> 
> * Show the json file (say that this is the official dep. list)
> * Mention that `libdwarf/elf` also need to be build (point to script section that builds these from source)
> * Talk about NBD kernel build (also point to the part of the script that builds it)
> 
> Need to also abstract out the "deploy host" dependencies (i.e. NBD) from the manager dependencies (what is currently the `machine-launch-script.sh`). Move `libdwarf/elf` to the machine launch script and make NBD optional if you are not running on AWS F1.

On further thought, this json file has to also probably encode the version wanted and any setup needed so that the package is installed properly (i.e. git needs to have `sudo yum install -y epel-release` and `sudo yum install -y https://repo.ius.io/ius-release-el7.rpm` run before). This seems like it would get annoying to maintain and would just move the complexity of the machine launch script into a json file.

scripts/machine-launch-script.sh
timsnyder-siv: I think my point was that the whoami wouldn't return centos when you wanted to chown the log. Also centos won't exist on Ubuntu. Should just make sure it is world readable

timsnyder-siv:
@abejgonzalez , this is superseded by #986  right?

abejgonzalez:
> @abejgonzalez , this is superseded by #986 right?

Correct. Thanks!

PR #936, opened 2022-02-15
Tag AGFI with sha of toplevel repo

timsnyder-siv:
Similar to how we changed repo_state_summary.sh to create repo_state file from
the toplevel git superproject, generate the commit tag for the AGFI from the toplevel
repo.  Also include repo_state in the results-build/<build>/cl_firesim/design/repo_state
Somehow, it isn't in the deps of replace-rtl anymore.

#### UI / API Impact

Doesn't change the UI/API at all.

Doesn't change the Verilog, only improves the metadata in AGFI tags if using something other than Firesim as toplevel repo.

### Contributor Checklist
- [X] Did you set dev as the base branch?
- [X] Did you add Scaladoc to every public function/method?
       n/a
       ~~no.  Adding tests for this would be moderately obnoxious.  I would have to reuse git testing infra to buildup hierarchies of superprojects and I just did some manual QA of this with firesim as top and something else as top.~~ second pass added tests.

- [X] (If applicable) Did you regenerate and publicly share default AGFIs?

- [X] (If applicable) Did you mark the PR as "Please Backport"?
- [X] (On merge) Did you update release notes in the dev-to-master PR ?

### ToDo
- [ ] add something that prints out the metadata that is being shared by shareagfi

### DIFF
--- a/deploy/awstools/afitools.py
+++ b/deploy/awstools/afitools.py
@@ -1,7 +1,9 @@
 """ Tools to help manage afis. """
-
+import distutils.util
 import logging
 import boto3
+from pprint import pformat
+from schema import Schema, And, SchemaError, Or, Use
 
 rootLogger = logging.getLogger()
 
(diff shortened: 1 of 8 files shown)


davidbiancolin (commented):
My preference would be to have tags for both firesim (with the existing key) and the top-level super project.

timsnyder-siv:
> My preference would be to have tags for both firesim (with the existing key) and the top-level super project.

Okay.  I'll do this.  I didn't do before because it seems like the firesim tag will be redundant if it is truely a submodule of what we're tagging with but `more > less` and you're the boss. ;)

If we're adding tags, would it be worthwhile to capture the URI for the 'origin' remote for both repos as well?  I stopped short of that but it might be helpful for repro purposes.

davidbiancolin:
> Okay. I'll do this. I didn't do before because it seems like the firesim tag will be redundant if it is truely a submodule of what we're tagging with but `more > less` and you're the boss. ;)

You're right of course, i just would rather not change the meaning of the tag.

> If we're adding tags, would it be worthwhile to capture the URI for the 'origin' remote for both repos as well? I stopped short of that but it might be helpful for repro purposes.

I think that's a great idea if it's not too much work.

deploy/awstools/afitools.py
```diff
@@ -95,18 +97,69 @@ def share_agfi_in_all_regions(agfi_id, useridlist):
...
+TAG_DICT_SCHEMA = Schema({
+    'firesim-buildtriplet': STR256,
+    'firesim-deploytriplet': STR256,
+    'firesim-commit': STR256,
+    'firesim-origin': STR256,
+    'firesim-ispushed': STRBOOL,
+    'top-commit': STR256,
+    'top-origin': STR256,
+    'top-ispushed': STRBOOL,
+})
```
timsnyder-siv, line 118: I like doing this because it is flexible enough to coerce the types at runtime and we're going from strings to typed dicts but perhaps https://docs.python.org/3/library/typing.html#typing.TypedDict should be used?

timsnyder-siv:
I've asked https://github.com/filipstamenkovic-sifive to also rewrite this PR onto the 1.14 manager.

timsnyder-siv:
Superceded by rewrite in #1162

PR #940, opened 2022-02-15
Launchrunfarm patient perseverence (local-branch-redux)

timsnyder-siv:
Trying #716 with a local branch to see if the CI works better.

#### UI / API Impact

Adds two parameters to config_runtime.ini to be able to set a timeout for how long to keep looping through AvailabilityZones trying to `launch_instances` and whether `launch_instances` looks at instances that already exist with the specified type and tag and only tries to launch enough to bring up to the requested count or if it always expands the tagged instance pool by the requested count.

Doesn't touch verilog or AGFI compatability

### Contributor Checklist
- [X] Did you set dev as the base branch?
       Nope because I didn't touch scala bro.

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -159,7 +159,7 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior, 
         assert False, "INVALID INSTANCE MARKET TYPE."
 
 def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavior, spotmaxprice, blockdevices=None,
-                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), additive=True):
+                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), expand_runfarm_by_count=True):
     """ Launch `count` instances of type `instancetype`
 
     Using `instancemarket`, `spotinerruptionbehavior` and `spotmaxprice` to define instance market conditions
@@ -186,10 +186,10 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
         Path to readable file.  Contents of file are passed as `UserData` to AWS
     timeout : datetime.timedelta, default=timedelta() (immediate timeout after attempting all subnets)
         `timedelta` object representing how long we should continue to try asking for instances
-    additive : bool, default=True
+    expand_runfarm_by_count : bool, default=True
         When true, create `count` instances, regardless of whether any already exist. When False, only
         create instances until there are `count` total instances that match `tags` and `instancetype`
-        If `tags` are not passed, `additive` must be `True` or `ValueError` is thrown.
+        If `tags` are not passed, `expand_runfarm_by_count` must be `True` or `ValueError` is thrown.
 
     Return type
     -----------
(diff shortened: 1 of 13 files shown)


davidbiancolin:
Yes, you'll need push to upstream to get CI to run entirely. We have mergify to automate copying PRs but for frequent contributors with write permissions this is superior.

davidbiancolin:
Interesting, you got transitive approval from #716... GH magic

deploy/awstools/awstools.py
```diff
@@ -153,13 +158,53 @@ def construct_instance_market_options(instancemarket, spotinterruptionbehavior,
...
-def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavior, spotmaxprice, blockdevices=None, tags=None, randomsubnet=False, user_data_file=None):
-    """ Launch count instances of type instancetype, optionally with additional
-    block devices mappings and instance tags
-
-         This will launch instances in avail zone 0, then once capacity runs out, zone 1, then zone 2, etc.
+def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavior, spotmaxprice, blockdevices=None,
+                     tags=None, randomsubnet=False, user_data_file=None, timeout=timedelta(), always_expand=True):
+    """ Launch `count` instances of type `instancetype`
+
+    Using `instancemarket`, `spotinerruptionbehavior` and `spotmaxprice` to define instance market conditions
```
sagark, line 165: spotinterruptionbehavior

deploy/tests/test_amis_snake.json
```diff
@@ -0,0 +1,19 @@
+[
```
sagark, line 1: is there a script to produce this when the AMI changes? (or really, is there an easy way to get all this information given the AMI id/name)?

deploy/awstools/awstools.py
timsnyder-siv: Thanks.  Good catch.
```suggestion
    Using `instancemarket`, `spotinterruptionbehavior` and `spotmaxprice` to define instance market conditions
```

scripts/aws-setup.py
```diff
@@ -1,60 +1,11 @@
-#/usr/bin/env python3
+#!/usr/bin/env python3
```
sagark, line 1: The wget path in /docs/Initial-Setup/Configuring-Required-Infrastructure-in-Your-AWS-Account.rst

needs to be updated to point to the new location.

deploy/tests/test_amis_snake.json
timsnyder-siv: I could/should create one.  I just did `aws ec2 describe-images` to get the AMI json and then munged it into the format that moto expects/accepts.  Before I create one, I'll ask the moto devs if they already have one.

scripts/aws-setup.py
timsnyder-siv: Good catch. I forgot about that expecting to be able to grab the single script to download and run.  Rather than changing the path, should we just check in a symlink so that the wget still works with the old path?  Don't know if anyone has that scripted into anything else.

sagark: Does wget follow symlinks? If it does, I like that solution.

sagark (approved):
LGTM modulo wget aws-setup fix!

deploy/tests/test_amis_snake.json
timsnyder-siv: Don't even need to ask, just have to look.  oof. https://github.com/spulec/moto/blob/master/scripts/get_amis.py. 

I'll vendor that script into firesim/scripts with smarts to look at the AMI defined in awstools.py and add a python test that fails with a nicer message if the AMI file needs to be updated (otherwise, you'll just get an index error the first time we try to query it and index into the result that doesn't find the AMI we think should always be found).

scripts/aws-setup.py
timsnyder-siv: GH doesn't serve symlinks as symlinks in the raw.githubusercontent.com URL, it serves them as txt files that have the target of the symlink.  Not what we want.  It looks like those docs may need to be updated to say python3 because I have a feeling that `python` on the FPGA dev AMI may still be python2...  I'm going to double check before updating the docs.

timsnyder-siv: Looks like @abejgonzalez already changed the docs on dev to use `./aws-setup.py` to fix the python 2 vs 3 issue.  However, the script had a malformed shebang line and my PR is fixing that.  I'll also update the docs to grab it from the new location with the full-flavor script.

deploy/tests/test_amis_snake.json
timsnyder-siv: I'm almost done with creating this script but @davidbiancolin asked me to push it in a follow-on PR so that we can cut the release.  I'll open a follow-on PR later tonight that has it.

davidbiancolin: Thanks Tim.

PR #955, opened 2022-02-18
[mergify] Remove milestone requirement in backport scheme

davidbiancolin:
This is a completely optional change, but it simplifies the backport system to require only the `Please Backport` tag as the PR template suggests instead of also requiring the milestone.

We've never really supported older versions of the code base beyond the last minor version (even that's debatable), so i don't think it's necessary to include the milestone from FIRRTL where they want to backport to multiple target branches for multiple minor releases. 

#### Related PRs / Issues

None

#### UI / API Impact
None

N/C

### Contributor Checklist

- [x] (If applicable) Did you mark the PR as "Please Backport"?

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,9 +1,11 @@
 pull_request_rules:
-  - name: backport to master
+  # For FireSim version 1.x.y, here let x = minor, y = patch
+  # Only support backporting to the last minor release branch.
+  # This rule will need to be updated on minor releases.
+  - name: backport to latest minor release
     conditions:
       - merged
       - base=main
-      - milestone=1.13.x
       - label="Please Backport"
     actions:
       backport:


.mergify.yml
```diff
@@ -1,9 +1,11 @@
 pull_request_rules:
-  - name: backport to master
+  # For FireSim version 1.x.y, here let x = minor, y = patch
+  # Only support backporting to the last minor release branch.
+  # This rule will need to be updated on minor releases.
+  - name: backport to latest minor release
     conditions:
       - merged
       - base=main
-      - milestone=1.13.x
```
timsnyder-siv, line 6: This seems like a reasonable change to me.  It's annoying that you have to hardcode the target branch and update that on every release below...

timsnyder-siv (approved):

.mergify.yml
abejgonzalez: This does seem like an easy thing to automate though. I assume there is some GH-A that can do some regex-ing and replace the string based on the release tag...

davidbiancolin: I'll track more release automation features in a follow up issue.

PR #985, opened 2022-03-11
Rewrite yaml api tests for where-to-build

timsnyder-siv:
@abejgonzalez, I started rewriting the YAML API tests and have gotten far enough that I should open this for others to see.

There are pros and cons of doing what I'm doing.  Here are the ones I see so far:
### PROS
* It is much easier to tightly test for something specific.  For example, your tests just check for non-zero exit code and I think most of them are currently 'passing' in your implementation but they are exiting non-zero because your test recipes are out of date with the YAML API (need to be updated with `default-build-farm` in config_build.yaml) but nothing in your tests notices that the test input is no longer valid.  See https://github.com/firesim/firesim/runs/5461951565?check_suite_focus=true#step:3:47
* Stands alone and is trivial to run locally for testing.  I had trouble trying to run your GH action script locally to see how the tests were expected to behave before my changes.
* Easy to mock out the AWS interaction so that the tests can't actually create AWS resources
* I try to base the tests on the sample config files in deploy/backup-sample-configs and then programmatically modify them so that as we update those, the tests primarily come along for the ride (unless they are testing the specific things that changed).  

### CONS
* actually running fabric stuff under pytest is interesting because pytest blocks all reading from stdin and fabric seems to be hitting the thing pytest installs to prevent reading from stdin during tests.  This is probably just a fabric configuration issue during testing and we shouldn't need to actually execute anything in the fabric tasks to test the YAML api
    * fab-classic seems to use nose instead of pytest.  I've steered away from nose in the past because it is more heavyweight and seemed like it was becoming unmaintained and less popular.
* requires learning about pytest and how fixtures work.  There's about 200 lines of testing infrastructure code for another 200 lines of tests in what I wrote.

#### Related PRs / Issues

Rewrites part of #850 that adds yaml API tests

No change in Verilog or AGFI

### Contributor Checklist


deploy/firesim
```diff
@@ -26,6 +27,25 @@ from util.streamlogger import StreamLogger
 ## below are tasks that users can call
 ## to add a task, add it here and to the choices array
 
+"""Registry of tasks that use `BuildConfigFile`"""
+BUILD_TASKS: Dict[str, Callable] = {}
```
abejgonzalez, line 31: Why is this preferred to using `globals()`?

abejgonzalez: Actually, I see why this is needed, we need this to call the function in the `pytests`. Is there a reason to split this up into `BUILD_` and `RUN_` or is that mainly for organization?

deploy/tests/test_yaml_api.py
```diff
@@ -0,0 +1,402 @@
...
+class TestConfigBuildAPI:
+    """ Test config_{build, build_recipes}.yaml APIs """
+
+    # run_test("invalid-build-section")
+    @patch('firesim.buildafi')
+    def test_invalid_build_section(self, buildafi_mock, build_yamls, fs_cl2args):
+        # @patch modifies the symbol table with our Mock but we're dispatching through
+        # a dict and that needs to be patched as well, otherwise our mock won't be called but
+        # the real function will be
+        firesim.BUILD_TASKS['buildafi'] = buildafi_mock
```
abejgonzalez, line 214: What defines this `buildafi_mock` (fixture right)? This is just removing the calls to the AWS EC2 API right?

deploy/tests/test_yaml_api.py
```diff
@@ -0,0 +1,402 @@
...
+        os.environ['FIRESIM_SOURCED'] = '1'
+        # at the beginning of the test build_yamls contains the backup-sample-configs
+        # but we can show exactly what we're doing different from the default by
+        # by programatically modifying from the sample to remove all the builds
+        #build_yamls.build.data['builds'] = []
+        # lol sometimes, it's hard to tell exactly how the yaml will get loaded because None is what you get
+        build_yamls.build.data['builds'] = None
```
abejgonzalez, line 247: Why do we care about the programmatic version of this test? Isn't the main/only API to modify the `builds` section through the file (which is tested in the  `test_invalid_build_section` test?

deploy/tests/test_yaml_api.py
```diff
@@ -0,0 +1,402 @@
...
+        build_yamls.write()
+        args = firesim_argparser.parse_args(['buildafi'] + build_yamls.args)
+        pprint(build_yamls.build.data)
+        firesim.main.when.called_with(args).should.throw(TypeError)
+        buildafi_mock.assert_not_called()
+
+
+    # run_test("invalid-aws-ec2-inst-type")
+    @pytest.mark.skip(reason="unclear as to what this is testing even in Abe's impl")
```
abejgonzalez, line 256: I wanted to see if the AWS EC2 API would error out saying invalid instance type. We can probably remove this test since it is just testing if the AWS EC2 API can properly catch its own errors.

abejgonzalez (approved):
This looks much much better than the hacky tests previously done. I see that throughout the code you can set individual yaml sections to ` = None` instead of having to write out yaml directly which is nice!

abejgonzalez:
> * actually running fabric stuff under pytest is interesting because pytest blocks all reading from stdin and fabric seems to be hitting the thing pytest installs to prevent reading from stdin during tests.  This is probably just a fabric configuration issue during testing and we shouldn't need to actually execute anything in the fabric tasks to test the YAML api
>   
>   * fab-classic seems to use nose instead of pytest.  I've steered away from nose in the past because it is more heavyweight and seemed like it was becoming unmaintained and less popular.

Does this manifest in an error?

deploy/firesim
timsnyder-siv: They are split because if `args.task` is in `BUILD_TASKS` then you make a `BuildConfigFile(args)` and pass that into calling the task.  Likewise for `RUN_TASKS` you make a `RuntimeConfig(args)` and pass that into the task.

deploy/tests/test_yaml_api.py
timsnyder-siv: The `@patch` decorator passes it in as the first fixture.  I kinda dislike writing it this way and I'm remembering that I dislike it because it's not obvious that `@patch` is doing that.  I'll look at how to write this a better way.

deploy/tests/test_yaml_api.py
timsnyder-siv: I only did it to show that you can do it this way for testing purposes.  I suppose I should point out that while you're programmatically modifying the YAML, it gets serialized out to a temporary file that gets read in by firesim.

abejgonzalez: Is it necessary for a unique test for this? Aren't the latter tests doing this anyways?

deploy/tests/test_yaml_api.py
timsnyder-siv: Ok.  I went back and reviewed some things.  I think the pytest-mock plugin makes this code look much better.  I was going to use it when I added the pytests but was hesitant to add yet another dependency.  However, looking back at https://github.com/pytest-dev/pytest-mock/#why-bother-with-a-plugin I'm really seeing how much nicer it would be.  I'm going to add that quick this afternoon and finish implementing the rest of the tests so we can merge this.

timsnyder-siv:
> Does this manifest in an error?

Yes,  If I hack one of the tests that doesn't currently blow up during reading the config files and let it actually call buildafi, the output looks like this:
```
   @patch('firesim.buildafi')
    def test_invalid_unmanaged_missing_args(self, buildafi_mock, build_yamls, firesim_argparser):
        # @patch modifies the symbol table with our Mock but we're dispatching through
        # a dict and that needs to be patched as well, otherwise our mock won't be called but
        # the real function will be
        #firesim.BUILD_TASKS['buildafi'] = buildafi_mock
        os.environ['FIRESIM_SOURCED'] = '1'
    
        build_yamls.farm.data['local-build-farm']['args']['hosts'] = None
    
        build_yamls.write()
        args = firesim_argparser.parse_args(['buildafi'] + build_yamls.args)
>       firesim.main(args)

test_yaml_api.py:354: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../firesim.py:285: in main
    BUILD_TASKS[args.task](build_config_file)
../firesim.py:118: in buildafi
    execute(replace_rtl, build_config, hosts=['localhost'])
../../../.local/lib/python3.8/site-packages/fabric/tasks.py:377: in execute
    results[host] = _execute(
../../../.local/lib/python3.8/site-packages/fabric/tasks.py:268: in _execute
    return task.run(*args, **kwargs)
../../../.local/lib/python3.8/site-packages/fabric/tasks.py:168: in run
    return self.wrapped(*args, **kwargs)
../buildtools/buildafi.py:49: in replace_rtl
    run(build_config.make_recipe("PLATFORM=f1 replace-rtl"))
../../../.local/lib/python3.8/site-packages/fabric/network.py:700: in host_prompting_wrapper
    return func(*args, **kwargs)
../../../.local/lib/python3.8/site-packages/fabric/operations.py:1068: in run
    return _run_command(
../../../.local/lib/python3.8/site-packages/fabric/operations.py:919: in _run_command
    result_stdout, result_stderr, status = _execute(
../../../.local/lib/python3.8/site-packages/fabric/operations.py:807: in _execute
    worker.raise_if_needed()
../../../.local/lib/python3.8/site-packages/fabric/thread_handling.py:26: in raise_if_needed
    six.reraise(e[0], e[1], e[2])
../../../miniforge3/envs/firesim_oss_220203/lib/python3.8/site-packages/six.py:719: in reraise
    raise value
../../../.local/lib/python3.8/site-packages/fabric/thread_handling.py:13: in wrapper
    callable(*args, **kwargs)
../../../.local/lib/python3.8/site-packages/fabric/io.py:270: in input_loop
    byte = f.read(1)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <_pytest.capture.DontReadFromInput object at 0x7fbf58c59b20>, args = (1,)

    def read(self, *args):
>       raise OSError(
            "pytest: reading from stdin while output is captured!  Consider using `-s`."
        )
E       OSError: pytest: reading from stdin while output is captured!  Consider using `-s`.

../../../miniforge3/envs/firesim_oss_220203/lib/python3.8/site-packages/_pytest/capture.py:192: OSError
```

My explanation in the meeting was wrong.  I think that fabric.run by default puts a PTY on stdin and the pytest input capturing is being triggered by that.  I see that new Fabric does have https://docs.fabfile.org/en/2.6/api/testing.html that provides some helpers for testing fabric driven code.

I've been thinking that there is a lot of room for unit-tests of the manager that don't need fabric and then like we discussed, the end-to-end testing at a higher level in CI is great.

deploy/tests/test_yaml_api.py
timsnyder-siv: Ok.  Thanks for clarifying.  I wasn't sure if you were somehow validating that in the YAML early or testing AWS.  I'll remove that.

deploy/tests/test_yaml_api.py
timsnyder-siv: No, I'll defiintely remove it before we merge.  Was only trying to have food for discussion in the PR.

davidbiancolin (commented):
Looks great to me. 

Spitballing, I wonder if _register_task should mandate providing the Mock there. That would clean up the patching. This may all be moot if the plugin work  well though.

deploy/firesim
```diff
@@ -270,11 +291,11 @@ def main(args):
                          args.terminatesomef12, args.terminatesomem416,
                          args.forceterminate)
 
-    if args.task in ['launchrunfarm', 'infrasetup', 'boot', 'kill', 'runworkload', 'runcheck']:
+    if args.task in RUN_TASKS:
         # construct simulation config
         simconf = RuntimeConfig(args)
         # run task
-        globals()[args.task](simconf)
+        RUN_TASKS[args.task](simconf)
```
timsnyder-siv, line 298: Thinking about @abejgonzalez's question about why I have BUILD_TASKS and RUN_TASKS in two separate datastructures, it would probably be nicer if it was a dict of dicts that also stored the Config class that the task wanted, we could just do this:
```
        f = TASKS[args.task]['function']
        conf = TASKS[args.task]['config_class'](args)
        f(conf)
```
And we could tweak managerinit and terminaterunfarm slightly and not need to have special casing  either.  I like it.

.github/scripts/yaml-tests/failing-runtime-files/hwdb-invalid-afi/sample_config_hwdb.yaml
```diff
@@ -1,4 +0,0 @@
-testing-hwdb-entry:
-    agfi: INVALID_AGFI
```
timsnyder-siv, line 2: @abejgonzalez , This test also seems like something that shouldn't be caught by firesim before asking AWS to really do something with the AGFI.  I'm going to leave it out

deploy/firesim
timsnyder-siv: @abejgonzalez , I've pushed the rework I did.  You should see that there is only a unified TASKS datastructure now and that register_task is used as a decorator.  I also switched the tests over to using the pytest-mock `mocker` fixture and I think it makes the mock creation look cleaner than using the `@patch` decorator like I was doing before.

deploy/tests/test_yaml_api.py
timsnyder-siv: This is removed now.

deploy/tests/test_yaml_api.py
timsnyder-siv: I added a test to test_awstools.py that shows `moto` will let you create instances with bogus types when using the mocking library.  I might also add a test for whether aws will properly catch something bogus as an example of how to write a test that does real AWS stuff.  Checking for that case isn't horrible since we are allowing something that a user puts in the config file to go directly to AWS without any kind of validation first.

timsnyder-siv:
@abejgonzalez , I think this is ready to merge into where-to-build if you take another look and agree.

abejgonzalez (approved):
LGTM with small nits

deploy/firesim
```diff
@@ -23,9 +30,72 @@ from buildtools.buildconfig import BuildConfig
...
+class FiresimTaskAccessViolation(Exception):
+    pass
+
+def register_task(task: Callable) -> Callable:
+    """Decorator to Map `task.__name__` to `task` in `TASKS`
+
+    Args:
+        task: callable task that either has no parameters or
+              the first parameter must be typed with a class
+              that accepts a argparse.Namespace as the only parameter
```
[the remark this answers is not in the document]
timsnyder-siv: I typically put code refs in backtics.  Thanks for catching this.

deploy/firesim
```diff
@@ -168,10 +247,12 @@ def terminaterunfarm(runtime_conf, terminatesomef1_16, terminatesomef1_4, termin
        that many instances of the specified types and leave all others
        untouched.
     """
-    runtime_conf.terminate_run_farm(terminatesomef1_16, terminatesomef1_4, terminatesomef1_2,
-                                    terminatesomem4_16, forceterminate)
+    args = runtime_conf.args
+    runtime_conf.terminate_run_farm(args.terminatesomef1_16, args.terminatesomef1_4, args.terminatesomef1_2,
```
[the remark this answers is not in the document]
timsnyder-siv: Sure.  I can do that.   I was just trying to see how much or if I got pushback on putting the args into RuntimeConfig.

deploy/firesim
```diff
@@ -237,7 +306,19 @@ def construct_firesim_argparser():
...
+def check_env() -> None:
+    """Check assumptions about environment setup
+
+    Raises:
+        `SystemExit` if any assumption is violated
+    """
+    # make sure that sourceme-f1 was sourced
+    if os.environ.get('FIRESIM_SOURCED') is None:
+        rootLogger.critical("ERROR: You must source firesim/sourceme-f1-manager.sh!")
+        exit(1)
```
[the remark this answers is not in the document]
timsnyder-siv: I'll look more closely at that.  This was refactored into a function without changing the call but it would be good to make things consistent.

deploy/firesim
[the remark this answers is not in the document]
abejgonzalez: I think putting args everywhere is fine for now... I'm not opposed to it right now.

deploy/firesim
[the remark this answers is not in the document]
abejgonzalez: This will probably be checked once we generate some document for the docstrings, so its not a big issue right now.

deploy/firesim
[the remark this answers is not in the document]
timsnyder-siv: done

deploy/firesim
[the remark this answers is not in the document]
timsnyder-siv: Thanks for pointing this out.  I changed all uses of `exit()` to `sys.exit()`

PR #986, opened 2022-03-11
Use conda for distribution-agnostic dependency management

timsnyder-siv:
This uses packages from https://conda-forge.org/ can be used to install all of the packages in machine-launch.sh in a distribution-agnostic manner.  

#### Related PRs / Issues

This needs https://github.com/ucb-bar/chipyard/pull/1140 to be able to cleanly do build-setup.

#### UI / API Impact

Rewrites `scripts/machine-launch-script.sh` to use conda and deals with the CI-visible impact of switching from gcc 4.8.5 to 10.3.0.

Should be no Verilog changes.

### Contributor Checklist


scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+    # setup argcomplete
+    "${DRY_RUN_ECHO[@]}" $SUDO "${CONDA_ENV_BIN}/activate-global-python-argcomplete"
+
+    conda_init_extra_args=()
+    if [[ -n "$SUDO"  || "$USER" == root ]]; then
+	# if we're installing into a root-owned directory using sudo, or we're already root
+	# initialize conda in the system-wide rcfiles
+	conda_init_extra_args=(--no-user --system)
+    fi
+    $SUDO "${CONDA_EXE}" init $DRY_RUN_OPTION "${conda_init_extra_args[@]}" bash
```
timsnyder-siv, line 254: This should move up to be inside the conditioning of doing the conda install and not happen unless we're actually installing conda (as opposed to reusing an existing install in `--prefix`)

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
 #!/bin/bash
 
+#if CONDA_PREFIX already defined (because you already have a conda environment activated), just do the env creation in that prefix
+#the cmdline --prefix overrides this
```
timsnyder-siv, line 4:
```suggestion
```
I removed the behavior but forgot to remove the comment.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
-echo "PS1='\u@\H:\w\\$ '" >> /home/centos/.bashrc
-echo "machine launch script completed" >> /home/centos/machine-launchstatus
+
+    if [[ ! -r /etc/os-release ]]; then
+	echo "::ERROR:: $0 depends on /etc/os-release for distro-specific setup and it doesn't exist here"
+	exit 1
+    fi
+
+    if [[ "$(uname -io)" != "x86_64 GNU/Linux" ]]; then
+	echo "::ERROR:: $0 only supports 'x86_64 GNU/Linux' not '$(uname -io)'"
```
timsnyder-siv, line 74: Could probably be made to support Darwin without too much effort.  I know there are many dependencies that just don't work on Windows.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+	echo "::INFO:: installing conda to '$CONDA_INSTALL_PREFIX'"
+	# -b for non-interactive install
+	$SUDO bash ./install_conda.sh -b -p "$CONDA_INSTALL_PREFIX"
+    fi
+
+    # see https://conda-forge.org/docs/user/tipsandtricks.html#multiple-channels
+    # for more information on strict channel_priority
+    $SUDO "$CONDA_EXE" config --system --set channel_priority strict
+    # By default, don't mess with people's PS1, I personally find it annoying
+    $SUDO "$CONDA_EXE" config --system --set changeps1 false
```
timsnyder-siv, line 130:
```suggestion

        # see https://conda-forge.org/docs/user/tipsandtricks.html#multiple-channels
        # for more information on strict channel_priority
        $SUDO "$CONDA_EXE" config --system --set channel_priority strict
        # By default, don't mess with people's PS1, I personally find it annoying
        $SUDO "$CONDA_EXE" config --system --set changeps1 false
    fi
```

Hoist the system-wide conda-config stuff into only happening if we are doing the conda install

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+    OS_VERSION=$(grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"')
+
+    echo "machine launch script started" > machine-launchstatus
+    chmod ugo+r machine-launchstatus
+
+    # platform-specific setup 
+    case "$OS_FLAVOR" in
+	ubuntu)
+	    # default /bin/sh to bash (not dash)
+	    sudo dpkg-reconfigure -p critical dash
```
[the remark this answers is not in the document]
abejgonzalez: I think we can just remove this for now, and for the scripts that have issues we can update them.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
 #!/bin/bash
 
+#if CONDA_PREFIX already defined (because you already have a conda environment activated), just do the env creation in that prefix
+#the cmdline --prefix overrides this
+CONDA_INSTALL_PREFIX=/opt/conda
+CONDA_INSTALLER=https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh
```
abejgonzalez, line 6: I feel like we should probably version this.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+
+
+    # https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#using-the-compiler-packages
+    #    for notes on using the conda compilers
+    CONDA_PACKAGE_SPECS=( gcc_linux-64 gxx_linux-64 )
+
+    # if we want to install prebuilt riscv toolchain, the bdutrosv channel has riscv-gnu-toolchain
+
+    # poky deps
+    CONDA_PACKAGE_SPECS+=( python=3.8 patch texinfo subversion chrpath git wget )
```
abejgonzalez, line 148: While we are at it, should we bump to a newer python (3.10.2)?

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+    CONDA_PACKAGE_SPECS+=( verilator==4.034 )
+
+    
+    # python packages
+    # While it is possible to install using pip after creating the
+    # conda environment, pip's dependency resolution can conflict with
+    # conda and create broken environments.  It's best to use the conda
+    # packages so that the environment is consistent
+    CONDA_PACKAGE_SPECS+=( \
+	boto3==1.20.21 \
```
abejgonzalez, line 202: I'm not sure if these versions will change on a bump to the python version. Probably worthwhile to double-check.

scripts/machine-launch-script.sh
timsnyder-siv: I know conda-forge is working through getting everything supported in 3.10.  https://conda-forge.org/status/#python310 looks like ~75% of the packages in conda-forge area available for python 3.10.  I can check.

I'd be highly in favor of also bumping the python version but didn't want to lump that in with this PR.

scripts/machine-launch-script.sh
timsnyder-siv:
```suggestion
CONDA_INSTALLER=https://github.com/conda-forge/miniforge/releases/download/4.11.0-4/Miniforge3-4.11.0-4-Linux-x86_64.sh
```
Sure, we can do this.

scripts/machine-launch-script.sh
timsnyder-siv: Sure.  If we bump python and the conda install fails because all of the requirements are unsolvable, it will tell us which ones have to be updated.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+	    ;;
+	*)
+	    echo "::WARNING:: Unknown OS flavor '$OS_FLAVOR'. Unable to do platform-specific setup."
+	    ;;
+    esac
+
+
+    # everything else is platform-agnostic and could easily be expanded to Windows and/or OSX
+
+    SUDO=""
```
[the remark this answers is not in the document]
timsnyder-siv: The script can be run without sudo if the CONDA_INSTALL_PREFIX is into a directory that is writable by the user running the script.  needing to use sudo and using an existing conda install are orthogonal issues.

The script defaults to installing in /opt/conda, even if you already have conda in your path.  If you want to use an existing conda install that you have activated (i.e. it is in your path) you should (as documented in the usage):
```
machine-launch-script.sh --prefix "$CONDA_PREFIX" --env <some_new_env_name>
```

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+        dtc \
+    )
+
+
+    # Why do we need these?
+    #   "perl(ExtUtils::MakeMaker)" \
+    # diffstat texi2html 
+
+    # verilator was added to conda-forge by timsnyder
+    CONDA_PACKAGE_SPECS+=( verilator==4.034 )
```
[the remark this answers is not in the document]
timsnyder-siv: Are you wanting to support building only Verilator from source in this script or any of the packages?  Just trying to understand the scope of what you're suggesting.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,260 @@
...
+        cmake \
+        graphviz \
+        expect
+        dtc \
+    )
+
+
+    # Why do we need these?
+    #   "perl(ExtUtils::MakeMaker)" \
+    # diffstat texi2html
```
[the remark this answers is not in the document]
timsnyder-siv:
```suggestion
```

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
timsnyder-siv: In theory, now that I have the verilator builds working again, any new versions that come out should be available within a day or two of them being released by the verilator guy and it would be as simple as just updating the version here.  I only say "in theory" because I was lazy between creating 4.034 and a few weeks ago and didn't bump for each version that came out.

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
abejgonzalez: Ah. I missed L113. My confusion is with the help string saying `<path>` which I assumed meant `$PATH`. Reading it more closely, its fine.

scripts/machine-launch-script.sh
abejgonzalez: Ok no problem. Future work.

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
abejgonzalez: Hmm. I was thinking if any package (Verilator or otherwise) needs to be built from source we can do that with a special hook. But the cleaner option is to just enforce that people add things to `conda-forge` before it can be added to this file. 

To make sure I understand, you added some automation that auto-builds new Verilator versions and adds it to `conda-forge`? Thats very nice.

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
abejgonzalez: Sure. Thats good with me.

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
timsnyder-siv: conda-forge has this automation for all packages and I'm committing to paying attention to new PR's that are opened in the verilator-feedstock repo for version bumps if we start using this.  The repo is https://github.com/conda-forge/verilator-feedstock if you are interested.  I guess there was a new PR opened two days ago that I hadn't noticed and I'm going to go now and look to make sure it seems okay and merge it so that the new version becomes available on anaconda.org. https://github.com/conda-forge/verilator-feedstock/pull/15

I think the cleanest thing for any conda package would be to contribute to bumping the feedstock recipe so that the new version gets built.  The easiest way to find the feedstock repos for a given package is on the conda-forge package listing at https://conda-forge.org/feedstock-outputs/. Each package there has a link to it's page on anaconda.org (where you can see the existing versions) and a link to it's feedstock repo (where you can see any new versions that are being worked on or submit a PR to build a different version.

I say it is "cleanest" because you already have a working "recipe" for doing the build and all of the dependencies captured.  Using the conda-forge CI to do the build will build it in their isolated docker images and you can have some confidence 

If you want to quickly do a build just to try things out, you can clone the feedstock repo, modify the `recipe/meta.yaml` to point to the new version and use the `./build-locally.py` to run the build in a local docker container and generate packages that you can install into your conda env without necessarily uploading them to anaconda.org.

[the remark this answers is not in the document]
timsnyder-siv: Oh, and the reason, it matters that I pay attention to verilator, is because I'm currently the only maintainer of the conda recipe...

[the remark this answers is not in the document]
timsnyder-siv: Perhaps I should add a page to the documentation that explains how to do all of what I said above with pointers into the conda-forge docs?

timsnyder-siv:
@abejgonzalez I've merged in the additional packages we added in #850 here and I think this is ready to try to run CI with.  I made it a draft last week and I'm a little surprised that when I converted it back today that CI didn't run.  Should I also put the special fpga label on here so that it runs more stuff?

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,264 @@
...
+    echo "                          NOTE: --dry-run will still install conda to --prefix"
+    echo
+    echo "Examples:"
+    echo "  % $0"
+    echo "     Install into default system-wide prefix (using sudo if needed) and add install to system-wide /etc/profile.d"
+    echo "  % $0 --prefix ~/conda --env my_custom_env"
+    echo "     Install into $HOME/conda and add install to ~/.bashrc"
+    echo "  % $0 --prefix \$CONDA_PREFIX --env my_custom_env"
+    echo "     Create my_custom_env in existing CONDA_PREFIX conda install"
+    echo "     NOTE: CONDA_PREFIX is set in your environment when you activate a conda env"
```
timsnyder-siv, line 29: I forgot that conda makes the CONDA_PREFIX environment variable what is called `active_prefix` inside the tool.  These instructions need to be updated to say:
```suggestion
    echo "  % $0 --prefix \$(conda info --json | jq -r '.conda_prefix') --env my_custom_env"
    echo "     Create my_custom_env in pre-existing conda install that you are using"
```

timsnyder-siv: A less convoluted suggestion or one that at least doesn't involve a pipeline of JSON being wrangled by jq might be:
```suggestion
    echo "  % $0 --prefix \${CONDA_EXE%/bin/conda} --env my_custom_env"
    echo "     Create my_custom_env in existing conda install"
    echo "     NOTE: CONDA_EXE is set in your environment when you activate a conda env"
```

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,264 @@
...
+
+    # Install python packages using pip that are not available from conda
+    #
+    # Installing things with pip is possible.  However, to get
+    # the most complete solution to all dependencies, you should
+    # prefer creating the environment with a single invocation of
+    # conda
+    PIP_PKGS=( \
+	fab-classic==1.19.1 \
+	mypy-boto3-ec2==1.21.9 \
```
timsnyder-siv, line 245:
```suggestion
	mypy-boto3-ec2==1.21.9 \
	sure==2.0.0 \
```
Noticed that I somehow missed `sure` when I grabbed the other new packages from where-to-build.

scripts/machine-launch-script.sh
```diff
@@ -1,83 +1,265 @@
...
+DRY_RUN_ECHO=()
+
+usage()
+{
+    echo "Usage: $0 [options]"
+    echo
+    echo "Options:"
+    echo "[--help]                  List this help"
+    echo "[--prefix <path>]         Install prefix for conda. Defaults to /opt/conda."
+    echo "                          If <path>/bin/conda already exists, it will be used."
```
timsnyder-siv, line 17:
```suggestion
    echo "[--prefix <prefix>]         Install prefix for conda. Defaults to /opt/conda."
    echo "                          If <prefix>/bin/conda already exists, it will be used."
```

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
timsnyder-siv: done

timsnyder-siv:
I think CI may be failing because I did what @davidbiancolin suggested in https://github.com/firesim/firesim/pull/932#discussion_r809629052 and started dropping the logs in the working directory and the CI script is still expecting the output to be in /home/centos?  I'm going to look more closely at the CI script and fix it.

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,265 @@
...
+DRY_RUN_OPTION=""
+DRY_RUN_ECHO=()
+
+usage()
+{
+    echo "Usage: $0 [options]"
+    echo
+    echo "Options:"
+    echo "[--help]                  List this help"
+    echo "[--prefix <prefix>]         Install prefix for conda. Defaults to /opt/conda."
```
davidbiancolin, line 16:
```suggestion
    echo "[--prefix <prefix>]       Install prefix for conda. Defaults to /opt/conda."
```

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,265 @@
...
+	exit 1
+    fi
+
+    OS_FLAVOR=$(grep '^ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"')
+    OS_VERSION=$(grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"')
+
+    echo "machine launch script started" > machine-launchstatus
+    chmod ugo+r machine-launchstatus
+
+    # platform-specific setup
```
davidbiancolin, line 82: So you said this was here as a demonstration right? You're going to remove it before PR?

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
davidbiancolin: It would be good if you added a comment explaining this logic, and made it more clear at the top of the file the conditions under which sudo is required to run this script.

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,265 @@
...
+	pyyaml==5.4.1 \
+	mypy==0.931 \
+	types-pyyaml==6.0.4 \
+	boto3-stubs==1.21.6 \
+	botocore-stubs==1.24.7 \
+	mypy-boto3-s3==1.21.0 \
+    )
+
+    if [[ "$CONDA_ENV_NAME" == "base" ]]; then
+	# we default to installing into the base environment against best practice because it
```
davidbiancolin, line 221: I know you've explained to me orally, but can you explain how it simplifies it here.

davidbiancolin (commented):
This looks really cool. Seems to me like it would be easy to further configure this to separate different bins of dependencies for deployment hosts. 

Some notes: 
- We need to track the required doc updates for this. I'll do that. 
- Probably need to run this on ubuntu and centos images in CI? Or is that already happening in some capacity?

scripts/machine-launch-script.sh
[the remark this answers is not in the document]
timsnyder-siv: @davidbiancolin and @abejgonzalez, I pushed the new page in the docs.

scripts/machine-launch-script.sh
timsnyder-siv: We could, or we could leave it in there so that there is a way for people to suggest that they need a new package and we can incrementally work on getting it added into conda so that it is distro agnostic?

If we remove it, then we should rework https://github.com/firesim/firesim/pull/986/files#diff-457f33389bd30641349e0da5cbe159fde6ae8455c42a051c0f723136c08577ebR146-R148

scripts/machine-launch-script.sh
timsnyder-siv: `conda init ` either adds `/etc/profile.d/conda.sh` (the default behavior of this script when installing to `/opt/conda` as root) or to the users `~/.bashrc`.   I should note that conda also has support for other shells but I didn't introduce that in this commit.  baby steps.

By default, the code added puts `/opt/conda/condabin` into your path, makes a `conda` shell function in your shell (`conda activate` has to be a shell function to be able to manipulate the current shell environment) and it activates the `base` environment.

I was trying to do a 'drop-in' replacement of using the system packages where any user who logged into the server would have `/opt/conda/bin` in their path and we wouldn't have to ripple knowledge of `conda` further into the flow (e.g. like having `env.sh` do something conda-specific).

We may decide that we want more of the flow to need to know about conda but if you want to decouple the rest of the flow from conda, this is what I came up with as a first cut.  The only cost was installing more stuff into the `base` environment which is frowned upon because you can break conda itself when you do certain things to the `base` environment and then fixing it requires re-installing the base environment from the installer tar-script.

I actually started out installing everything into a `machine-launch` environment and then wanting to add `conda activate machine-launch` to env.sh and that leads to wanting to add another option to `build-setup.sh` where you can specify which environment you want to use.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+is that you are able to write into the install location.  See ``machine-launch-script.sh --help`` for more details.
+
+.. warning::
+
+    Without root access you can do many things, like :doc:`/Building-a-FireSim-AFI`,
+    :ref:`meta-simulation` of a FireSim system using Verilator or even developing new features in FireSim.
+    However, to :ref:`run a simulation on a F1 FPGA <running_simulations>` , FireSim currently requires that
+    you are able to act as root via ``sudo``.
+
+Updating a Package Version
```
timsnyder-siv, line 26: Need to add a section near the top of the page that explains some of the quirks with using the compiler toolchain.

You may have come to this page saying, "Where the heck is gcc?".  Well, welcome to using the conda toolchain.  Explain that activating the compilers sets the standard gnumake environment variables to direct make to the correct toolchain but if you are building something that hardcodes `gcc` instead of using `$CC`, you are going to notice it.  

Essentially give an updated version of https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html that points to the conda-forge toolchain information.

ALSO: it is possible for us to create a package that installs the short-names for the tools into /bin as symlinks to the longer tool names.  However, I dislike doing that because it's extra work and build flows shouldn't suck and hardcode toolnames in them.

docs/README.md
```diff
@@ -6,7 +6,7 @@ Building Docs
 
 Look in the `_build/html` directory for output. You can also run
 
-    python3 -m SimpleHTTPServer
+    python3 -m http.server
```
timsnyder-siv, line 9: `SimpleHTTPServer` changes to `http.server` moving to python3.  I relearned that while working on the docs.

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,271 @@
...
+    #   documentation on package_spec syntax
+    CONDA_PACKAGE_SPECS=()
+
+
+    # https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#using-the-compiler-packages
+    #    for notes on using the conda compilers
+    CONDA_PACKAGE_SPECS=( gcc_linux-64 gxx_linux-64 )
+
+    # if we want to install prebuilt riscv toolchain, the bdutrosv channel has riscv-gnu-toolchain
+
```
timsnyder-siv, line 150: I've been thinking that it would be really nice to have some more handy shell utilities installed that weren't part of what we were installing before.  My list so far of stuff I like to have around is:
* jq
* shyaml
* ripgrep
* shellcheck

But I didn't want to start throwing in more stuff that I like to have.  It's easy for me to just add them.

scripts/machine-launch-script.sh
davidbiancolin: Sorry, i wasn't clear, by explain how it simplifies it here, i meant as a comment in the code..

davidbiancolin: You could just as well link to your GH comment in your code comment though lol

timsnyder-siv: Sure.   I can do that and it will also be here in the PR! :)

timsnyder-siv:
CI is failing because it can't find `expect`.  I'm looking into it.

scripts/machine-launch-script.sh
davidbiancolin: Then i think we should maybe make the warning an error right?

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,271 @@
...
+        git  \
+        openjdk \
+        gengetopt \
+        libffi \
+        expat \
+        libusb1 \
+        ncurses \
+        cmake \
+        graphviz \
+        expect
```
timsnyder-siv, line 187: Stupid shell scripting and missing newline escapes.
```suggestion
        expect \
```

scripts/machine-launch-script.sh
timsnyder-siv:
```suggestion
	# see also: https://github.com/firesim/firesim/pull/986#discussion_r830207675
	# we default to installing into the base environment against best practice because it
```

.github/scripts/setup-manager-self-hosted.py
```diff
@@ -22,12 +22,11 @@ def initialize_manager_hosted():
...
-            with settings(warn_only=True):
-                rc = run("timeout 20m grep -q '.*machine launch script complete.*' <(tail -f machine-launchstatus)").return_code
-                if rc != 0:
-                    run("cat machine-launchstatus.log")
-                    raise Exception("machine-launch-script.sh failed to run")
+        with settings(warn_only=True):
+            rc = run("timeout 20m grep -q '.*machine launch script complete.*' <(tail -f /machine-launchstatus)").return_code
+            if rc != 0:
+                run("cat /machine-launchstatus.log")
+                raise Exception("machine-launch-script.sh failed to run")
```
timsnyder-siv, line 29: Ok.  I understand why `expect` isn't being found for the script later.  When `machine-launch-script.sh` was installing new stuff into an existing PATH directory, after it's done, this script already has it's environment configured like it needs to be.  However, now that it is creating a new PATH directory and we really want to 'conda activate' the environment (because there are also many other variables set in the environment when you activate it), stuff lower in this script doesn't see the environment it needs to.

I'm going to look into whether it is possible to activate the environment from within python somehow (it must be).  If it isn't, then @abejgonzalez, would you be opposed to me refactoring CI a bit so that this waiter happens at the end of the step that runs machine-launch-script.sh?  If we wait to start the shell until after we know machine-launchstatus.log is complete, then everything should work as expected.

timsnyder-siv: I think we can just put a precmd on the remaining commands in the script to make it work for now.  It's ugly and probably better to refactor CI so that we follow our own advice from https://github.com/firesim/firesim/blame/use_conda/docs/Initial-Setup/Setting-up-your-Manager-Instance.rst#L82-L83

timsnyder-siv: Come to think of it, what python are you running this in?  Seems like we are possibly missing a chance to dogfood our setup by using a different python before the machine is completely setup by our script in CI...

scripts/machine-launch-script.sh
timsnyder-siv: Sure.  It's probably better to force someone to at least add the new platform and have the section empty if it doesn't need to have something.

timsnyder-siv:
> Probably need to run this on ubuntu and centos images in CI? Or is that already happening in some capacity?

@davidbiancolin, from what I can see, it is running on an AWS host to set it up and not on Ubuntu.  @abejgonzalez, correct me if I'm wrong.

I was waiting to have a discussion of how much of the flow we want to test on both platforms.  I don't know how much GHA bandwidth the project has and whether you want to do everything apart form fpga-deploy on both types of host or a subset on some of the platforms...

.github/scripts/setup-manager-self-hosted.py
```diff
@@ -21,13 +36,6 @@ def initialize_manager_hosted():
 
     # Catch any exception that occurs so that we can gracefully teardown
     try:
-        # wait until machine launch is complete
-        with cd(manager_home_dir):
-            with settings(warn_only=True):
-                rc = run("timeout 20m grep -q '.*machine launch script complete.*' <(tail -f machine-launchstatus)").return_code
-                if rc != 0:
-                    run("cat machine-launchstatus.log")
-                    raise Exception("machine-launch-script.sh failed to run")
```
timsnyder-siv, line 39:
```suggestion
    with warn_only():
        run("cat /machine-launchstatus.log")
        run("env | sort")
        run("conda info")
```

I'm going to add some temporary stuff for debugging before asking to be able to ssh to the CI hosts.

.github/scripts/setup-manager-self-hosted.py
```diff
@@ -21,14 +36,11 @@ def initialize_manager_hosted():
...
-
+    with settings(warn_only=True):
+        run("cat /machine-launchstatus.log")
+        run("env | sort")
+        run("conda info")
+        run("which python")
```
timsnyder-siv, line 43:
```suggestion
        with settings(warn_only=True):
            run("cat /machine-launchstatus.log")
            run("env | sort")
            run("conda info")
            run("which python")
```

timsnyder-siv:
Well, the cloud-init environment is interesting.  I guess it doesn't have `$USER` set and that breaks my logic in machine-launch-script.sh to determine whether or not we're installing as root and the `conda init` isn't getting the `--system --no-user` options and that's why the setup isn't getting hooked into the environment (I hope).

I'm going to tweak the logic to look at `$(id -u)` but I have a fear that it will also be something weird.  I'll try this out with my own AWS instances instead of hammering on the CI.

timsnyder-siv:
Looking at some info from my own AWS host launching, I see:
```
+ env
+ sort
LANG=en_US.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
PWD=/
SHLVL=1
_=/usr/bin/env
+ id
uid=0(root) gid=0(root) groups=0(root)
```

So, the environment cloud-init runs the userdata script in is really super bare bones but `$(id -u) == 0` should tell me if we're root.

.github/scripts/setup-manager-self-hosted.py
```diff
@@ -21,14 +36,11 @@ def initialize_manager_hosted():
...
     try:
-        # wait until machine launch is complete
-        with cd(manager_home_dir):
-            with settings(warn_only=True):
-                rc = run("timeout 20m grep -q '.*machine launch script complete.*' <(tail -f machine-launchstatus)").return_code
-                if rc != 0:
-                    run("cat machine-launchstatus.log")
-                    raise Exception("machine-launch-script.sh failed to run")
-
+        with settings(warn_only=True):
```
abejgonzalez, line 39: I assume these are going to be removed once you are done testing things?

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,279 @@
...
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/pkg-specs.html#package-match-specifications
+    #   documentation on package_spec syntax
+    CONDA_PACKAGE_SPECS=()
+
+
+    # https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#using-the-compiler-packages
+    #    for notes on using the conda compilers
+    CONDA_PACKAGE_SPECS=( gcc_linux-64 gxx_linux-64 )
+
+    # if we want to install prebuilt riscv toolchain, the bdutrosv channel has riscv-gnu-toolchain
```
abejgonzalez, line 156: I think we can remove this for now and put this as an issue to track in the repo (I fear that putting this here will make us lose track of this).

scripts/machine-launch-script.sh
abejgonzalez: I'm fine adding these as long as they 1. don't take too much extra time to add and 2. they are clearly marked as optional.

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,279 @@
...
+    # Install python packages using pip that are not available from conda
+    #
+    # Installing things with pip is possible.  However, to get
+    # the most complete solution to all dependencies, you should
+    # prefer creating the environment with a single invocation of
+    # conda
+    PIP_PKGS=( \
+        fab-classic==1.19.1 \
+        mypy-boto3-ec2==1.21.9 \
+        sure==2.0.0 \
```
abejgonzalez, line 260: What is `sure` used for? This seems new.

scripts/machine-launch-script.sh
```diff
@@ -1,85 +1,279 @@
...
+    )
+    if [[ -n "$PIP_PKGS[*]" ]]; then
+        "${DRY_RUN_ECHO[@]}" $SUDO "${CONDA_PIP_EXE}" install "${PIP_PKGS[@]}"
+    fi
+
+
+    argcomplete_extra_args=()
+    if [[ "$INSTALL_TYPE" != system ]]; then
+        # if we're aren't instealling into a system directory, then initialize argcomplete
```
abejgonzalez, line 269:
```suggestion
        # if we're aren't installing into a system directory, then initialize argcomplete
```

docs/Advanced-Usage/Conda.rst
abejgonzalez, line 26: Nit: Shouldn't `moto` be in backticks everywhere (i.e "\`moto\`")

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+      graphql-core       conda-forge/noarch::graphql-core-3.2.0-pyhd8ed1ab_0
+
+    The following packages will be UPDATED:
+
+      moto                                  2.2.19-pyhd8ed1ab_0 --> 3.1.0-pyhd8ed1ab_0
+
+    Proceed ([y]/n)?
+
+
+The addition of ``graphql-core`` makes sense when looking at the `diff of moto's setup.py between
```
abejgonzalez, line 59:
```suggestion
The addition of ``graphql-core`` makes sense when looking at the diff of moto's setup.py between
```

docs/Advanced-Usage/Conda.rst
abejgonzalez: Looks like the same thing for "conda"

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+#. `The existing conda-forge packages list <feedstock-list>`_.  Keep in mind that since conda spans several domains, the
+   package name may not be exactly the same as a name from PyPI or one of the system package managers.
+#. `Adding a conda-forge recipe <https://conda-forge.org/#add_recipe>`_
+#. `PyPI <https://pypi.org/>`_ (for Python packages).  While it is possible to install packages with pip into a conda
+   environment, `there are caveats <https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html?highlight=pip#using-pip-in-an-environment>`_.
+   In short, you're less likely to create a mess if you have one system calculating dependencies and manipulating packages
+   in your environment.
+#. System packages as a last resort.  It's very difficult to have the same tools on different platforms when they are being
+   built and shipped by different systems and organizations.  That being said, in a pinch, you can find platform-specific
+   setup in ``machine-launch-script.sh``
```
abejgonzalez, line 148:
```suggestion
   setup in ``machine-launch-script.sh``.
```

abejgonzalez (commented):
Seems good to me so far. Just waiting on CI to pass for a full approval.

scripts/machine-launch-script.sh
timsnyder-siv: It is new.  I've used it pretty extensively in the pytests.  It allows you to write assertions in a nice way.  See https://sure.readthedocs.io/en/latest/ for examples.

docs/Advanced-Usage/Conda.rst
timsnyder-siv: Sure.  I'll update those.

abejgonzalez:
> > Probably need to run this on ubuntu and centos images in CI? Or is that already happening in some capacity?
> 
> @davidbiancolin, from what I can see, it is running on an AWS host to set it up and not on Ubuntu. @abejgonzalez, correct me if I'm wrong.
> 
> I was waiting to have a discussion of how much of the flow we want to test on both platforms. I don't know how much GHA bandwidth the project has and whether you want to do everything apart form fpga-deploy on both types of host or a subset on some of the platforms...

There is enough BW in GH-A to add much more CI. GH-A is really just orchestrating the CI pipelines (what runs sequentially/in-parallel). The actual work is done on the AWS EC2 manager instance that is spawned "nicely" (very debatable) in the CI and used. If we want to test the Ubuntu flow, we probably need to launch a Ubuntu AWS AMI and run the setup there. I think this is something we can do later because we need to abstract out where the driver is built from where it is run. When we figure out that we can test this a bit more.

docs/Advanced-Usage/Conda.rst
timsnyder-siv: Need that backtick for the link to render properly I think.

timsnyder-siv: Maybe I should join line 60 back into 59 so it is easier to tell that it is a link to a specific diff on GH.

.github/scripts/setup-manager-self-hosted.py
timsnyder-siv: Yes.  These will get removed.

scripts/machine-launch-script.sh
timsnyder-siv:
```suggestion
```
https://github.com/firesim/firesim/issues/993

scripts/machine-launch-script.sh
timsnyder-siv: I think until and unless we're using them in scripts, I'll keep them out of the script.

timsnyder-siv:
Well, now we're past 'setup-manager' at least!  
run-test-groupA is failing because Verilator binary segfaults:
```
/bin/sh: line 2: 11081 Segmentation fault      ./VResetPulseBridgeTest +reset-pulse-length0=1023 +dramsim +waveform=/home/centos/firesim/sim/output/f1/ResetPulseBridgeTest-ResetPulseBridgeActiveLowConfig-NoSynthAsserts_HostDebugFeatures_DefaultF1Config/ResetPulseBridgeTest.verilator.vcd 2> /home/centos/firesim/sim/output/f1/ResetPulseBridgeTest-ResetPulseBridgeActiveLowConfig-NoSynthAsserts_HostDebugFeatures_DefaultF1Config/ResetPulseBridgeTest.verilator.out
```
I suspect that something wonky is hapening either during the verilator build for firesim or possibly is wrong in my verilator package.  I'll try to reproduce this locally and dig into it.

The pytest failure is known and I'll go ahead and fix it in this PR.

I don't completely understand why build-default-workloads is failing with br-base.json not being found.

timsnyder-siv:
The verilator segfault is being caused by one of the additional CFLAGS, CXXFLAGS or LDFLAGS.  When I manually edited the verilator makefile to take them out (and also unset them in my environment), rebuilt and got one of the segfaulting verilator sims to pass.

I'll figure out a good way to avoid this problem in the morning.

timsnyder-siv:
It's an ordering thing.  If I put all of the flags back into the Vverilated_top.mk and don't have the conda CFLAGS, CXXFLAGS and LDFLAGS in my environment, it will build and run fine.  If I put those back into my environment, rebuild and then it segfaults.

timsnyder-siv:
ci failed last night for some mysterious reason (log stopped in the middle of the clone of firesim repo) and rerunning doesn't work because the directory already exists and git clone doesn't like that.  Going to close and reopen in the hopes that will give me a fresh setup for CI.

timsnyder-siv:
Marking DONOTMERGE because I"m going to put a `git cherry-pick` in firemarshal's `init-submodules.sh` in CI until we determine how we want to do the fix in a mergeable way.

scripts/machine-launch-script.sh
```diff
@@ -1,82 +1,311 @@
...
+    gcc_version=$($CONDA_ENV_BIN/gcc --version | head -n1 | awk '{print $NF}')
+    CONDA_ENV_PREFIX=$(readlink -e $CONDA_ENV_BIN/..)
+    specdir=$CONDA_ENV_PREFIX/lib/gcc/$gcc_machine/$gcc_version
+
+    "${DRY_RUN_ECHO[@]}" $SUDO sed -i -e "/\*link_command:/,+1 s+%.*+& %{\!static:-rpath ${CONDA_ENV_PREFIX}/lib -rpath-link ${CONDA_ENV_PREFIX}/lib -disable-new-dtags} -L ${CONDA_ENV_PREFIX}/lib+" $specdir/specs
+    # use -idirafter to put the conda "system" includes where /usr/local/include would typically go
+    # in a system-packaged non-cross compiler
+    "${DRY_RUN_ECHO[@]}" $SUDO sed -i -e "/\*cpp_options:/,+1 s+%.*+& -idirafter ${CONDA_ENV_PREFIX}/include+" $specdir/specs
+    # cc1_options also get used for cc1plus... at least in 11.2.0
+    "${DRY_RUN_ECHO[@]}" $SUDO sed -i -e "/\*cc1_options:/,+1 s+%.*+& -idirafter ${CONDA_ENV_PREFIX}/include+" $specdir/specs
```
timsnyder-siv, line 305: I would ideally like to be able to remove this as well because it hacks the specfile installed by conda gcc.

scripts/machine-launch-script.sh
davidbiancolin: My preference is to limit this to only required dependencies, but if they are clearly labelled as @abejgonzalez suggested, it's fine.

docs/Advanced-Usage/Conda.rst
timsnyder-siv: I actually switched to doing this.  The conda-forge guys actually started making a package that is just the symlinks of the shortnames to the longnames and I'm adding the minimal set of options you have to have to make it act like a system-packaged toolchain in a PR that should be merged soon.  FYI, it's the specfile hacking that I currently have at the end of machine-launch-script.sh

timsnyder-siv: I think I got all these after Abe suggested it.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+    However, it is always a better idea to modify the version in ``machine-launch-script.sh`` so that:
+    #. you remember to commit and share the new version requirement.
+    #. you are providing a complete set of requirements for ``conda`` to solve
+
+So, modify ``machine-launch-script.sh`` with the updated version of ``moto``, and run it.  If you'd like to see what
+it will do first, feel free to give it the ``--dry-run`` option, look at the output and then run again leaving ``--dry-run``
+off.
+
+In this case, when you are finished, you can run ``conda list --revisions`` and you should see output
+like the following:
```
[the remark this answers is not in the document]
timsnyder-siv: sweet.  I'll switch to that.

docs/Advanced-Usage/Conda.rst
timsnyder-siv: b3d7ff69ee6c4be23e1b23570c9b216c8af5bec4 is the commit where I switched to using the package that installs the shortnames instead of the one that uses the activation scripts with all of the default options for building packages shoved into FLAGS env vars.

timsnyder-siv:
@sagark , your question in the meeting made me wonder if there was a good report that can be generated with a tool about all the licenses in the conda enviornment.  I don't see something existing but the information is readily available:
```
-bash-4.2$ grep license_family *.json | awk '{print $NF}' | sort | uniq -c
      1 "AGPL",
     19 "Apache",
     16 "APACHE",
     68 "BSD",
      1 "CC",
     44 "GPL",
      2 "GPL2",
      1 "GPL3",
      9 "LGPL",
      2 "lib/python3.8/site-packages/conda_build/license_family.py",
      2 "lib/python3.8/site-packages/conda_build/__pycache__/license_family.cpython-38.pyc",
     91 "MIT",
      3 "MOZILLA",
      6 "Other",
      4 "PSF",
```
From inside the $PREFIX/conda-meta directory (which contains a JSON for every installed package).  This is for everything that machine_launch.sh installs (plus some extra stuff that I've installed by hand).

scripts/machine-launch-script.sh
```diff
@@ -1,82 +1,311 @@
...
+        moto==3.1.0 \
+        pyyaml==5.4.1 \
+        mypy==0.931 \
+        types-pyyaml==6.0.4 \
+        boto3-stubs==1.21.6 \
+        botocore-stubs==1.24.7 \
+        mypy-boto3-s3==1.21.0 \
+    )
+
+    if [[ "$CONDA_ENV_NAME" == "base" ]]; then
```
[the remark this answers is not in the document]
timsnyder-siv: I've made changes so that the script no longer installs the firesim deps into the base environment but instead defaults to creating a 'firesim' environment.  I'll be pushing that along with come cleanup of the docs.  

I've also added a `--reinstall-conda` option that people can use to reinstall the base environment if it gets broken.

.github/scripts/initialize-manager.py
```diff
@@ -39,6 +39,9 @@ def initialize_manager(max_runtime):
         # run concurrently with ./init-submodules.sh
         with cd(manager_marshal_dir):
             run("./init-submodules.sh")
+            # WIP: patch buildroot to build cleanly with make 4.3, otherwise, it will incorrectly
+            # error out in the first 'make defconfig'
+            run("git -C boards/default/distros/br/buildroot/ cherry-pick -x 9e2128bf5072e5f2fd69e2fc0239558782dfc677")
```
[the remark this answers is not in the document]
timsnyder-siv: I've also decided to pin `make` to not use 4.3 until we get firemarshal to bump buildroot.  That way we can decouple that change from merging this PR.

deploy/runtools/runtime_config.py
```diff
@@ -65,13 +69,374 @@ def get_local_driver_path(self):
...
+            if 'linux-vdso.so' in dso.soname:
+                continue
+            assert dso.path is not None, f"{dso.soname} has no linkage reported by ldd for {exe} in: {dso}"
+            if dso.path in glibc_shared_libs:
+                continue
+
+            # we do some moderate evilness here and collapse
+            # the version symlinks of our dso into the actual dso
+            # so that we have fewer pieces to copy over
+            libs.append((realpath(dso.path), dso.soname))
```
timsnyder-siv, line 437: @davidbiancolin this is the code that does the introspecting via `ldd` to figure out what to copy.  The contents of glibc could be derived dynamically by running what I documented on line 91 but I was reluctant to put rpm-specific stuff in the manager.  

Like we discussed, I'll post an alternate PR that accomplishes the same thing in a more portable way using `conda-build` and `constructor`.

timsnyder-siv:
I haven't been able to reproduce the latest failure in building the docs, even when I install things into the system python36 site-packages with pip like we're doing in the CI.  I'm going to wait for linux-poweroff to finish and then see if I can rerun only the docs part of the workflow.  I also have a commit that dumps all of the objects.inv so we can see what does show up in there because I see `meta-simulation` as a thing that can be referred to.

timsnyder-siv:
Ok.  So, the most recent failure in linux-poweroff https://github.com/firesim/firesim/runs/5826125643?check_suite_focus=true#step:3:1112 is happening in the cluster sim and it looks like the switch is stopped and all of the nodes are sitting at Networking started waiting for their DHCP or something.  Looking at the switchlog, I don't see any indication as to why it exited, nor do I see how it was started by the manager by looking in it's log.  I'm going to dig into this tomorrow.  I suspected that it would be a linking problem when it started up because we're not copying any DSOs over for the switch, however it seems to start and spit out stuff into it's 
[switchlog.txt](https://github.com/firesim/firesim/files/8414840/switchlog.txt)

If @sagark or someone who knows more about the switch could look at that log or suggest what to look at, I'd appreciate it.  I'm sure as I look closer, it will become apparent and I'll figure out the cmdline to relaunch the switch tomrorow.

timsnyder-siv:
I'm fairly confident that the switch dying right away is because we compile and link against newer libgomp provided by the conda environment but then link&load the centos7 libgomp.  I've done some looking to see if I can improve the logging to indicate why the switch died.  It's interesting that when I manually logged into the `f1.16xlarge` and manually restarted the switch, it kept running and the simulations advanced and finished.

Looking in `/var/log/messages` I was able to identify messages associated with the switch crash:
```
Apr  5 15:34:34 ip-192-168-2-160 kernel: switch0[14223]: segfault at 7fbd928cfff8 ip 0000564d397fd371 sp 00007fbd928d0000 error 6
Apr  5 15:34:34 ip-192-168-2-160 kernel: switch0[14222]: segfault at 7fbd930d0ff8 ip 0000564d397fd371 sp 00007fbd930d1000 error 6
Apr  5 15:34:34 ip-192-168-2-160 kernel: switch0[14225]: segfault at 7fbd918cdff8 ip 0000564d397fd371 sp 00007fbd918ce000 error 6
Apr  5 15:34:34 ip-192-168-2-160 kernel: in switch0[564d397fd000+4000]
Apr  5 15:34:34 ip-192-168-2-160 kernel: 
Apr  5 15:34:34 ip-192-168-2-160 kernel: in switch0[564d397fd000+4000]
Apr  5 15:34:34 ip-192-168-2-160 kernel: 
Apr  5 15:34:34 ip-192-168-2-160 kernel: switch0[14227]: segfault at 7fbd908cbff8 ip 0000564d397fd371 sp 00007fbd908cc000 error 6
Apr  5 15:34:34 ip-192-168-2-160 kernel: 
Apr  5 15:34:34 ip-192-168-2-160 kernel: switch0[14224]: segfault at 7fbd920ceff8 ip 0000564d397fd371 sp 00007fbd920cf000 error 6
Apr  5 15:34:34 ip-192-168-2-160 kernel: 
Apr  5 15:34:34 ip-192-168-2-160 kernel: in switch0[564d397fd000+4000]
Apr  5 15:34:34 ip-192-168-2-160 kernel: 
```

@sagark, is it a "feature" that the manager continues waiting if a switch is found to be not running?  I would expect that you would want the manager to tear things down and treat the simulation as if it failed but maybe this happens sometimes and you typically ssh to the host and restart the switch to keep things going?

timsnyder-siv:
@sagark, if it is a feature, should the manager be taught to restart the switches if it finds them not running?  I'm not saying we should do it in this PR but possibly capture in an issue for the future?

timsnyder-siv:
I was able to get some more info about the segfault by running under gdb (right before I fixed the library copying):
```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff3b9c700 (LWP 116993)]
0x0000555555556371 in BasePort::write_flits_to_output (this=0x555555560a30)
    at baseport.h:134
134     baseport.h: No such file or directory.
Missing separate debuginfos, use: debuginfo-install glibc-2.17-325.el7_9.x86_64 libgcc-4.8.5-44.el7.x86_64 libgomp-4.8.5-44.el7.x86_64 libstdc++-4.8.5-44.el7.x86_64
(gdb) w
Ambiguous command "w": watch, wh, whatis, where, while, while-stepping, winheight, ws.
(gdb) where
#0  0x0000555555556371 in BasePort::write_flits_to_output (this=0x555555560a30)
    at baseport.h:134
#1  0x0000555555556376 in BasePort::write_flits_to_output (
    this=0x555555556376 <BasePort::BasePort(int, bool)>) at baseport.h:134
```
NOTE: the stack was super deep.  I paged through a few thousand recursive calls of this same function before giving up.

timsnyder-siv:
Hmm.  Now that I have the libraries being copied over, I'm still hitting the segfault, at least sometimes.

timsnyder-siv:
I remembered how to turn off the gdb pagination and dumped the full stacktrace.  The end of it looks like:
```
#134593 0x0000555555556376 in BasePort::write_flits_to_output (this=0x555555556376 <BasePort::BasePort(int, bool)>) at baseport.h:134
        timestampdiff = <optimized out>
        i = <optimized out>
        thispacket = 0x555555556376 <BasePort::BasePort(int, bool)>
        space_available = <optimized out>
        outputtimestamp = <optimized out>
        outputtimestampend = <optimized out>
        flitswritten = 93824992240502
        basetime = 93824992240502
        maxtime = 93824992240502
#134594 0x0000555555556376 in BasePort::write_flits_to_output (this=0x555555556376 <BasePort::BasePort(int, bool)>) at baseport.h:134
        timestampdiff = <optimized out>
        i = <optimized out>
        thispacket = 0x555555556376 <BasePort::BasePort(int, bool)>
        space_available = <optimized out>
        outputtimestamp = <optimized out>
        outputtimestampend = <optimized out>
        flitswritten = 93824992240502
        basetime = 93824992240502
        maxtime = 93824992240502
#134595 0x000055555555744c in do_fast_switching () at switch.cc:196
        thisport = <optimized out>
#134596 0x00007ffff7e1d4a8 in GOMP_parallel () from /home/centos/switch_slot_0/libgomp.so.1
No symbol table info available.
#134597 0x00005555555587e5 in do_fast_switching() () at switch.cc:193
        pqueue = {c = {<std::_Vector_base<do_fast_switching()::tspacket, std::allocator<do_fast_switching()::tspacket> >> = {_M_impl = {<std::allocator<do_fast_switching()::tspacket>> = {<__gnu_cxx::new_allocator<do_fast_switching()::tspacket>> = {<No data fields>}, <No data fields>}, <std::_Vector_base<do_fast_switching()::tspacket, std::allocator<do_fast_switching()::tspacket> >::_Vector_impl_data> = {_M_start = 0x555555574a20, _M_finish = <optimized out>, _M_end_of_storage = <optimized out>}, <No data fields>}}, <No data fields>}, comp = {<std::binary_function<do_fast_switching()::tspacket, do_fast_switching()::tspacket, bool>> = {<No data fields>}, <No data fields>}}
#134598 0x0000555555556967 in main () at switch.cc:269
        bandwidth = <optimized out>
#134599 0x00007ffff72f4555 in __libc_start_main (main=0x555555556670 <main>, argc=4, argv=0x7fffffffe638, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe628) at ../csu/libc-start.c:266
        result = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {0, 5590096846624731810, 93824992242160, 140737488348720, 0, 0, -5590096846407519582, -5590113754634046814}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
#134600 0x0000555555556a19 in _start () at switch.cc:241
```

timsnyder-siv:
Confirmed that rebuilding libgomp with `--disable-linux-futex` and using `-fsanitize=thread` when building the switch didn't cause any amazing warnings to jump out and spot the issue.

I'm also pretty sure that the switch works the second time I run it if I don't `sudo rm -rf /dev/shm/*` but if I do remove the shared memory pages, it doesn't work the second time either.

With `-O0` I didn't hit the segfault but both `-O2` and `-O3` do hit it.

sagark:
> @sagark, if it is a feature, should the manager be taught to restart the switches if it finds them not running? I'm not saying we should do it in this PR but possibly capture in an issue for the future?

I don't think we should do this, since restarting a switch that crashed probably means you lost tokens and the sim is no longer cycle-accurate. I think tearing down and reporting an error in this situation is fine (maybe with a --debug flag that does the opposite, for debugging).

deploy/runtools/runtime_config.py
timsnyder-siv: As an alternative to excluding the contents of glibc, we could look at the RPATH of the executable and only copy things that are coming from the RPATH (and whatever LD_LIBRARY_PATH is set to until we no longer need it)

docs/Advanced-Usage/Conda.rst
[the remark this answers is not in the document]
timsnyder-siv: done.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+and adds the required setup to the system-wide ``/etc/profile.d/conda.sh`` init script to add
+``/opt/conda`` to everyone's path.
+
+However, the script is also flexible.  For example, if you do not have root access, you can specify
+an alternate install location with the ``--prefix`` option to ``machine-launch-script.sh``.  The only requirement
+is that you are able to write into the install location.  See ``machine-launch-script.sh --help`` for more details.
+
+.. warning::
+
+    Without root access you can do many things, like :doc:`/Building-a-FireSim-AFI`,
```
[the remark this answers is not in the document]
timsnyder-siv: done.

docs/Advanced-Usage/Conda.rst
timsnyder-siv: Didn't need to explain a bunch of details about the compilers because we switched away from using the activation scripts and it is more straightforward.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,183 @@
...
+
+If you find that a package is missing an optional feature, consider looking up it's 'feedstock' (aka recipe) repo in
+`The existing conda-forge packages list <feedstock-list>`_.  and submitting an issue or PR.
+
+If you instead need to enable debugging or possibly actively hack on the source of something we package:
+
+#. Find the feedstock repo in the `feedstock-list`_
+#. Clone the feedstock repo and modify ``recipe/build.sh`` (or ``recipe/meta.yaml`` if there isn't a build script)
+#. ``python build-locally.py`` to `build using the conda-forge docker container <https://conda-forge.org/docs/maintainer/updating_pkgs.html#testing-changes-locally>`_
+   If the build is successful, you will have an installable ``conda`` package in ``build_artifacts/linux-64`` that can be
```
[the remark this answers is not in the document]
timsnyder-siv: I added the stuff about --debug.  I'll worry about multiple outputs if it becomes an issue worth documenting.

target-design/switch/flit.h
```diff
@@ -49,7 +49,7 @@ void write_valid_flit(uint8_t * send_buf, int tokenid) {
     *lrv |= (1L << bitoffset);
 }
 
-int write_last_flit(uint8_t * send_buf, int tokenid, int is_last) {
+void write_last_flit(uint8_t * send_buf, int tokenid, int is_last) {
```
timsnyder-siv, line 52: @sagark , this is what was causing the switch to segfault.  Don't ask me how this causes it, I just know that when I fixed the return type of this one function, it stops happening (at least the handful of times I've run it).  I have poweroff running in a loop for awhile on my EC2.  After it does it without the switch dying a few more times, I'll run iperf3 like we talked about.

deploy/runtools/runtime_config.py
timsnyder-siv: This code moved into `deploy/runtools/utils.py`.  I think it is fine to merge as is for now and possibly revisit in a future PR to simplify it.

timsnyder-siv:
@abejgonzalez, @sagark, @davidbiancolin this is finally ready to merge.  Please give it a final lookthrough and approve.

timsnyder-siv:
I was able to `deploy/workloads/run-bw-test.sh` and get it to run through all of the bandwidths without any of the switches crapping out.  However, it seems like some of the math in the plotting script or the configuration of something is out of sync because each series in the plot gets a single point.  I'm going to call that not related to this PR and move forward.  Actually, it's the `\r` in https://github.com/firesim/firesim/blob/7616ae5b1e34c04d490af66ae81d76bedf0f2850/deploy/workloads/bw-test-two-instances/bw-test-graph.py#L9 

Not sure why the pattern had the carriage return at the end of it but in python3, that doesn't match any of the lines in the log but taking it out, matches like it should and I get a graph that looks similar to fig 6 of https://sagark.org/assets/pubs/firesim-isca2018.pdf

image

target-design/switch/flit.h
sagark: oh wow, thanks for figuring this out!

sagark: oh wow, thanks for figuring this out!

.github/scripts/build-default-workloads.py
```diff
@@ -13,8 +13,9 @@ def build_default_workloads():
         # avoid logging excessive amounts to prevent GH-A masking secrets (which slows down log output)
         with settings(warn_only=True):
             rc = run("marshal -v build br-base.json &> br-base.full.log").return_code
+            print(f"marshal exit code was '{rc}'")
+            run("cat br-base.full.log")
```
sagark, line 17: Is this change supposed to be included? In that case, the comment above it is no longer correct.

timsnyder-siv: No, I intended to revert that after fixing it and also addressing https://github.com/firesim/FireMarshal/issues/233 Thanks for pointing out.  I'll revert now.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,224 @@
...
+    +zipp-3.7.0 (conda-forge/noarch)
+    +zlib-1.2.11 (conda-forge/linux-64)
+    +zstd-1.5.2 (conda-forge/linux-64)
+
+    2022-03-15 19:34:06  (rev 1)
+         moto  {2.2.19 (conda-forge/noarch) -> 3.1.0 (conda-forge/noarch)}
+
+This shows you that the first time ``machine-launch-script.sh`` was run, it created 'revision' 0 of the environment with
+many packages.  After updating the version of ``moto`` and rerunning, 'revision' 1 was created by updating the version
+of ``moto``.  At any time, you can revert your conda environment back to an older 'revision' using ``conda install -revision <n>``
```
sagark, line 103: This is awesome!

.github/scripts/build-default-workloads.py
timsnyder-siv: I actually decided to wait and see if there is more feedback I need to apply before rerunning CI.

scripts/build-libdwarf.sh
```diff
@@ -9,11 +9,14 @@ fi
 
 prefix=$1
```
sagark, line 11: Does conda have pre-built libdwarf and libelf we could use instead now?

timsnyder-siv: They have elfutils and I started creating a libdwarf recipe.  I had a question for you.  Latest libdwarf has an option to be built without libelf.  Are you fully using both or could we transition to using only latest libdwarf?

timsnyder-siv: Also, if we can use latest libdwarf, can we bump to using that along with switching or would you want me to build the current version first?  It's not a huge deal to build the old one first but if we can bump to the latest, its a little less work.

timsnyder-siv: https://github.com/davea42/libdwarf-code/releases/tag/libdwarf-0.1.1 is the version where libdwarf stopped needing libelf.  Unfortunately the github tag comments aren't as insightful as what he posts in his website at https://www.prevanders.net/dwarf.html#releases

docs/Initial-Setup/Setting-up-your-Manager-Instance.rst
```diff
@@ -71,7 +71,7 @@ before, so we need to first ssh into the instance and make sure the
 setup is complete.
 
 In either case, ``ssh`` into your instance (e.g. ``ssh -i firesim.pem centos@YOUR_INSTANCE_IP``) and wait until the
-``~/machine-launchstatus`` file contains all the following text:
+``/machine-launchstatus`` file contains all the following text:
```
sagark, line 74: Is the `machine-launchstatus` file at `/` now?

timsnyder-siv: Yes.  It's the fallout of just doing https://github.com/firesim/firesim/pull/932#discussion_r809629052

scripts/build-libdwarf.sh
sagark: I looked at the `tracerv` code again and it looks like we use both.

I think it's probably fine to just bump it to latest? @a0u any input?

docs/Initial-Setup/Setting-up-your-Manager-Instance.rst
timsnyder-siv: Because it was previously hardcoded to put it in `/home/centos` and that won't exist when not on an Centos VM

scripts/build-libdwarf.sh
a0u: Bumping libdwarf should be fine.  The current version was just the most recent when we started development.

And to confirm, we do use both libelf and libdwarf.  The stack reconstructor falls back to the ELF symbol table if DWARF debugging info is unavailable.

timsnyder-siv: Ok.  I'll finish submitting a PR to conda-forge to add libdwarf latest and we can do the bump in a follow-on.

docs/Advanced-Usage/Conda.rst
```diff
@@ -0,0 +1,224 @@
...
+an alternate install location with the ``--prefix`` option to ``machine-launch-script.sh``.  The only requirement
+is that you are able to write into the install location.  See ``machine-launch-script.sh --help`` for more details.
+
+.. warning::
+
+    To :ref:`run a simulation on a F1 FPGA <running_simulations>` , FireSim currently requires that
+    you are able to act as root via ``sudo``.
+
+    However, you can do many things without having root, like :doc:`/Building-a-FireSim-AFI`,
+    `<meta-simulation>`_ of a FireSim system using Verilator or even developing new features in FireSim.
```
davidbiancolin, line 26: You may have to remove the hyphen here now for this ref to resolve.

docs/Advanced-Usage/Conda.rst
davidbiancolin: I wonder if it makes sense to "tag", in some capacity, the revision with the sha that produced it, but this is great.

davidbiancolin (approved):
I don't have much new to add. The `so` filtering is mildly scary to me but that is probably rooted in my own ignorance. If it looks alright to @sagark we should ship it.

timsnyder-siv:
>  The so filtering is mildly scary to me

It is to me as well.  It's likely to be a bit brittle.  The way I'd really like to fix it requires getting libdwarf and libelf to be coming from conda packages and I think this will work for now.  (and I have libdwarf on deck for review to be added to conda-forge in https://github.com/conda-forge/staged-recipes/pull/18638)

scripts/build-libdwarf.sh
timsnyder-siv: https://github.com/conda-forge/staged-recipes/pull/18638 is the PR to add libdwarf to conda-forge.  Once it has merged, I'll come back through and start a PR that bumps them and switches over to using the conda packages.

.github/scripts/build-default-workloads.py
timsnyder-siv: I reverted this in my latest push.  @sagark , can you take another look through and approve?

timsnyder-siv:
I'm a tad hesitant to bump Chipyard yet again for their oneline change to invalidate the CI tool cache in https://github.com/ucb-bar/chipyard/pull/1140/commits/2b6e1362dfd59341e1b69ac9cb36ba6be0e018c1

davidbiancolin:
> I'm a tad hesitant to bump Chipyard yet again for their oneline change to invalidate the CI tool cache in [ucb-bar/chipyard@2b6e136](https://github.com/ucb-bar/chipyard/commit/2b6e1362dfd59341e1b69ac9cb36ba6be0e018c1)

Don't worry about it.

sagark (approved):
LGTM, this is great!

scripts/build-libelf.sh
```diff
@@ -8,10 +8,14 @@ fi
...
+export CFLAGS+=" -Wno-unused-variable"
+# leave the assertions in elfutils
+export CPPFLAGS="${CPPFLAGS/-DNDEBUG/}"
+
 cd sim/firesim-lib/src/main/cc/lib/elfutils
 test -f configure || autoreconf -i -f
 ./configure --prefix="${prefix}" --enable-maintainer-mode
 make
-make install
+make install
```
abejgonzalez, line 19:
```suggestion
make install
```

abejgonzalez (approved):
Small nits but overall looks great!

Issue #968, opened 2022-02-27
Bump Python to 3.7+ and moto to 3.0.4+


timsnyder-siv:
* #940 added tests to test_awstools.py that depend on newer moto and that requires newer Python
* #850 introduces weird import for typing in buildconfig.py that can be simplified in Python 3.7+

timsnyder-siv:
I'm nervous about possible breakages in fab-classic when moving to 3.8+ @abejgonzalez, fyi https://github.com/ploxiln/fab-classic/issues/27

PR #991, opened 2022-03-17
Use pagination for boto3 calls in the manager

sagark:
Several boto3 APIs require pagination, which we have been ignoring, this fixes that:

https://boto3.amazonaws.com/v1/documentation/api/1.20.21/guide/paginators.html

This rarely seems to happen in practice for our use cases, but we did see it once during the tutorial setup for AGFIs.

Also, this updates the relevant tests in deploy/tests/ and bumps moto to 3.1.0 which enables running all of the tests. Of those tests, 9 passed and 1 xpassed (confirmed as expected result with @timsnyder).

#### Related PRs / Issues

Partially addresses #968 (moto3 bump).

There's no issue for the actual problem this fixes.

#### UI / API Impact

None.

No change.

### Contributor Checklist

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -28,6 +28,14 @@ rootLogger = logging.getLogger()
 # by running scripts/update_test_amis.py
 f1_ami_name = "FPGA Developer AMI - 1.11.1-40257ab5-6688-4c95-97d1-e251a40fd1fc"
 
+def depaginated_boto_query(client, operation, operation_params, return_key):
+    paginator = client.get_paginator(operation)
+    page_iterator = paginator.paginate(**operation_params)
+    return_values_all = []
+    for page in page_iterator:
+        return_values_all += page[return_key]
+    return return_values_all
+
 def aws_resource_names():
     """ Get names for various aws resources the manager relies on. For example:
     vpcname, securitygroupname, keyname, etc.
@@ -78,8 +86,8 @@ def aws_resource_names():
         instanceid = res.stdout
         rootLogger.debug(instanceid)
 
-        resp = client.describe_tags(
-            Filters=[
+        operation_params = {
+            'Filters': [
                 {
                     'Name': 'resource-id',
                     'Values': [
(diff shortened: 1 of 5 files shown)


deploy/awstools/awstools.py
```diff
@@ -78,21 +86,22 @@ def aws_resource_names():
...
+        }
+        resp_pairs = depaginated_boto_query(client, 'describe_tags', operation_params, 'Tags')
     except client.exceptions.ClientError:
         return base_dict
 
     resptags = {}
-    for pair in resp['Tags']:
+    for pair in resp_pairs:
         resptags[pair['Key']] = pair['Value']
     rootLogger.debug(resptags)
```
timsnyder-siv, line 107: @sagark , I thought the thing that was failing in the tutorial was the `describe_images` call below here in `get_f1_ami_id`.  I really dislike how GH tries to keep me from reading and commenting on code that is too far from one of your changes but I'm pretty sure that needs to be updated too, right?

Even if we're fairly confident that the server-side filtering should prevent pagenation, we should still probably handle it.

timsnyder-siv (commented):
@sagark , do you think we should depaginate the `describe_images` in get_f1_ami_id?

deploy/awstools/awstools.py
abejgonzalez: Separate from functionality, I think it also makes sense to change any client call to the paginated version anyways.

abejgonzalez (commented):
Other than the `get_f1_ami_id` call, I think this looks good.

deploy/awstools/awstools.py
sagark: The problem at the tutorial was with `describe_fpga_images`.

The `describe_images` API doesn't use pagination (note that there's no `NextToken` or `MaxResults` param like the other calls have):
https://boto3.amazonaws.com/v1/documentation/api/1.20.21/reference/services/ec2.html?highlight=describe_images#EC2.Client.describe_images

To double check, I tried to create a paginator for describe_images and I get an `Operation cannot be paginated` error.

sagark:
@timsnyder-siv @abejgonzalez This should be ready for approval, see note above about describe_images not supporting pagination.

timsnyder-siv (approved):
@sagark makes sense. I guess they haven't read the orthogonal design paper

:shipit:

abejgonzalez (approved):

PR #996, opened 2022-03-24
Collect Bridge Stream RTL Implementation under StreamEngine module

davidbiancolin:
To support streams under Alveo i plan to break my PRs in to the following pieces:

1. Push the existing FPGA-side implementation of bridge streams into a module class hierarchy that can support different implementations (this PR)
  - Description to follow.
2. Push the existing driver-side implementation of bridge streams behind `simif_t`
  - At this point, bridges will use the proposed `push`, `pull` and `flush` API calls we have brainstormed
  - Bridge code will make no MMIO calls to service their streams (which exposes the underlying implementation)
3. Provide an implementation of `FPGAManagedStreamEngine` that works under metasimulation for toCPU streams
4. Provide remaining U250 (simif_t, PlatformShim, Kernel / Vitis side junk)
 
The general theme of the first two PRs revolve around moving the _implementation_ of bridge streams out of bridges themselves -- this PR does that for the RTL.  Bridges connect to a `StreamEngine` using `DecoupledIO` interfaces with 512b UInt payloads. Bridges request streams by mixing in traits defined in UsesBridgeStreams.scala. A `StreamEngine` binds to all of these 512b channels and optionally to an AXI4M and/or AXI4S graph based on its implementation. Host platform shim writers select an Engine implementation by using the `StreamEngineInstantiatorKey`. 

Driver-side changes are temporary, since the push + pull API will be implemented in the next PR and the drivers will not be aware of the Stream-related MMIO addresses. 

I've elected to keep the engine itself under FPGATop so it will be included in a conventional metasimulation. This means the simif for metasimulations will need to support multiple `StreamEngine` implementations, but i don't think our FPGA-level simulation flows are good enough to justify pushing it into the `PlatformShim`. 
- Alternatively one could imagine doing metasimulation at the DecoupledIO boundary of the streams -- it'd be faster, and would require no changes to support other StreamEngine implementations. 

Other changes:
- Replaces Nasti-based implementation of PCIS with a diplomatic AXI4-based implementation. Note: the StreamEngine is the only slave / master currently, but it would be straightforward to use it to drive the DRAM channels in the future.
- Implements count in BRAMQueues (subsumes the counters used in the old DMA transport)
- Adds a simulator memory map summary. 

TODO:
- [x] Check URAM constraints.

 #### Related PRs / Issues

#### UI / API Impact

Users must now mix in `HasToHostCPUStream` and `HasFromHostCPUStream` into their `BridgeModules` (not the implementation like previously).
- outgoingPCISdat.io.enq is renamed streamEnq
- incomingPCISdat.io.deq is renamed streamDeq

Changes to the BridgeDrivers is temporary, and should not be scrutinized to carefully. 

Will need regeneration.

### Contributor Checklist


sim/midas/src/main/scala/midas/core/CPUManagedStreamEngine.scala
```diff
@@ -0,0 +1,213 @@
...
+    dma.b.bits.user := dma.aw.bits.user
+    // This will be set by the channel given the grant using last connect semantics
+    dma.b.valid := false.B
+    dma.aw.ready := false.B
+    dma.w.ready := false.B
+
+
+    // TODO: Chisel naming prefix to indicate what channel this hw belongs to.
+    // This demultiplexes the AW, W, and B channels onto the decoupled ports representing each stream.
+    def elaborateFromHostCPUStream(
```
davidbiancolin, line 79: The bodies of these functions are more or less copied directly from the old Implementation.

sim/midas/src/main/scala/midas/core/StreamEngine.scala
```diff
@@ -0,0 +1,78 @@
...
+  (e: StreamEngineParameters, p: Parameters) => new CPUManagedStreamEngine(p, e))
+
+
+/**
+  * Parameters that define a stream that are defined by bridges and collected
+  * by the engine.
+  */
+trait StreamParameters {
+  def name: String
+  def fpgaBufferDepth: Int
```
davidbiancolin, line 29: Perhaps a good name for this that might be implementation agnostic: `minimumQueueCapacity` which could then be set to the batch size the driver expects to use.

davidbiancolin:
I guess there's no strong reason this couldn't be made against main.

sim/midas/src/main/scala/midas/widgets/Widget.scala
```diff
@@ -251,6 +251,18 @@ trait HasWidgets {
...
+    println("Simulator Memory Map:")
+    for (AddrMapEntry(name, MemRange(start, size, _)) <- addrMap.flatten) {
+      println(f"  [${start}%4h, ${start + size - 1}%4h]: ${name}")
+    }
+  }
+
+  /**
+    * Iterates through each bridge, generating the header fragment. Must be
+    * called after bridge address assignment is complete.
+    *
```
abejgonzalez, line 264:
```suggestion
```

sim/midas/src/main/scala/midas/widgets/UsesBridgeStreams.scala
```diff
@@ -0,0 +1,82 @@
...
+  * which the driver is the producer and the BridgeModule is the consumer,
+  * [[StreamToHostCPU]] does the converse. BridgeModules can mix in one or both
+  * of these traits, to implement streams in either direction.
+  *
+  * Limitations:
+  * - Streams are 512b wide. Bridge modules and drivers must manually handle
+  *   width adaptations
+  * - Bridges are limited to one stream in each direction. Bridge designers
+  *   must multiplex multiple streams onto a single physical stream themselves.
+  * ,.
```
abejgonzalez, line 28:
```suggestion
```

sim/midas/src/main/scala/midas/widgets/UsesBridgeStreams.scala
```diff
@@ -0,0 +1,82 @@
...
+  */
+
+/**
+  * Constants defined here apply to interfaces exposed directly to the bridges
+  * and thus the user.
+  */
+object BridgeStreamConstants {
+  // The width of the decoupled UInt payload presented to the bridge.
```
abejgonzalez, line 36: Probably needs "in bits" here.

abejgonzalez (commented):
This API seems fine to me (ignoring some of the more detailed changes)

sim/midas/src/main/scala/midas/core/StreamEngine.scala
davidbiancolin: better: desiredBatchDepth (in 512b chunks) or desiredBatchBytes? 

Unfortunately good choices for these values probably depend on the transport / Stream implementation

sim/midas/src/main/scala/midas/core/StreamEngine.scala
```diff
@@ -0,0 +1,78 @@
...
+  /**
+    * Pretty prints a description of this stream.
+    */
+  def summaryString: String =
+    s"${name}, FPGA Buffer Depth: ${fpgaBufferDepth} Beats"
+}
+
+case class StreamSourceParameters(name: String, fpgaBufferDepth: Int) extends StreamParameters
```
davidbiancolin, line 37: Until these actually differ remove the unnecessary abstraction?

davidbiancolin:
FMR is in inline with performance reported on main (2.1-ish).

sim/midas/src/main/scala/midas/core/StreamEngine.scala
abejgonzalez: To be clear, this could also mean the host CPU queue/circular-buffer size? Like in the U250 case this would be the designated size of the circular buffer?

abejgonzalez: If so, I tend to prefer the `batchBytes` version.

sim/midas/src/main/scala/midas/core/CPUManagedStreamEngine.scala
```diff
@@ -0,0 +1,213 @@
...
+  }
+}
+
+class CPUManagedStreamEngine(p: Parameters, val params: StreamEngineParameters) extends StreamEngine(p) {
+
+  val dmaBytes = p(DMANastiKey).dataBits / 8
+  val pcisNodeOpt = Some(AXI4SlaveNode(
+      Seq(AXI4SlavePortParameters(
+        slaves = Seq(AXI4SlaveParameters(
+          address       = Seq(AddressSet(0, (BigInt(1) << p(DMANastiKey).dataBits) - 1)),
```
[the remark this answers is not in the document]
abejgonzalez: I'm fine punting it to another PR to help merge this in. For my understanding, `DMANastiKey` should be purged because it just sets the parameters for the CPU-managed DMA (AXI) - which is all done in this file? Would this be used somewhere else or would this just be here? Shouldn't there be different "DMANasti" parameters for potential different DMA cases or would a new stream engine be made?

abejgonzalez (approved):
LGTM. I took a much longer look at this and it seems to make sense (at least the general breakdown/abstraction). I think we can punt a bit of the Vitis specific naming to a later PR

sagark (approved):
LGTM

sim/midas/src/main/scala/midas/core/StreamEngine.scala
davidbiancolin: I was thinking the circular buffer would be much larger (since they are relatively inexpensive). I think we agreed to punt on the names until we had the PCI-M based implementation in front of us.

sim/midas/src/main/scala/midas/core/CPUManagedStreamEngine.scala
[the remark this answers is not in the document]
davidbiancolin: By purge i mean, replace with a new `Field` as i have done for the DRAM-subsystem that doesn't use `Nasti` classes, but a dedicated class, say `CPUManagedAXI4Params` or `PCISParams` or something to this effect, that captures all of the true configuration options for the interface. Also better names.

PR #1003, opened 2022-03-31
revert the change from #842 that makes launchrunfarm block on instances passing status checks

sagark:
This reverts the change from #842 that makes `launchrunfarm` block on instances passing status checks. This prevents us from overlapping ec2 instance boot with building fpga drivers (and soon metasim drivers). The manager already relies on `instance_liveness` as a cross-platform way of checking for instance reachability, because we don't want to run firesim-related commands on a runfarm where some instances are down for some reason.

CI doesn't need any special form of waiting because it's only booting one manager and thus the first command it tries to run on the manager via fabric will block anyway (and it looks like CI is using the same fabric connection params as the manager).

I also updated the `instance_liveness` docstring to clarify why it is run during `infrasetup`.

#### UI / API Impact

No functionality change.

`firesim launchrunfarm` will be faster (like before #842). infrasetup, etc. still check for `instance_liveness` like before.

N/A

### Contributor Checklist

### DIFF
--- a/deploy/runtools/firesim_topology_with_passes.py
+++ b/deploy/runtools/firesim_topology_with_passes.py
@@ -20,7 +20,7 @@ rootLogger = logging.getLogger()
 
 @parallel
 def instance_liveness():
-    """ Confirm that all instances are running first so that we don't run any
+    """ Confirm that all instances are accessible (are running and can be ssh'ed into) first so that we don't run any
     actual firesim-related commands on only some of the run farm machines."""
     rootLogger.info("""[{}] Checking if host instance is up...""".format(env.host_string))
     with StreamLogger('stdout'), StreamLogger('stderr'):
--- a/deploy/runtools/firesim_topology_with_passes.py
+++ b/deploy/runtools/firesim_topology_with_passes.py
@@ -20,7 +20,8 @@ rootLogger = logging.getLogger()
 
 @parallel
 def instance_liveness():
-    """ confirm that all instances are running first. """
+    """ Confirm that all instances are running first so that we don't run any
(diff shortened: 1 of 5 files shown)


sagark:
@abejgonzalez I think this has made it sufficiently far in CI for review purposes.

abejgonzalez (commented):
Can you also update this comment: https://github.com/firesim/firesim/blob/366ceabe13ce27056ec1c53bf93d017ec182d85e/deploy/firesim#L136 and https://github.com/firesim/firesim/blob/366ceabe13ce27056ec1c53bf93d017ec182d85e/deploy/runtools/run_farm.py#L274 to be a bit clearer on what "waiting" means (i.e. we are waiting for IP addresses to get populated not necessarily that the machine is ready to SSH into)

abejgonzalez (commented):
This should also be backported (with the label)

sagark:
@abejgonzalez all comments have been addressed

davidbiancolin:
Just a reminder @sagark you need to resolve the conflicts and merge this backport: 
https://github.com/firesim/firesim/pull/995

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -20,7 +20,8 @@
 
 @parallel
 def instance_liveness():
-    """ confirm that all instances are running first. """
+    """ Confirm that all instances are running first so that we don't run any
```
abejgonzalez, line 23:
```suggestion
    """ Confirm that all instances are accessible (are running and can be ssh'ed into) first so that we don't run any
```

abejgonzalez (approved):
Small nit on the comment otherwise I think this is fine.

PR #1016, opened 2022-04-11
Propagate manager tags to child instances

abejgonzalez:
In CI there are some cases where instances launched from a specific manager are not shut down. This PR addresses this by propagating all tags associated with the manager (except for tags that have the word `ci-manager` in the name) to child instances spawned from that manager. This should allow the "cull instances" CI job to remove leftover instances.

#### Related PRs / Issues

N/A

#### UI / API Impact

N/A

N/A

### Contributor Checklist

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -359,6 +359,20 @@ def launch_instances(instancetype: str, count: int, instancemarket: str, spotint
 
     first_subnet_wraparound = None
 
+    # append the tags from the manager to the launched instance
+    extra_tags = get_localhost_instance_tags()
+    extra_tags.pop('Name', None)
+    # only for ci, avoid naming child launched instances "ci-manager"
+    # to avoid ci erroring on finding multiple "manager" instances
+    for k in extra_tags.keys():
+        if "ci-manager" in k:
+            del extra_tags[k]
+    if tags is None:
+        tags = extra_tags
+    else:
+        tags.update(extra_tags)
+    rootLogger.debug(tags)
+
     while len(instances) < count:
         chosensubnet = subnets[startsubnet].subnet_id
         try:


deploy/awstools/awstools.py
```diff
@@ -36,6 +36,60 @@ def depaginated_boto_query(client, operation, operation_params, return_key):
...
+def get_manager_instance_id():
+    """Get current manager instance id, if applicable.
+
+    Returns:
+        A ``str`` of the instance id or ``None``
+    """
+    # This takes multiple minutes without a timeout from the CI container. In
+    # practise it should resolve nearly instantly on an initialized EC2 instance.
+    curl_connection_timeout = 10
+    with settings(warn_only=True), hide('everything'):
```
davidbiancolin, line 48: This code must be duplicated in multiple places in our python codebase -- can it not be DRYed out?

deploy/awstools/awstools.py
```diff
@@ -261,6 +278,17 @@ def launch_instances(instancetype, count, instancemarket, spotinterruptionbehavi
         if len(instances) < count:
             rootLogger.info("Launching remaining {} {} instances".format(count - len(instances), instancetype))
 
+    # append the tags from the manager to the launched instance
+    extra_tags = get_manager_tags()
+    # filter out the 'Name'
+    extra_tags.pop('Name', None)
+    # filter out any tags with the word 'manager' (for CI)
```
davidbiancolin, line 285: Re: these commends. I think the code, which is relatively straightforward, explains clearly _what_ is happening here -- it would be better to explain _why_ in the comment by expanding the parenthetical `(for CI)` to explain, for instance, what would happen if you didn't filter these out.

davidbiancolin (commented):
I know i'm responsible for the creation of a bunch of these tags, and i might not have done a good job explaining their function, but i'm wondering if a cleaner solution would have been to provide a command-line argument to manager to specify additional tags to pass on to build and runfarm? Then you wouldn't need to do this filtering behind the scenes, just a thought.

abejgonzalez:
> I know i'm responsible for the creation of a bunch of these tags, and i might not have done a good job explaining their function, but i'm wondering if a cleaner solution would have been to provide a command-line argument to manager to specify additional tags to pass on to build and runfarm? Then you wouldn't need to do this filtering behind the scenes, just a thought.

That is definitely a cleaner solution but requires a manager modification, I tried to avoid this (since adding stuff to the manager requires docs). I think I can just reuse some of the functionality of `awstools.py` cmdline to pass in tags through the manager.

https://github.com/firesim/firesim/blob/26313702cde4ca93d68d8ecf7b7d1bd94591fb23/deploy/awstools/awstools.py#L581

davidbiancolin:
Yeah, that's understandable. Another question is whether this field should also be exposed in the `ini` / `yaml` so maybe we should just punt on it for now. 

If the tags are just a dict it seems fairly straightforward to implement host-specific behavior behind it too.

deploy/awstools/awstools.py
abejgonzalez: I grepped for `169.254...` and only found the two matches in this file? Were you referring to something else?

abejgonzalez:
So is the plan to put this PR in for now and change the API later? If so, I can create an issue for this when we merge.

deploy/awstools/awstools.py
abejgonzalez: Better?

timsnyder-siv (commented):
I don't think we should have to inject this CI-specific stuff into awstools.  It feels wrong to be solving a problem with GH-A CI by changing how all instances are launched by all of FireSim but the only problems I can see are hypothetical, like "what if a user puts a special tag similar to your 'ci-manager' tag that they don't intend to be propagated to other instances that get launched and they have other scripts they use to query their running instances?

I'm not sure that my reservations warrant doing anything different at this point and other than the spot where you're passing arguments that are already a list into `list()` I don't see anything obviously broken.

abejgonzalez:
> I don't think we should have to inject this CI-specific stuff into awstools. It feels wrong to be solving a problem with GH-A CI by changing how all instances are launched by all of FireSim but the only problems I can see are hypothetical, like "what if a user puts a special tag similar to your 'ci-manager' tag that they don't intend to be propagated to other instances that get launched and they have other scripts they use to query their running instances?
> 
> I'm not sure that my reservations warrant doing anything different at this point and other than the spot where you're passing arguments that are already a list into `list()` I don't see anything obviously broken.

I agree that this isn't ideal. However, Berkeley is having some AWS credit issues right now that this PR will fix quickly (we can put an issue up later to fix this properly).

davidbiancolin:
I think switching to spot instances and being parsimonious with FPGA-based CI deployments is sufficient in the short term to bring our spending down. @abejgonzalez you might as well address Tim's comments now, get this right once, and merge it at your leisure.

Aside: if we're going to push something through on the basis of it being a critical fix, please limit the scope of the changes in the PR to only what is strictly necessary. I know many of the changes are related to type annotations, but this PR was previously only a handful of lines so it was relatively simple to understand the implications of the change.

abejgonzalez:
> I think switching to spot instances and being parsimonious with FPGA-based CI deployments is sufficient in the short term to bring our spending down. @abejgonzalez you might as well address Tim's comments now, get this right once, and merge it at your leisure.

Ack. SGTM as long as we have something else in the pipeline to lower costs.

> Aside: if we're going to push something through on the basis of it being a critical fix, please limit the scope of the changes in the PR to only what is strictly necessary. I know many of the changes are related to type annotations, but this PR was previously only a handful of lines so it was relatively simple to understand the implications of the change.

Ack. I'll split stuff up.

abejgonzalez:
Ok I opened #1037 and #1035. This PR will be rebased on those branches once they are reviewed / put in `main`

deploy/awstools/awstools.py
```diff
@@ -411,58 +450,56 @@ def get_instances_by_tag_type(tags, instancetype):
...
+def get_instance_ids_for_instances(instances: List[EC2InstanceResource]) -> List[str]:
     """" Take list of instances (as returned by create_instances), return instance ids. """
     return [instance.id for instance in instances]
 
-def instances_sorted_by_avail_ip(instances):
+def instances_sorted_by_avail_ip(instances: List[EC2InstanceResource]) -> List[EC2InstanceResource]:
     """ This returns a list of instance objects, first sorted by their private ip,
     then sorted by availability zone. """
-    ips = get_private_ips_for_instances(instances)
+    ips = get_private_ips_for_instances(list(instances))
```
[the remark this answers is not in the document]
abejgonzalez: This was a bug. Fixed in the other PR

PR #1018, opened 2022-04-13
post-merge CI needs to source env changes from machine-launch-script.sh

timsnyder-siv:
pre-merge GHA CI seems to be using completely different scripts than post-merge https://github.com/firesim/firesim/runs/6005045635?check_suite_focus=true. I can understand how this situation developed organically but it is a fairly stinky antipattern.  Rather than addressing the larger issue to make post-merge use the same scripts that are tested pre-merge, I'm just fixing the post-merge scripts.  I don't see how to make GH-A logout and log back in to the machine.  So, instead, I source the environment changes done by machine-launch-script.sh into the environment and cause them to persist via `env >> $GITHUB_ENV` as described in https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsenv. Looking into it more, that is just a file and it should accept the output of `env` just fine.

I also tweaked the way that `machine-launch-script.sh` introspects the system it is on to determine whether the system is supported (only Linux x86_64).  Just for grins, I Q/A'd the script on my macbook and the error messages were a tad nonsensical the way I had them before.  Shouldn't impact anyone running on a supported platform but the error messages on something unsupported will be better.

#### Related PRs / Issues

Hopefully fixes broken post-merge workflows that use actions/machinelaunchscript in a composite action.  i.e. https://github.com/firesim/firesim/runs/6005045635?check_suite_focus=true

### Contributor Checklist

### DIFF
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -178,7 +178,7 @@ set -o pipefail
         # run conda-init and look at it's output to insert 'conda activate $CONDA_ENV_NAME' into the
         # block that conda-init will update if ever conda is installed to a different prefix and
         # this is rerun.
-        $SUDO "${CONDA_EXE}" init $DRY_RUN_OPTION "${conda_init_extra_args[@]}" bash |& \
+        $SUDO "${CONDA_EXE}" init $DRY_RUN_OPTION "${conda_init_extra_args[@]}" bash 2>&1 | \
             tee >(grep '^modified' | grep -v "$CONDA_INSTALL_PREFIX" | awk '{print $NF}' | \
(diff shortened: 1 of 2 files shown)


scripts/machine-launch-script.sh
```diff
@@ -83,16 +83,28 @@ set -o pipefail
...
+	exit 1
+    fi
+
+    if [[ "$(uname)" != "Linux" ]]; then
+        echo "::ERROR:: $0 only supports 'Linux' not '$(uname)'"
+	exit 1
     fi
 
     if [[ "$(uname -mo)" != "x86_64 GNU/Linux" ]]; then
         echo "::ERROR:: $0 only supports 'x86_64 GNU/Linux' not '$(uname -io)'"
```
davidbiancolin, line 99: Unrelated to this PR, but the `uname -mo` vs `uname -io` was intentional here (as opposed to being -mo for both)?

PR #1021, opened 2022-04-13
XDC-Driven Memory Hints for Xilinx FPGAs

davidbiancolin:
This will be needed to support local FPGA work as it removes duplication of constraint files across different platforms. 

You deploy this like so:
```
import midas.targetutils.xdc.{RAMStyleHint, RAMStyles} 

val myMem = SyncReadMem(<params>)
RAMStyleHint(myMem, RAMStyles.ULTRA) // For a URAM-based implementation
```

#### UI / API Impact

Expands. Makes it possible to coax vivado into using URAMs without changing the directives.

No -- however I'll regenerate AGFIs when i deploy these in the StreamEngine.

### Contributor Checklist

### DIFF
--- a/sim/midas/targetutils/src/main/scala/midas/xdc/RAMStyleHint.scala
+++ b/sim/midas/targetutils/src/main/scala/midas/xdc/RAMStyleHint.scala
@@ -1,0 +1,73 @@
+// See LICENSE for license details.
+
+package midas.targetutils.xdc
+
+import chisel3.experimental.{ChiselAnnotation}
+import firrtl.annotations.ReferenceTarget
+import midas.targetutils.xdc.{XDCFiles, XDCAnnotation}
+
+sealed trait RAMStyle
+
+/**
+  * Some rough guidance, based on Ultrascale+, is provided in the scala doc for
+  * each hint. Consult the Xilinx UGs for your target architecture and the
+  * synthesis UG (UG901).
+  */
+object RAMStyles {
+  /**
+    *  From UG901 (v2020.2): Instructs the tool to use the UltraScale+ URAM primitives.
+    *
+    *  URAMs are 288Kb, 72b wide, 4096 deep.
+    */
+  case object ULTRA extends RAMStyle
+
+  /**
+    *  From UG901 (v2020.2): Instructs the tool to infer RAMB type components
+    *
+    *  In Ultrascale+ (and older families), BRAMs are 36Kb, and have flexible aspect
+    *  ratio: 1b x 64K to 72b x 512
+    */
+  case object BRAM extends RAMStyle
+
+  /** From UG901 (v2020.2): Instructs the tool to infer registers instead of RAMs. */
+  case object REGISTERS extends RAMStyle
+
+  /** From UG901 (v2020.2): Instructs the tool to infer the LUT RAMs. */
+  case object DISTRIBUTED extends RAMStyle
+
+  /** Introduced in v2020.2. From UG901 (v2020.2):
+    *
+    * Instructs the tool to infer a combination of RAM types designed to minimize the
+    * amount of space that is unused.
+    *
+    * Note: This should probably be avoided for non-emulation-class
+    * FPGAs (e.g., VU19P), which tend to have rich embedded memory resources, as
+    * these designs tend to be under heavy LUT pressure. Here sparsely using
+    * BRAM / URAM resources insead of a space-optimal hybrid is desirable.
+    */
+  case object MIXED extends RAMStyle
+}
+
+object RAMStyleHint {
+  // _reg suffix is applied to memory cells by Vivado, the glob manages
+  // duplication for multibit memories.
+  private [midas] def propertyTemplate(style: RAMStyle): String =
+    s"set_property RAM_STYLE ${style} [get_cells {}_reg*]"
+
+  private def annotate(style: RAMStyle, rT: =>ReferenceTarget): Unit = {
+    chisel3.experimental.annotate(new ChiselAnnotation {
+      def toFirrtl = XDCAnnotation(
+        XDCFiles.Synthesis,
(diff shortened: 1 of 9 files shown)


sim/midas/targetutils/src/main/scala/midas/xdc/RAMStyleHint.scala
```diff
@@ -0,0 +1,73 @@
...
+    * Instructs the tool to infer a combination of RAM types designed to minimize the
+    * amount of space that is unused.
+    *
+    * Note: This should probably be avoided for non-emulation-class
+    * FPGAs (e.g., VU19P), which tend to have rich embedded memory resources, as
+    * these designs tend to be under heavy LUT pressure. Here sparsely using
+    * BRAM / URAM resources insead of a space-optimal hybrid is desirable.
+    */
+  case object MIXED extends RAMStyle
+}
```
timsnyder-siv, line 49: What happens if/when we support platforms other than Xilinx?  Do you have thoughts on how to extend this?  Doesn't have to be done in this PR.

sim/midas/targetutils/src/test/scala/RAMStyleHintSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  it should "correctly annotate a chisel3.Mem as URAM" in {
+    val annos = elaborateAndCollectXDCAnnos(new CombReadMemModule(annotateMem(RAMStyles.ULTRA)))
+    assert(annos.size == 1)
+  }
+
+  it should "correctly annotate a chisel3.Mem as BRAM" in {
+    val annos = elaborateAndCollectXDCAnnos(new CombReadMemModule(annotateMem(RAMStyles.BRAM)))
+    assert(annos.size == 1)
```
timsnyder-siv, line 86: Ummm.  You sure `annos.size == 1` is enough of an assertion here?  It's braver than I would be but you understand the innerworkings of chisel and annotation better than I.

sim/midas/targetutils/src/test/scala/RAMStyleHintSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  it should "not trivially break deduplication" in {
+    val annos = elaborateAndLower(new WrapperModule(annotateMem(RAMStyles.ULTRA)))
+
+    val dedupResultAnnos = annos.collect { case a: firrtl.transforms.DedupedResult => a }
+    assert(dedupResultAnnos.size == 2)
+
+    val xdcAnnos = annos.collect { case a: XDCAnnotation => a }
+    assert(xdcAnnos.size == 2)
+  }
```
timsnyder-siv, line 97: Am I reading this correctly... you're annotating the duplicated `modA` and `modB` with the same type of RAMStyle and you don't expect dedup to collapse that into a single annotated module?  I'm confused why this is.

Should you test for the inverse case?  I would assume that would be where you anntate `modA` and `modB` with different `RAMStyle` and you would expect that to not be deduped, right?

timsnyder-siv (commented):
I have some questions.  Maybe after we discuss more, I'll feel confident that I'm more than rubber stamping this.

sim/midas/targetutils/src/test/scala/RAMStyleHintSpec.scala
davidbiancolin: Yeah, this is not very precise. It didn't seem to me to be that useful to reconstitute the expected annotation but i'll just do that. 

If i push the machinery that actually generates the XDC into targetutils, i can test the end to end flow here. That might be more desireable in the long run.

sim/midas/targetutils/src/test/scala/RAMStyleHintSpec.scala
davidbiancolin: Dedup does deduplicate them (there would be no DedupResultAnnos otherwise; these show the remapping). The XDC annotations themselves point at instances of the deduped module instead, at least under current "annotation deduplication" semantics. 

AFAIK there is really no specification -- there is just the code -- that governs annotation behavior under deduplication, and as of 1.4 (i believe), the code changed to do this module -> instance style deduplication. 
In practice you want different behavior for different kinds of annotations. This was the source of much of our pain when doing the 1.4 bump. If you don't want the deduplication via mapping to instances, but instead want to combine the annotation into a single module-local annotation, i think you have to roll your own pass to do that.

In any case, the way these XDC annotations behave should work under both "styles" of deduplication since it can handle instance paths and module-local targets.

sim/midas/targetutils/src/main/scala/midas/xdc/RAMStyleHint.scala
davidbiancolin: I did think about it. The big question is whether we want the specialization to happen at elaboration time or RTL compile time. 

If we choose the prior, i think it would be natural to add more classes of style which would imply a different constraint string and different output file. There could be companion objects for each arch type for doing arch-specific annotations, and a general purpose one which could use an arch-agnostic hint, or a heuristic about the size / parameterization. 

If we choose the latter, well the whole annotation interface would change, but the pass could do global optimizations in RAM selection. This strikes me as somewhat fraught though, since Vivado does the same thing, and its non trivial to reason through the interactions with other Vivado compile directives. 

Really, my intuition tells me that we should only really use these to coax vivado into doing something non-default (i.e., deploy URAMs even if they are going to be very-sparsely filled), but otherwise remain hands off. If that's the case -- having extremely precise arch-specific annotations is probably preferable.

timsnyder-siv (approved):
I'm satisfied with the answers I got to my questions and this looks good to me. :shipit:

abejgonzalez (approved):
The main API/annotation changes LGTM. I didn't look at the tests.

PR #1024, opened 2022-04-14
[ci] Cleaner propagation of env after machine launch for containers

davidbiancolin:
Resolves a small error with #1018. 

#### Related PRs / Issues

#1018 

#### UI / API Impact

None.

No

### Contributor Checklist

### DIFF
--- a/.github/actions/machinelaunchscript/action.yml
+++ b/.github/actions/machinelaunchscript/action.yml
@@ -1,12 +1,20 @@
 name: machinelaunchscript
-description: "Run FireSim's machine-launch-script.sh"
+description: "Run FireSim's machine-launch-script.sh. Used only for GH-hosted CI containers"
 
 runs:
   using: "composite"
   steps:
     - run: |
         sudo yum -y remove git git224 git224-core ius-release.noarch # remove any older git versions and collateral first
+        # The EC2 developer AMI provides libmpc and the prebuilt toolchain
+        # links against it.  Install it here so we can build-setup.sh --fast
+        # instead of building a toolchain from scratch.
+        sudo yum install -y libmpc
         cd scripts/ && /usr/bin/bash machine-launch-script.sh
         source /etc/profile.d/conda.sh
-        env >> $GITHUB_ENV # persist the machine-launch-script env changes to other actions
+        # Mark the github workspace as a safe directory so we can run arbitrary
+        # git commands during build-setup.sh.
+        # See https://github.blog/2022-04-12-git-security-vulnerability-announced/
+        # and https://github.com/actions/checkout/issues/760 for the workaround
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
       shell: bash
--- a/.github/actions/buildsetup/action.yml
+++ b/.github/actions/buildsetup/action.yml
@@ -1,8 +1,14 @@
 name: buildsetup
-description: "Run firesim's build-setup.sh"
+description: "Run firesim's build-setup.sh. Used only for GH-hosted CI containers."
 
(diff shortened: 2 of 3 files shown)


timsnyder-siv:
I tried testing this by pulling it into my fork and running and I'm hitting another issue that seems related to https://github.blog/2022-04-12-git-security-vulnerability-announced/ which was just released a couple of days ago.

https://github.com/timsnyder/firesim/runs/6030380191?check_suite_focus=true#step:4:2590

@davidbiancolin, have you tested this somehow?

davidbiancolin:
> I tried testing this by pulling it into my fork and running and I'm hitting another issue that seems related to https://github.blog/2022-04-12-git-security-vulnerability-announced/ which was just released a couple of days ago.
> 
> https://github.com/timsnyder/firesim/runs/6030380191?check_suite_focus=true#step:4:2590
> 
> @davidbiancolin, have you tested this somehow?

I'm going to try to merge #1025 and use that to test this. I could also combine the two PRs if thats preferable.

timsnyder-siv:
> I could also combine the two PRs if thats preferable.

However you want to get it done is fine.  I just wanted to point out that I don't think this is enough to get it working again because of the change in git behavior we seem to be hitting now.

davidbiancolin:
@mergifyio rebase

.github/actions/machinelaunchscript/action.yml
```diff
@@ -6,7 +6,8 @@ runs:
   steps:
     - run: |
         sudo yum -y remove git git224 git224-core ius-release.noarch # remove any older git versions and collateral first
+        sudo yum install -y libmpc
```
timsnyder, line 9:
```suggestion
        sudo yum install -y libmpc # needed for build-toolchains.sh ec2fast precompiled toolchain in CI
```
Might be good to leave a comment about why we are installing mpc here.

timsnyder (approved):
Seems like we've got it working now.  You might add a comment to why we're installing mpc if you feel like it.

.github/actions/machinelaunchscript/action.yml
davidbiancolin: Do you think i should move it into the docker image though? It's probably worth re-assessing what's needed here:
https://github.com/firesim/firesim/blob/main/.github/scripts/Dockerfile

davidbiancolin: It seems to me like most of it can be ripped out, since conda will bring in much of this.

timsnyder-siv: Most of what @abejgonzalez is installing in the Dockerfile is trying to avoid really weird behavior that happens if you allow GH-A to clone the repo using the default centos7.6 version of git and then you're removing everything installed by the Dockerfile before machine-launch-script.sh gets run.

Although conda is going to pull in everything you want, I think you'll still end up hitting the weirdness Abe encountered with the GH-A checkout action using ancient git if you remove his workaround.

To me, it might make sense to change the Dockerfile so that it runs machine-launch-script.sh to create the docker image and have a new version get generated whenever machine-launch-script.sh changes?  That seems like it would be more work than I wanted to do right now though.

timsnyder-siv: > I think i'd prefer to keep the container images as close to vanilla as possible, and then maybe rely on GH-A provided caching to speed up CI (and save cost).

Sure.  Sounds reasonable.

> We shouldn't need to upgrade pip anymore or add the the centos user though, right?

pip, defintiely not and I don't think we need to have centos user anymore but I'm not positive about that.

PR #1010, opened 2022-04-05
`_` instead of `-`-based YAML files + API changes

abejgonzalez:
This PR changes the default names of all YAML facing APIs to be `_`-based instead of "none" and `-`-based. 

Also changes a few other APIs:
In `_build.yaml` - `default-build-farm` to `build_farm`
In `_build.yaml` - `builds` to `builds_to_run`
In `_build_farm.yaml` - `build_dir` to `default_build_dir` in the AWSEC2 (BuildFarm)
In `_build_farm.yaml` - use the build farm class name as the `type`
In `_build_farm.yaml` and `buildfarm.py` - rename "IPAddr" build farm to "ExternallyProvisioned" and "EC2" build farm to "AWSEC2"

This PR does not change documentation as that needs to be updated anyways with changes from PR #1006.

#### Related PRs / Issues

#1006 

#### UI / API Impact

Changes YAML facing APIs to be `_`-based.

#### Verilog / AGFI Compatibility

N/A

### Contributor Checklist
- [x] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [x] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous prints/debugging code?
- [x] Did you state the UI / API impact?
- [x] Did you specify the Verilog / AGFI compatibility impact?

- [x] If applicable, did you regenerate and publicly share default AGFIs?

- [x] If applicable, did you apply the `ci:fpga-deploy` label?

- [x] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
- [x] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [x] Did you mark the proper release milestone?
- [x] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1028, opened 2022-04-15
("Where to Run") Initial support running on different run farm hosts

abejgonzalez:
This work creates the abstractions needed to have multiple different "run hosts" associated with different "build platforms".

**CI Changes**

* Added a `RUN_LOCAL` bool. variable that allows CI tests to run locally. This was added to a new file.
* Added new "externally provisioned" version of the Linux poweroff test (only the no NIC version supported). It does the following:
  1. launches a runfarm using the AWS EC2 specific runfarm code (i.e. the default runfarm type)
  2.find out the hostnames of spawned instances and copies them into a new externally provisioned runfarm/runtime file
  3. run launch/infra/runworkload/terminate w/ that new runtime
  4. run the terminate w/ the old AWS EC2 runfarm
* Since both linux tests (the default AWS EC2 version and externally provisioned version) can run in parallel, the `RUNFARM_PREFIX` env. variable is unique to both tests.
* Removed extra marshal calls (this is duplicated since this is done in an earlier step).

**Notable Non-API Changes:**

* Adds a new function called `file_line_swap` into `util`s area. This is a helper function to replace lines of text in a file with another (used in CI and in `managerinit`).
* Moves the `MockBoto3Instance` into `awstools.py` since it is supposed to represent a Boto object
* Moves the `inheritors` function (how to determine child classes from a parent class) into a shared area (`util`)
* Adapts a few features from #1005
  * Includes various changes to simplify future multi-platform support that arose as a result of trying to minimize differences between metasim/fpga sim
    * No more hard-coded per-instance-type python classes. We now have arbitrary instance type support in all commands, e.g., new `--terminatesome=inst_type:count` flag
    * Topology mapping no longer cares about specific instance types, only the max number of fpgas a machine has (there are new functions for this to work.
* Simplifies argument passing for debugging features like TracerV, AutoCounter, HostDebug, and SynthPrints - Creates dataclasses in `simulation_data_classes.py` for them with defaults mentioned there. 
* Breaks up the driver command into multiple lines for future readability + better git diffs
* `InstanceDeployManager` is now the replaceable class to modify how simulations are run.
  * `NBDTracker` is now only used in the AWS case (this can be moved later for the Vitis case)
  * It requires that the `Inst` that it refers to has it's hostname specified (should be completed by the top-level run farm class that uses it)
* `run_farm.py` file diff is horrible FYI - You really should just look at the file itself instead of the diff.
  * `AWSEC2F1` run farm - manages how to boot/terminate sims. binds hostnames to inst objects.
  * `ExternallyProvisioned` run farm - binds hostnames to inst objects. Note: The default switch type name just returns the 1st instance that doesn't have FPGA slots... this makes it so that you can't run a networked simulation on a single FPGA node. This can be updated later. Technically this code is never reached since I assert that only `no_net_configs` can be used with non-AWS runfarms (again this could probably be relaxed - but later).

**Leftover TODOs:**
- Documentation (will punt to another PR)
- Clean up naming of things
  - `simulation_data_classes.py` (name of file + classes)
  - Mapper functions used by topologies. Handle?
- Should we move the copying/setup functions (i.e. stuff that mounts FS's etc) into the instance deploy manager instead of the `Inst` type)? This includes arguments to the driver.  - From a meeting this is punted to future work.

#### Related PRs / Issues

#1005

#### UI / API Impact

**`run_farm_recipe.yaml`:**

This is a similar structure to the `config_build_farm.yaml` file but now it isn't a main `config_*` file. Here a `run_farm_type` is given (pointing to a class in `run_farm.py`) with its requisite `args`. There are two provided run farms.

* `AWSEC2F1` run farm - This moves alot of the AWS specific flags from `runtime.yaml` to here. Similar to the build farms, I expose a flag for changing where the simulation directory is located (`default_simulation_dir`). Additionally, the FPGA inst. types are now under `run_farm_hosts`. Like #1005 the list with single value dicts (Key = INSTTYPE -> Value = NUM). Right now, what is "allowed" is defined in two dicts at the top of `AWSEC2F1` in `run_farm.py`. All other manager code now does not care about specific instance names. e.g. If an `f1.256xlarge` instance were released, the only place it needs to be added is the 2 dicts at the top of `AWSEC2F1` and in `run_farm_hosts`.
* `ExternallyProvisioned` run farm - Similar to the build farm, this allows users to give arbitrary host names to run simulations on. In this case, we have a set of default values like `num_fpgas`, `default_platform`, and `default_simulation_dir` that can be overridden by the specific run host. `run_farm_hosts` are like `build_farm_hosts`, they have just a name (i.e. use the defaults for everything else) or a name and arguments overriding the defaults. The `*platform` argument is how you configure this to target different platforms. In this case, only 1 platform is supported `EC2InstanceDeployManager`. If you want to use Vitis (or something else), you would write another "instance deploy manager" class and replace the `*platform` arg. For this case, the instance deploy manager for EC2 `EC2InstanceDeployManager` works with the `ExternallyProvisioned` run farm. You just need to manually spawn your own F1 instances, and then point to their hostnames in the file. I.e. do `launchrunfarm` with the `AWSEC2F1` runfarm, switch to the `ExternallyProvisioned` one with the hostnames specified, run the simulations, then switch back to the `AWSEC2F1` run farm and terminate the instances.

**`config_runtime.yaml`:**

Moved the `run_farm` specific arguments into the `run_farm.yaml` file. This `run_farm_config:` argument just takes the file name of the run farm recipe configuration to use as defaults. It also has an override section that allows you to override recipe arguments but not the run farm type (i.e. once you use an AWS EC2 specific default run farm file you can't switch the run farm type to Externally Provisioned).

**`firesim` manager:**

* `--terminatesome{f116,f14,f12,m116}=count` flags to the manager are all deprecated. The user can instead specify an arbitrary number of: `--terminatesome=INST_TYPE:count` flags. The `INST_TYPE` corresponds to the AWS instance names given by the 2 dicts at the top of the `AWSEC2F1` run farm type in `run_farm.py`.  This flag doesn't do anything for the externally provisioned run farm (in the Abseil PR this would be resolved and the flag would be specific to the `AWSEC2F1` class).
* `managerinit`
  * Modifies `config_runtime.yaml` by adding the defaults of the AWS EC2 runfarm into the `override_args` section. If the platform isn't F1 then it just skips this step.
  * Moves the `get_aws_userid` call into an if statement depending on the platform.

No.

### Contributor Checklist


abejgonzalez:
@sagark This should be easier to look at now.

timsnyder-siv:
What is `lfmimv-fixed` branch and why is this merging into that?

abejgonzalez:
> What is `lfmimv-fixed` branch and why is this merging into that?

I am assuming this PR goes through (which is the lfmimv-fixed branch): https://github.com/firesim/firesim/pull/1010

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -448,44 +452,46 @@ def boot_simulation_passes(self, use_mock_instances_for_testing: bool, skip_inst
...
-            self.run_farm.bind_real_instances_to_objects()
+        self.run_farm.post_launch_binding(use_mock_instances_for_testing)
 
-        all_runfarm_ips = [x.get_private_ip() for x in self.run_farm.get_all_host_nodes()]
+        all_runfarm_ips = [x.get_ip() for x in self.run_farm.get_all_bound_host_nodes()]
 
         @parallel
         def kill_switch_wrapper(runfarm: RunFarm) -> None:
             my_node = runfarm.lookup_by_ip_addr(env.host_string)
+            assert my_node is not None
```
abejgonzalez, line 486:
```suggestion
```

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -448,44 +452,46 @@ def boot_simulation_passes(self, use_mock_instances_for_testing: bool, skip_inst
...
 
-        all_runfarm_ips = [x.get_private_ip() for x in self.run_farm.get_all_host_nodes()]
+        all_runfarm_ips = [x.get_ip() for x in self.run_farm.get_all_bound_host_nodes()]
         execute(instance_liveness, hosts=all_runfarm_ips)
         execute(boot_switch_wrapper, self.run_farm, hosts=all_runfarm_ips)
 
         @parallel
         def boot_simulation_wrapper(runfarm: RunFarm) -> None:
             my_node = runfarm.lookup_by_ip_addr(env.host_string)
+            assert my_node is not None
```
abejgonzalez, line 471:
```suggestion
```

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -448,44 +452,46 @@ def boot_simulation_passes(self, use_mock_instances_for_testing: bool, skip_inst
...
         def kill_switch_wrapper(runfarm: RunFarm) -> None:
             my_node = runfarm.lookup_by_ip_addr(env.host_string)
+            assert my_node is not None
+            assert my_node.instance_deploy_manager is not None
             my_node.instance_deploy_manager.kill_switches_instance()
 
         @parallel
         def kill_simulation_wrapper(runfarm: RunFarm) -> None:
             my_node = runfarm.lookup_by_ip_addr(env.host_string)
+            assert my_node is not None
```
abejgonzalez, line 493:
```suggestion
```

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -511,12 +517,14 @@ def screens() -> None:
...
-            self.run_farm.bind_mock_instances_to_objects()
-        else:
-            self.run_farm.bind_real_instances_to_objects()
 
-        all_runfarm_ips = [x.get_private_ip() for x in self.run_farm.get_all_host_nodes()]
+        if isinstance(self.run_farm, AWSEC2F1):
+            if use_mock_instances_for_testing:
+                self.run_farm.bind_mock_instances_to_objects()
+            else:
+                self.run_farm.bind_real_instances_to_objects()
```
abejgonzalez, line 525:
```suggestion
            self.run_farm.post_launch_binding(use_mock_instances_for_testing)
```

deploy/runtools/run_farm.py
```diff
@@ -257,58 +169,62 @@ def bind_real_instances_to_objects(self) -> None:
...
-
         # actually launch the instances
         f1_16s = launch_run_instances('f1.16xlarge', num_f1_16xlarges, runfarmtag,
                                       runinstancemarket, spotinterruptionbehavior,
                                       spotmaxprice, timeout, always_expand)
         f1_4s = launch_run_instances('f1.4xlarge', num_f1_4xlarges, runfarmtag,
-                                     runinstancemarket, spotinterruptionbehavior,
-                                     spotmaxprice, timeout, always_expand)
+                                      runinstancemarket, spotinterruptionbehavior,
+                                      spotmaxprice, timeout, always_expand)
```
abejgonzalez, line 199:
```suggestion
                                     runinstancemarket, spotinterruptionbehavior,
                                     spotmaxprice, timeout, always_expand)
```

deploy/runtools/run_farm.py
```diff
@@ -257,58 +169,62 @@ def bind_real_instances_to_objects(self) -> None:
...
                                       spotmaxprice, timeout, always_expand)
         f1_2s = launch_run_instances('f1.2xlarge', num_f1_2xlarges, runfarmtag,
                                      runinstancemarket, spotinterruptionbehavior,
                                      spotmaxprice, timeout, always_expand)
 
-        # wait for instances to get to running state, so that they have been
-        # assigned IP addresses
+        # wait for instances to finish launching
+        # TODO: maybe we shouldn't do this, but just let infrasetup block. That
+        # way we get builds out of the way while waiting for instances to launch
```
abejgonzalez, line 209:
```suggestion
        # wait for instances to get to running state, so that they have been
        # assigned IP addresses
```

abejgonzalez:
I've verified that this works booting a 1-node no NIC simulation. Additionally, I've tested externally provisioned with a 2-node no NIC simulation. Specifically launchrunfarm w/ AWS runfarm (2 nodes).

I think the final things needed for this PR is better comments / naming. And 2 fixing the example workload files with the runfarm changes.

abejgonzalez:
@timsnyder-siv @davidbiancolin @sagark This is ready for the 1st round of feedback (with the caveats stated above)

deploy/runtools/runtime_config.py
```diff
@@ -40,7 +42,18 @@ class RuntimeHWConfig:
...
         self.name = name
+
+        # TODO: this will change based on the "what-to-build" PR
+        #self.agfi = None
+        #self.xclbin = None
+        #self.platform = hwconfig_dict['platform']
+        #if self.platform == 'vitis':
+        #    self.xclbin = hwconfig_dict['xclbin']
+        #elif self.platform == 'f1':
+        #    self.agfi = hwconfig_dict['agfi']
```
abejgonzalez, line 53:
```suggestion
```

davidbiancolin:
I'm starting to sound like a broken record, but in the future please break this into smaller PRs with focussed changes. 

Stuff like: 
> Simplifies argument passing for debugging features like TracerV, AutoCounter, HostDebug, and SynthPrints - Creates dataclasses in simulation_data_classes.py for them with defaults mentioned there.

Can be one PR, etc... This reduces the cognitive load on the reviewers, and makes it easier to give precise and timely feedback.

davidbiancolin (commented):
First drop, oops. Will review more later

deploy/sample-backup-configs/sample_config_runtime.yaml
```diff
@@ -1,17 +1,7 @@
...
-    always_expand_runfarm: yes
-    f1_16xlarges: 1
-    m4_16xlarges: 0
-    f1_4xlarges: 0
-    f1_2xlarges: 0
-    launch_instances_timeout_minutes: 60
-    run_instance_market: ondemand
-    spot_interruption_behavior: terminate
-    spot_max_price: ondemand
+run_farm: aws_ec2_f1_run_farm
```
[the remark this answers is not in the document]
abejgonzalez: I think this is something that we might want to discuss in more detail. IMO once you start adding overrides it makes it confusing to find out what values are used. Here are a couple of schemas that might work:

1. Have the run farm configuration inlined (in addition to the ability to put it in a different file):

```yaml
# inlined case
run_farm:
  run_farm_name:
  run_farm_type: XYZ
  args:
    ....
    
# non-inlined case
run_farm: run_farm_special_name

# or we could keep it uniform
run_farm:
  run_farm_name: run_farm_special_name
```

2. Specify an override dict - like what you want. Basically it will override an argument given. To override you would have to do something like this?

```yaml
run_farm:
  name: XYZ
  overrides:
    run_farm_tag: MY OVERRIDE
    run_farm_hosts:
       - f1.2xlarge: 20
```

This problem also exists for the build farms as well (mainly for the externally provisioned case).

I think this gets tricky if there are nested dicts. For example, should `run_farm_hosts` keep the values of the other F1 instances and only change `f1.2xlarge`? Or should you have to re-specify all the F1 instances (respecify the count of `m4.16`s etc (in the current example the `m4.16` would be 0 by default as well as the other F1 instances).

timsnyder-siv (commented):
I think there are still some things that we need to talk through some more.

deploy/workloads/run-workload.sh
```diff
@@ -11,9 +11,11 @@ terminate=1
...
 function usage
 {
-    echo "usage: run-workload.sh <workload.yaml> [-H | -h | --help] [--noterminate] [--withlaunch]"
-    echo "   workload.yaml:  the firesim-relative path to the workload you'd like to run"
-    echo "                  e.g. workloads/gapbs.yaml"
+    echo "usage: run-workload.sh <workload> [-H | -h | --help] [--noterminate] [--withlaunch]"
+    echo "   workload:  the firesim-relative path to the basename of yaml files you'd like to run."
+    echo "              in this case, the path will be appended with -runtime.yaml for the runtime.yaml and -runfarm.yaml for the runfarm.yaml files"
+    echo "              e.x. workloads/gapbs gives"
+    echo "                   workloads/gapbs-runtime.yaml and workloads/gapbs-runfarm.yaml"
```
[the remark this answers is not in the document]
abejgonzalez: So I think this is a problem that needs to be solved better. I have some small discussion here on it: https://github.com/firesim/firesim/pull/1028#discussion_r867070819. Maybe build farm / run farm configurations should be inlined into the same file as the `config_build.yaml` and `config_runtime.yaml`. To reduce the amount of files needed. This probably should be discussed in our meeting.

deploy/firesim
```diff
@@ -298,17 +301,22 @@ def construct_firesim_argparser() -> argparse.ArgumentParser:
...
-                        help='Only used by terminatesome. Terminates this many of the previously launched f1.4xlarges.',
+                        help='DEPRECATED. Use --terminatesome=f1.4xlarge:count instead. Will be removed in the next major version of FireSim (1.15.X). Old help message: Only used by terminaterunfarm. Terminates this many of the previously launched f1.4xlarges.',
                         default=-1)
     parser.add_argument('-m', '--terminatesomem416', type=int,
-                        help='Only used by terminatesome. Terminates this many of the previously launched m4.16xlarges.',
+                        help='DEPRECATED. Use --terminatesome=m4.16xlarge:count instead. Will be removed in the next major version of FireSim (1.15.X). Old help message: Only used by terminaterunfarm. Terminates this many of the previously launched m4.16xlarges.',
                         default=-1)
+    def terminatesomesplitter(raw_arg):
+        split_arg = raw_arg.split(":")
+        return split_arg[0], int(split_arg[1])
```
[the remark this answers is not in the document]
abejgonzalez: Most likely, this was done by Sagar so I didn't modify it.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -191,33 +185,22 @@ def pass_no_net_host_mapping(self) -> None:
...
-                f1_4x.add_simulation(servers[serverind])
-                serverind += 1
-                if len(servers) == serverind:
-                    return
-        for f1_2x in self.run_farm.f1_2s:
-            for x in range(f1_2x.get_num_fpga_slots_max()):
-                f1_2x.add_simulation(servers[serverind])
+        while len(servers) > serverind:
+            # this call will error if no such instances are available.
+            instance_type = self.run_farm.mapper_get_min_sim_host_inst_type_name(1)
```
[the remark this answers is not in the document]
abejgonzalez: Again, better names always recommended. Basically the function is "Return a run host name (i.e. `f1.2xlarge` (in AWS's case) or `192.168.1.1` (in the IP address run farm case)) that is able to host `num_sims` simulation slots. If there are multiple hosts that can host `num_sims` slots, return the smallest host (i.e. the host with the smallest number of sim. slots".

In any case, I think the function should return what we call a "handle" (since this "handle" represents really depends on the underlying run farm). In the case of the AWS run farm a "handle" is an instance name (`f1.2xlarge`). In the case of a self managed run farm it is the IP address of a host. In other cases, it could potentially be something else (an object?). So maybe the function is better named `get_smallest_sim_host_handle(min_sim_supported=N)` (again probably needs more work to name nicely)?

deploy/sample-backup-configs/sample_config_runtime.yaml
abejgonzalez, line 4:
```
run_farm:
  run_farm_recipe: <PATH TO DEFAULTS FILE>
  overrride_args:
     <ARG> : <OVERRIDE>
```

abejgonzalez: Supposedly its easy to combine the two different dicts (w/ nesting taken into account).

abejgonzalez: managerinit will copy the default args of the platform given into the override section and point the runfarmrecipe to that default file. in the workloads yaml we will just point to a f1 recipe by default (in a file).


instance counts are defaulted to 0 (or whatever it was b4)

abejgonzalez: Implemented. Closing

deploy/firesim
```diff
@@ -125,9 +126,30 @@ def managerinit(args: argparse.Namespace):
...
-            rootLogger.debug(m)
-            rootLogger.debug(m.stderr)
+
+            if args.platform == 'f1':
+                m = local("""sed -i 's/AWSUSERNAME/{}/g' config_{}.yaml""".format(get_aws_userid(), conf_file), capture=True)
+                rootLogger.debug(m)
+                rootLogger.debug(m.stderr)
+
+    rootLogger.info("Adding default overrides to default runtime.yaml file")
+    if args.platform == 'f1':
```
abejgonzalez, line 136: To confirm, this is what people intended. I think this can be cleaned up later so that platform specific code doesn't leak into the general manager code.

deploy/regression.sh
```diff
@@ -20,7 +20,7 @@ if [[ $local_diff_rc = 0 && $remote_diff_rc = 0 ]]; then
 fi
 
 # Run linux-poweroff on unnetworked targets
-./workloads/run-workload.sh workloads/linux-poweroff-all-no-nic.yaml --withlaunch
+./workloads/run-workload.sh workloads/linux-poweroff-all-no-nic --withlaunch
```
abejgonzalez, line 23:
```suggestion
./workloads/run-workload.sh workloads/linux-poweroff-all-no-nic.yaml --withlaunch
```

deploy/sample-backup-configs/sample_config_runtime.yaml
```diff
@@ -1,17 +1,13 @@
...
-    f1_16xlarges: 1
-    m4_16xlarges: 0
-    f1_4xlarges: 0
-    f1_2xlarges: 0
-    launch_instances_timeout_minutes: 60
-    run_instance_market: ondemand
-    spot_interruption_behavior: terminate
-    spot_max_price: ondemand
+run_farm_config:
+  # managerinit replace start
```
abejgonzalez, line 5: The `managerinit replace start/end` is used to replace this section in `managerinit`. Naming could potentially change here:
- `run_farm_config`? `run_farm`? `run_farm_recipe`?
- `defaults`? `default_recipe`?
- `override_args`? `overrides`?

deploy/runtools/runtime_config.py
```diff
@@ -268,25 +295,34 @@ def __init__(self, runtimeconfigfile: str, configoverridedata: str) -> None:
...
+                    result[bk] = deepcopy(bv)
+            return result
 
-        self.f1_16xlarges_requested = int(runtime_dict['run_farm']['f1_16xlarges']) if 'f1_16xlarges' in runtime_dict['run_farm'] else 0
-        self.f1_4xlarges_requested = int(runtime_dict['run_farm']['f1_4xlarges']) if 'f1_4xlarges' in runtime_dict['run_farm'] else 0
-        self.m4_16xlarges_requested = int(runtime_dict['run_farm']['m4_16xlarges']) if 'm4_16xlarges' in runtime_dict['run_farm'] else 0
-        self.f1_2xlarges_requested = int(runtime_dict['run_farm']['f1_2xlarges']) if 'f1_2xlarges' in runtime_dict['run_farm'] else 0
+        override_args = runtime_dict['run_farm_config'].get('override_args')
+        if override_args:
+            run_farm_args = deep_merge(run_farm_args, override_args)
```
abejgonzalez, line 320: This is the "override" code, it only merges the dicts and not lists. Take the following case:

```
override_args:
    KEY0: VAL0
    KEY1:
      KEY2: VAL2
      KEY3: VAL3    
    KEY4:
      - val4
      - val5
 ```
 
 In this case, keys 0-3 could be merged with the default dict (aka. default args). But `KEY4` would completely override `KEY4` in the default dict. This would leave a list of `val4, val5` only. This matters for `run_farm_hosts` since that is also a list. Thus you need to re-specify all types/amts-to-spawn (in AWS case) if you override.

File: deploy/sample-backup-configs/sample_config_runtime.yaml [hunk 16]
```diff
@@ -1,17 +1,13 @@
...
+  defaults: sample-run-farm-recipes/aws_ec2.yaml
+  # Uncomment and add args to override defaults
+  #override_args:
+  #  <ARG>: <OVERRIDE>
```
abejgonzalez, line 9: See https://github.com/firesim/firesim/pull/1028/files#r885857553 for how overrides work.

deploy/runtools/runtime_config.py
```diff
@@ -138,26 +155,47 @@ def array_to_lognames(values, prefix):
...
+        permissive_driver_args += command_niclogs
+        permissive_driver_args += command_blkdev_logs
+        permissive_driver_args += [f"{tracefile}", f"+trace-select={tracerv_config.select}", f"+trace-start={tracerv_config.start}", f"+trace-end={tracerv_config.end}", f"+trace-output-format={tracerv_config.output_format}", dwarf_file_name]
+        permissive_driver_args += [f"+autocounter-readrate={autocounter_config.readrate}", autocounterfile]
+        permissive_driver_args += [command_dromajo]
+        permissive_driver_args += [print_cycle_prefix, f"+print-start={synthprint_config.start}", f"+print-end={synthprint_config.end}"]
+        permissive_driver_args += command_linklatencies
+        permissive_driver_args += command_netbws
+        permissive_driver_args += command_shmemportnames
+        driver_call = f"""sudo ./{driver} +permissive {" ".join(permissive_driver_args)} +permissive-off {" ".join(command_bootbinaries)}"""
```
abejgonzalez, line 188: Not relevant to this PR (but to the Vitis work later). Do we need `sudo` for the driver?

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,424 @@
...
+        This deliberately exposes sim_host_handles to users, so that if
+        they know exactly how to map to a particular platform, they always
+        have an escape hatch."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_default_switch_host_handle(self) -> str:
+        """Get the default host instance type name that can host switch
+        simulations. """
+        raise NotImplementedError
```
abejgonzalez, line 136: These functions are used by topologies to map `FireSimServerNodes` (and the like) to specific run farm hosts. Do we like this naming? Does this naming make sense?

abejgonzalez: Right now this API works with "handles" which can be anything. In the AWS EC2 case, this "handle" is a name of an inst like `f1.2xlarge` while in the externally provisioned case it is a hostname. Do we like "handle" as the name?

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -0,0 +1,614 @@
...
+from typing import List, Dict, Optional, Union, TYPE_CHECKING
+if TYPE_CHECKING:
+    from mypy_boto3_ec2.service_resource import Instance as EC2InstanceResource
+    from runtools.firesim_topology_elements import FireSimSwitchNode, FireSimServerNode
+    from runtools.run_farm import Inst
+    from awstools.awstools import MockBoto3Instance
+
+rootLogger = logging.getLogger()
+
+class InstanceDeployManager(metaclass=abc.ABCMeta):
```
abejgonzalez, line 25: In other words, this class does everything after being given a hostname to access an instance.

deploy/sample-backup-configs/sample_config_runtime.yaml
davidbiancolin: Make sure you capture that in the docs.

davidbiancolin: my 2c: 

> * `run_farm_config`? `run_farm`? `run_farm_recipe`?
`run_farm`
> * `defaults`? `default_recipe`? `recipe`?
I personally like `base` or `base_recipe`. 
> * `override_args`? `overrides`?
`overrides`

deploy/sample-run-farm-recipes/aws_ec2.yaml
```diff
@@ -0,0 +1,24 @@
+run_farm_type: AWSEC2F1
```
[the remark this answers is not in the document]
davidbiancolin: I think removing the `sample` bit wold be nice since, unlike the other sample files, which are copied and modified, these will see active and direct use. So maybe: `build-farm-recipes` and `run-farm-recipes`

deploy/sample-run-farm-recipes/aws_ec2.yaml
```diff
@@ -0,0 +1,24 @@
...
+args:
+    # REQUIRED: tag to apply to run farm hosts
```
davidbiancolin, line 3: If everything is required, maybe we can just remove the "REQUIRED: prefix" put a header comment saying all of the fields below must be defined in a recipe.yaml, but can be selectively overriden in a runtime-config.yaml?

deploy/util/filelineswap.py
```diff
@@ -0,0 +1,50 @@
...
+        in_file: input file to use as a base
+        out_file: output file to write into
+        start_marker: use this to determine start of replace area
+        end_marker: use this to determine end of replace area
+        inject_lines: lines to inject in the replacement area
+    """
+    with open(in_file, "r") as f:
+        og_lines = f.readlines()
+
+    with open(out_file, "w") as f:
```
davidbiancolin, line 35: Can you make this a little bit stricter and ensure the existence of exactly one start and end marker string.

davidbiancolin (commented):
Mostly bike shedding.

davidbiancolin:
I would like to see the config generation stuff be a bit better abstracted for different platforms now, but in the interest of time lets push that to a follow up PR.

deploy/sample-run-farm-recipes/aws_ec2.yaml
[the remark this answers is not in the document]
sagark: +1 to what @davidbiancolin said

deploy/firesim
```diff
@@ -125,9 +126,30 @@ def managerinit(args: argparse.Namespace):
             m = local("""cp sample-backup-configs/sample_config_{}.yaml config_{}.yaml""".format(conf_file, conf_file), capture=True)
             rootLogger.debug(m)
             rootLogger.debug(m.stderr)
-            m = local("""sed -i 's/AWSUSERNAME/{}/g' config_{}.yaml""".format(get_aws_userid(), conf_file), capture=True)
-            rootLogger.debug(m)
-            rootLogger.debug(m.stderr)
+
+            if args.platform == 'f1':
+                m = local("""sed -i 's/AWSUSERNAME/{}/g' config_{}.yaml""".format(get_aws_userid(), conf_file), capture=True)
```
sagark, line 131: I thought there was a PR somewhere that removed the need for this?

i.e. we were just going to append the ID to the bucket name automatically with an enable/disable flag.

deploy/sample-run-farm-recipes/aws_ec2.yaml
sagark: +1

deploy/sample-backup-configs/sample_config_runtime.yaml
sagark: I also like `run_farm` and `base_recipe`, but for `override_args`, I would prefer something more verbose, e.g. `recipe_arg_overrides`.

Also, I would add a comment that emphasizes that the expected structure of the contents of `overrides` is identical to that of the recipe you're overriding parts of.

deploy/runtools/runtime_config.py
```diff
@@ -138,26 +155,47 @@ def array_to_lognames(values, prefix):
...
         # TODO supernode support
         dwarf_file_name = "+dwarf-file-name=" + all_bootbinaries[0] + "-dwarf"
 
-        # TODO: supernode support (tracefile, trace-select.. etc)
-        basecommand = f"""screen -S fsim{slotid} -d -m bash -c "script -f -c 'stty intr ^] && sudo ./{driver} +permissive $(sed \':a;N;$!ba;s/\\n/ /g\' {runtimeconf}) +slotid={slotid} +profile-interval={profile_interval} {zero_out_dram} {disable_asserts} {command_macs} {command_rootfses} {command_niclogs} {command_blkdev_logs}  {tracefile} +trace-select={trace_select} +trace-start={trace_start} +trace-end={trace_end} +trace-output-format={trace_output_format} {dwarf_file_name} +autocounter-readrate={autocounter_readrate} {autocounterfile} {command_dromajo} {print_cycle_prefix} +print-start={print_start} +print-end={print_end} {command_linklatencies} {command_netbws}  {command_shmemportnames} +permissive-off {command_bootbinaries} && stty intr ^c' uartlog"; sleep 1"""
-
-        return basecommand
-
+        screen_name = "fsim{}".format(slotid)
+        run_device_placement = "+slotid={}".format(slotid) if self.platform == "f1" else "+device_index={}".format(slotid)
```
sagark, line 166: Why are these plusargs named differently? Why not just pick one?

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,424 @@
...
-        super().__init__()
-        self.instance_id = M4_16.instance_counter
-        M4_16.instance_counter += 1
+    @abc.abstractmethod
+    def terminate_run_farm(self, terminate_some_dict: Dict[str, int], forceterminate: bool) -> None:
+        """Terminate run hosts for simulations.
 
-class RunFarm:
+        Args:
+            terminate_some_dict: Dict of run host names to amount of that type to terminate.
```
sagark, line 157:
```suggestion
            terminate_some_dict: Dict of run host handles to amount of that type to terminate.
```

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -584,17 +551,17 @@ def loop_logger(instancestates: Dict[str, Any], terminateoncompletion: bool) ->
             rootLogger.info("Instances")
             rootLogger.info("-"*80)
             for instance in instancestate_map.keys():
-                rootLogger.info("""Instance IP:{:>15} | Terminated: {}""".format(instance, truefalsecolor[instancestate_map[instance]]))
+                rootLogger.info("""Inst. hostname:{:>15} | Terminated: {}""".format(instance, truefalsecolor[instancestate_map[instance]]))
```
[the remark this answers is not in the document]
sagark: I would go with "Hostname/IP".

Also, unlike IPs we don't have a fixed max hostname length, so we should probably pad to the length of the longest hostname?

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,424 @@
...
+
+    Attributes:
+        MAX_SWITCH_SLOTS_ALLOWED: max switch slots allowed (hardcoded)
+        switch_slots: switch node slots
+        _next_switch_port: next switch port to assign
+        MAX_SIM_SLOTS_ALLOWED: max simulations allowed. given by `config_runfarm.yaml`
+        sim_slots: simulation node slots
+        sim_dir: name of simulation directory on the run host
+        instance_deploy_manager: platform specific implementation
+        hostname: hostname of the instance (normally set by the `RunFarm` type)
```
sagark, line 39: throughout this codebase, isn't this really `hostname_or_ip`?

deploy/firesim
sagark: The mapping of platform values from args.platform to run farm recipe yamls seems weird, but I guess we can fix later.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -191,33 +185,22 @@ def pass_no_net_host_mapping(self) -> None:
...
 
     def pass_simple_networked_host_node_mapping(self) -> None:
         """ A very simple host mapping strategy.  """
         switches = self.firesimtopol.get_dfs_order_switches()
-        f1_2s_used = 0
-        f1_4s_used = 0
-        f1_16s_used = 0
-        m4_16s_used = 0
+
+        switch_host_inst_type = self.run_farm.get_default_switch_host_handle()
```
sagark, line 203: this should be `switch_host_inst_handle` throughout the code?

deploy/sample-backup-configs/sample_config_runtime.yaml
abejgonzalez: I tend to agree that "args" should also be present since I want to emphasize that you can only override the "args" section of the run farm recipe. I have little preference on the "recipe_*" or just "arg_overrides"

deploy/firesim
abejgonzalez: I'm not sure there is another way to implement this in the meantime... you need to know what run farm recipe corresponds to the platform that you setup for.

deploy/sample-backup-configs/sample_config_runtime.yaml
abejgonzalez: Also updated the comment on `run_farm_type`. This will be made clear in the follow on documentation.

deploy/runtools/firesim_topology_with_passes.py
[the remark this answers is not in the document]
abejgonzalez: I agree but I'll leave this as a follow on PR.

deploy/firesim
abejgonzalez: Thats part of this PR: https://github.com/firesim/firesim/pull/1046. I think the plan was to merge this first then revisit it. IIRC, we wanted to figure out where the recipes for the different configurations would reside (as part of the manager config files or separately) before merging it in.

deploy/sample-run-farm-recipes/aws_ec2.yaml
```diff
@@ -0,0 +1,24 @@
...
+    # REQUIRED: if using spot instances, determine the interrupt behavior
+    spot_interruption_behavior: terminate
+    # REQUIRED: if using spot instances, determine the max price
+    spot_max_price: ondemand
+    # REQUIRED: default location of the simulation directory on the run farm host
+    default_simulation_dir: /home/centos
+    # REQUIRED: run farm hosts to spawn
+    run_farm_hosts:
+      # fpga instances
+      - f1.16xlarge: 1
```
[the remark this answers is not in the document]
abejgonzalez: No preference. I just kept this to what the current defaults are.

deploy/runtools/runtime_config.py
abejgonzalez: Ah. Good catch. This needs to get removed since we don't support Vitis in this PR.

deploy/runtools/run_farm.py
abejgonzalez: My mistake, I thought a hostname encompassed an IP address. Since they are separate things I'll update the function name.

deploy/runtools/firesim_topology_with_passes.py
abejgonzalez: I wanted to see if "handle" was accepted before changing it everywhere.

deploy/firesim
sagark: ah okay yeah, thanks

deploy/firesim
sagark: I meant more that we call a single platform many different things:

- f1
- AWSEC2
- aws_ec2
...

We should probably pick one, but let's just fix it after this PR.

deploy/util/filelineswap.py
sagark: I meant without parsing (just tracking indent/path to root as you go through the lines). You can just leave this as is and I can update it in the metasim PR.

deploy/sample-run-farm-recipes/externally_provisioned.yaml
```diff
@@ -0,0 +1,24 @@
+# Unmanaged list of run farm hosts. Assumed that they are pre-setup to run simulations.
+run_farm_type: ExternallyProvisioned
+args:
+    # REQUIRED: defaults used for all run farm hosts
+    default_num_fpgas: 4
+    default_platform: EC2InstanceDeployManager
```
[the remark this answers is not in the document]
timsnyder-siv: If it is correct, then what is `default_platform` doing?  Add a comment.

timsnyder-siv (commented):
Looks like it is getting close.

deploy/runtools/run_farm.py
```diff
@@ -364,537 +427,163 @@ def terminate_run_farm(self, terminatesomef1_16: int, terminatesomef1_4: int, te
...
+        assert False, "Unable to return run host to host switches. Make sure at least one run host is available with no FPGAs in use."
+
+    def post_launch_binding(self, mock: bool = False) -> None:
+        return
+
+    def launch_run_farm(self) -> None:
+        return
+
+    def terminate_run_farm(self, terminate_some_dict: Dict[str, int], forceterminate: bool) -> None:
+        return
```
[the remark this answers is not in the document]
abejgonzalez: Done. LMK if you want a cleaner print.

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,424 @@
...
+        (according to run host counts you've specified in config_run_farm.ini).
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def allocate_sim_host(self, sim_host_handle: str) -> Inst:
+        """Let user allocate and use an run host (assign sims, etc.).
+        This deliberately exposes sim_host_handles to users, so that if
+        they know exactly how to map to a particular platform, they always
+        have an escape hatch."""
```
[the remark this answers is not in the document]
abejgonzalez: I'll update this a bit with what I think Sagar meant.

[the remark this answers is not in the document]
abejgonzalez: Actually I'll just remove this.

deploy/runtools/run_farm.py
abejgonzalez: I like this since it matches the terminology of Fabric.

deploy/runtools/run_farm.py
```diff
@@ -364,537 +427,163 @@ def terminate_run_farm(self, terminatesomef1_16: int, terminatesomef1_4: int, te
...
+        """ Let user allocate and use an instance (assign sims, etc.).
+        This deliberately exposes sim_host_handles to users, so that if
+        they know exactly how to map to a particular platform, they always
+        have an escape hatch. """
+        inst_ret = self.run_farm_hosts_dict[sim_host_handle]
+        assert self.mapper_consumed[sim_host_handle] == False, "{sim_host_handle} is already allocated."
+        self.mapper_consumed[sim_host_handle] = True
+        return inst_ret
+
+    def get_default_switch_host_handle(self) -> str:
```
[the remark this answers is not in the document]
abejgonzalez: I don't support networked sims (mainly because I never tested them - I figure this can be punted to later). The logic in this function shouldn't be used since networked sims are unsupported, but the logic here is sufficient for a basic networked sim case (though it can be improved).

deploy/runtools/run_farm.py
abejgonzalez: I updated the docstrings. LMK if they need more clarification.

deploy/runtools/runtime_config.py
abejgonzalez: Punting the description to the documentation PR.

deploy/sample-backup-configs/sample_config_runtime.yaml
abejgonzalez: Settled on `run_farm, base_recipe, recipe_arg_overrides`. Closing

deploy/sample-run-farm-recipes/externally_provisioned.yaml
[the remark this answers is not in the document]
abejgonzalez: Yup. This is correct. This class basically details an AWS EC2 instance and how to run simulations on it. Added more to the comments here. LMK if this is clear enough.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -325,18 +326,22 @@ def copy_back_job_results_from_run(self, slotno: int) -> None:
...
             is_qcow2 = rfsname.endswith(".qcow2")
-            mountpoint = """/home/centos/sim_slot_{}/mountpoint""".format(simserverindex)
+            mountpoint = """{}/sim_slot_{}/mountpoint""".format(dest_sim_dir, simserverindex)
             with StreamLogger('stdout'), StreamLogger('stderr'):
                 run("""sudo mkdir -p {}""".format(mountpoint))
 
                 if is_qcow2:
-                    rfsname = self.get_host_instance().nbd_tracker.get_nbd_for_imagename(rfsname)
+                    host_inst = self.get_host_instance()
+                    assert isinstance(host_inst.instance_deploy_manager, EC2InstanceDeployManager)
```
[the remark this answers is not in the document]
abejgonzalez: Ideally yes, but for now we only support the NBD stuff on AWS EC2 (other deploy managers will not have this by default). Therefore, this has to be specific to the EC2InstanceDeployManager.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -476,9 +480,9 @@ def allocate_nbds(self) -> None:
 
         for rootfsname in rootfses_list:
             if rootfsname is not None and rootfsname.endswith(".qcow2"):
-                allocd_device = self.get_host_instance().nbd_tracker.get_nbd_for_imagename(rootfsname)
-
-
+                host_inst = self.get_host_instance()
+                assert isinstance(host_inst.instance_deploy_manager, EC2InstanceDeployManager)
```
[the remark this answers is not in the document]
abejgonzalez: See my other comment.

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,424 @@
...
 
-        self.launch_timeout = launch_timeout
-        self.always_expand = always_expand
+    SIM_HOST_HANDLE_FOR_SWITCH_ONLY_SIM: str = 'm4.16xlarge'
 
-    def bind_mock_instances_to_objects(self) -> None:
-        """ Only used for testing. Bind mock Boto3 instances to objects. """
-        for index in range(len(self.f1_16s)):
-            self.f1_16s[index].assign_boto3_instance_object(MockBoto3Instance())
+    SORTED_SIM_HOST_HANDLE_TO_MAX_FPGA_SLOTS: List[Tuple[int, str]]
```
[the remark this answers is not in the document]
abejgonzalez: These are constants correct. However, they should be consistent with one another. I followed your other comment to DRY this out a bit.

.github/scripts/local_flags.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+RUN_LOCAL = False # change to True to run CI scripts locally
```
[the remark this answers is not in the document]
abejgonzalez: Done.

sagark (approved):
LGTM!

deploy/runtools/firesim_topology_elements.py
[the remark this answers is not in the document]
davidbiancolin: I'd prefer if the DeployManager explicitly called out if it supported QCOW images, (e.g., it could have a member indicating that), or we just rip out the unused qcow support altogether.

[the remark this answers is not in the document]
abejgonzalez: Or if its better, I can do this in a follow up PR so that this PR gets in an unblocks Sagars meta-sim PR.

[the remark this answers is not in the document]
davidbiancolin: You can do it in a separate PR.

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,397 @@
...
-            self.m4_16s[index].assign_boto3_instance_object(instance)
-
-        for index, instance in enumerate(available_f1_2_instances):
-            self.f1_2s[index].assign_boto3_instance_object(instance)
-
+        for sim_host_handle in self.SIM_HOST_HANDLE_TO_MAX_FPGA_SLOTS:
+            for index, instance in enumerate(available_instances_per_handle[sim_host_handle]):
+                inst_tup = self.run_farm_hosts_dict[sim_host_handle][index]
+                inst = inst_tup[0]
+                inst.set_host("centos@" + instance.private_ip_address)
```
timsnyder-siv, line 326: is adding the `centos@` prefix here really required?  Have we always done this? I don't see it in the old code and I'm concerned that it will break `RunFarm.lookup_by_host()`.

deploy/runtools/runtime_config.py
```diff
@@ -295,42 +323,26 @@ def __init__(self, runtimeconfigfile: str, configoverridedata: str) -> None:
...
+            self.hostdebug_config.zero_out_dram = runtime_dict['host_debug'].get('zero_out_dram') == "yes"
+            self.hostdebug_config.disable_synth_asserts = runtime_dict['host_debug'].get('disable_synth_asserts') == "yes"
+        self.synthprint_config = SynthPrintConfig()
         if 'synth_print' in runtime_dict:
-            self.print_start = runtime_dict['synth_print'].get("start", "0")
-            self.print_end = runtime_dict['synth_print'].get("end", "-1")
-            self.print_cycle_prefix = runtime_dict['synth_print'].get("cycle_prefix", "yes") == "yes"
+            self.synthprint_config.start = runtime_dict['synth_print'].get("start", "0")
+            self.synthprint_config.end = runtime_dict['synth_print'].get("end", "-1")
+            self.synthprint_config.cycle_prefix = runtime_dict['synth_print'].get("cycle_prefix", "yes") == "yes"
```
timsnyder-siv, line 345: nit. In a future PR, it would probably be good to push the initialization of these configs into their respective `__init__` methods.

timsnyder-siv (commented):
There are a number of minor things that shouldn't block merging that I commented on.  The only two things that I think need to be discussed before I approve are:
* are you confident in the `AWSEC2F1::allocate_sim_host()` logic with regard to always incrementing `self.mapper_consumed[sim_host_handle]`?  And can you add a docstring for `mapper_consumed`?
* are we certain that we want to hardcode adding `centos@` prefix to the host ip address in `AWSEC2F1`?

.github/scripts/run-linux-poweroff-externally-provisioned.py
```diff
@@ -0,0 +1,91 @@
...
+                with settings(warn_only=True):
+                    # do the following:
+                    #   1. launch the run farm w/ the AWS EC2 runfarm
+                    #   2. copy the hostnames given into a new externally provisioned runfarm/runtime file
+                    #   4. run launch/infra/runworkload/terminate w/ that runtime
+                    #   5. if successful or fail, run the terminate w/ the old AWS EC2 runfarm
+
+                    rc = run(f"firesim launchrunfarm -c {workload_full}")
+
+                    time.sleep(3 * 60) # TODO: replace w/ instance_liveness check
```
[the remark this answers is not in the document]
abejgonzalez: This is punted to a later PR since this is very low priority and adds 0 time to testing (since the AWS EC2 FPGA CI takes longer to run than this).

.github/scripts/run-linux-poweroff.py
```diff
@@ -22,7 +19,7 @@ def run_w_timeout(workload, timeout):
             """
             log_tail_length = 100
             # rename runfarm tag with a unique tag based on the ci workflow
-            with prefix(f"export FIRESIM_RUNFARM_PREFIX={ci_workflow_run_id}"):
+            with prefix(f"export FIRESIM_RUNFARM_PREFIX={ci_workflow_run_id}-default"):
```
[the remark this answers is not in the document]
abejgonzalez: Maybe this would make more sense if I made this automatically determined based off the file name + CI ID so that you don't have to make up a postfix? The postfix is needed so that the run farm name doesn't collide between the two different `run-linux-*.py` scripts.

deploy/run-farm-recipes/aws_ec2.yaml
```diff
@@ -0,0 +1,27 @@
+# AWS EC2 run farm hosts recipe.
+# all fields are required but can be overridden in the `*_runtime.yaml`
+
+run_farm_type: AWSEC2F1
+args:
+    # tag to apply to run farm hosts
+    run_farm_tag: mainrunfarm
+    # TODO
+    always_expand_run_farm: true
+    # TODO
```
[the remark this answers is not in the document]
abejgonzalez: All the comments could probably be updated so I'll leave this to a later PR (probably when I updated the docs).

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,397 @@
...
+        assert len(self.switch_slots) < self.MAX_SWITCH_SLOTS_ALLOWED
         self.switch_slots.append(firesimswitchnode)
         firesimswitchnode.assign_host_instance(self)
 
     def allocate_host_port(self) -> int:
         """ Allocate a port to use for something on the host. Successive calls
         will return a new port. """
-        retport = self._next_port
+        retport = self._next_switch_port
         assert retport < 11000, "Exceeded number of ports used on host. You will need to modify your security groups to increase this value."
```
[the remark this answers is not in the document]
abejgonzalez: Agreed, but this is old code untouched by me so I'm gonna punt this to later.

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,397 @@
...
+                continue
+            return sim_host_handle
+
+        rootLogger.critical(f"ERROR: No instances are available to satisfy the request for an instance with support for {num_sims} simulation slots. Add more instances in your run farm configuration (e.g., config_run_farm.ini).")
+        raise Exception
+
+    def allocate_sim_host(self, sim_host_handle: str) -> Inst:
+        inst_tup = self.run_farm_hosts_dict[sim_host_handle][self.mapper_consumed[sim_host_handle]]
+        inst_ret = inst_tup[0]
+        self.mapper_consumed[sim_host_handle] += 1
```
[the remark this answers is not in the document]
abejgonzalez: I think this mapper is oblivious to the number of slots used on the instance. With this API, the mapping of slots to simulations is done in the user topology so its up to the user to fill the slots (or not). In short, this `+= 1` is fine.

deploy/runtools/run_farm.py
```diff
@@ -1,361 +1,397 @@
...
+    def get_default_switch_host_handle(self) -> str:
+        return self.SIM_HOST_HANDLE_FOR_SWITCH_ONLY_SIM
+
+    def bind_mock_instances_to_objects(self) -> None:
+        """ Only used for testing. Bind mock Boto3 instances to objects. """
+        for inst_handle, inst_list in self.run_farm_hosts_dict.items():
+            for idx, run_farm_host_tup in enumerate(inst_list):
+                boto_obj = MockBoto3Instance()
+                inst = run_farm_host_tup[0]
+                inst.set_host("centos@" + boto_obj.private_ip_address)
```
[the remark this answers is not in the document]
abejgonzalez: I added centos everywhere here to make it clear what was running on centos (vs. not). I can remove this in the places that I added it so that it matches old functionality.

deploy/runtools/run_farm.py
abejgonzalez: See other comment - Basically no. No this wasn't done before, I'll revert this. This doesn't break the lookup by host though.

deploy/runtools/run_farm.py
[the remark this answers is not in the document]
timsnyder: I guess this is another reason why having a docstring for `mapper_consumed` would be really nice because I'm not understanding how it works.

.github/scripts/run-linux-poweroff.py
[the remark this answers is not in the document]
timsnyder: Ok.  Yeah, it would be good to make it use `__file__` in the runfarm prefix to make it clearer what you're doing.

deploy/runtools/run_farm.py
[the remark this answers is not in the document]
timsnyder: I think it's probably a good idea to revert it.

timsnyder-siv (approved):
I'm confident that the rest of my feedback can be done in a followon PR.

abejgonzalez:
Ok assuming CI passes I clicked auto-merge!

PR #1029, opened 2022-04-15
Port Python Flags to Abseil

abejgonzalez:
This switches the Python manager flags to use Abseil. While this work doesn't make the flags seem super useful yet, it does the following:
- Flags are now distributed into each `.py` file instead of on the top level. This avoids the need to propagate flags from the top level into the class that needs it. (this becomes very useful if each "build/run-farm" needs its own flags - no need to modify the `firesim` file)
- This allows for more verbose `--help` (and `--helpful`).
- Allows nifty features in the future (like flag validators and the ability to load flags from a file)

See https://abseil.io/docs/python/guides/flags for info on how flags are potentially useful.

A future PR will change the C++ APIs to Abseil (if this work is viewed as helpful)

#### UI / API Impact

Technically, the help text is different and there are a couple new flags that you can use (see `--helpfull` for more info)

N/A

### Contributor Checklist


abejgonzalez:
People say to wait on this until after local fpga changes are merged

timsnyder-siv:
The only alternative to absl that is worth considering is probably click.  I see that it has shell completion. https://click.palletsprojects.com/en/7.x/bashcomplete/

timsnyder-siv:
I did some more searching for "absl shell completion" and didn't find an existing project with support for it.  I still maintain that it shouldn't be too hard to extend argcomplete for absl.

Slightly related, I did find https://github.com/iterative/shtab which looks like an interesting alternative to argcomplete.  It:
* supports zsh without having to add pyzshcomplete
* removes the need to dynamically execute the firesim manager for each completion (this also comes with limitations on the types of completion that can be supported but we aren't doing anything complicated enough to not be supported yet)

abejgonzalez:
@mergifyio rebase

abejgonzalez:
Turns out argcomplete works for this implementation of Abseil! This is what I see:

image

abejgonzalez:
You are right. It doesn't complete `--FLAGS`.

abejgonzalez:
Closing since this doesn't work well with arg completion.

PR #1031, opened 2022-04-20
Adding Termination Bridge with target defined conditions and messaging.

rahulb10:
This PR provides bridge for generalized exit conditions

#### UI / API Impact
Doesn't impact existing APIs

Doesn't impact Verilog/AGFI Compatibility

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
@@ -1,0 +1,133 @@
+
+package firesim.bridges
+
+import chisel3._
+import chisel3.util._
+import freechips.rocketchip.config.Parameters
+import midas.widgets._
+import midas.targetutils._
+import freechips.rocketchip.util.{DecoupledHelper}
+
+/**
+  * Defines a condition under which the simulator should halt. 
+  * @param Message A string printed by the driver to indicate why the simulator is terminating.
+  * @param isError When true, instructs the driver to return a non-zero value when exiting under this condition.
+**/
+case class TerminationCondition(isErr: Boolean, message : String)
+
+/** A seq of termination conditions one for each concurrently running program
+ */
+case class TerminationBridgeParams(
+  conditionInfo: Seq[TerminationCondition]
+)                
+
+class TerminationBridgeTargetIO(params: TerminationBridgeParams) extends Bundle {
+  val clock = Input(Clock())
+  val errCode = Input(UInt((log2Ceil((params.conditionInfo).size)).W))
+  val valid = Input(Bool()) 
+}
+
+class TerminationBridgeHostIO(params: TerminationBridgeParams)
+                              (private val targetIO : TerminationBridgeTargetIO = new TerminationBridgeTargetIO(params)) 
+                               extends Bundle with ChannelizedHostPortIO{
+  def targetClockRef = targetIO.clock 
+  //It is just inputs from the target to indicate test completion
+  //The HostPort annotation would not work with aggregates so creating individual channels.
+  val errCodeH = InputChannel(targetIO.errCode)
+  val validH = InputChannel(targetIO.valid)
+}
+
+class TerminationBridge(params: TerminationBridgeParams) extends BlackBox
+    with Bridge[TerminationBridgeHostIO, TerminationBridgeModule] {
+  val io = IO(new TerminationBridgeTargetIO(params))
+  val bridgeIO = new TerminationBridgeHostIO(params)(io)
+
+  val constructorArg = Some(params)
+
+  generateAnnotations()
+}
+
+object TerminationBridge {
+
+  private def annotateTerminationBridge(clock: Clock, reset: Reset, conditionBools: Seq[Bool], params: TerminationBridgeParams): TerminationBridge = {
+    val finish = conditionBools.reduce(_||_)
+    val errMessageID = PriorityEncoder(conditionBools)
+    //val gReset = WireInit(false.B)
+    //GlobalResetCondition.sink(gReset)
+    require(params.conditionInfo.size==conditionBools.size)
+    val terminationBridgeTarget = Module(new TerminationBridge(params))
+    terminationBridgeTarget.io.errCode := errMessageID
+    terminationBridgeTarget.io.valid := finish && !reset.asBool
+    terminationBridgeTarget.io.clock := clock 
+    terminationBridgeTarget  
+  }
+
+/** Instatiates the target side of the Bridge, selects one of 
+ *  the available finish conditions and passes on the corresponding 
+ *  error message.
+ *  @param clock: Clock to the bridge which it must run on.
+ *  @param reset: local reset for the bridge.
+ *  @param conditionBools: Seq of finished conditions indicated by running programs
+ *  @param params: Possible list of message info to be returned by simulator  
+ */
+
+  def apply(clock: Clock, reset: Reset, conditionBools: Seq[Bool], params: TerminationBridgeParams): TerminationBridge = {
+    annotateTerminationBridge(clock, reset, conditionBools, params)
+  }
+
+/** Simpler way to instantiate target side of the bridge by using implicit clock and reset
+ */
+  def apply(conditionBools: Seq[Bool], params: TerminationBridgeParams): TerminationBridge = {
+    annotateTerminationBridge(Module.clock, Module.reset, conditionBools, params)
+  }
+
+}
+
+class TerminationBridgeModule(params: TerminationBridgeParams)(implicit p: Parameters) extends BridgeModule[TerminationBridgeHostIO]()(p) {
+  lazy val module = new BridgeModuleImp(this) {
+    // This creates the interfaces for all of the host-side transport
+    // AXI4-lite for the simulation control bus, =
+    // AXI4 for DMA
+    val io = IO(new WidgetIO())
+
+
+
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TerminationBridgeHostIO(params)())
+    
+    val statusDone = Module(new Queue(Bool(), 1))
+    val errCode = Module(new Queue(hPort.errCodeH.bits.cloneType, 1))
+
+    val cycleCountWidth = 64 //Data width is capped by nastikey 
+
+    val tokenCounter = genWideRORegInit(0.U(cycleCountWidth.W), "out_counter")    
+
+
+    statusDone.io.enq <> hPort.validH
+    errCode.io.enq <> hPort.errCodeH 
+
+    val noTermination = !statusDone.io.deq.bits
+
+    val tFireHelper = DecoupledHelper(errCode.io.deq.valid, statusDone.io.deq.valid, noTermination)
+
+    when(tFireHelper.fire()) {
+      tokenCounter := tokenCounter + 1.U
+    }
+
+    statusDone.io.deq.ready := tFireHelper.fire(statusDone.io.deq.valid)
+    errCode.io.deq.ready := tFireHelper.fire(errCode.io.deq.valid)
+
+    genROReg(statusDone.io.deq.bits && statusDone.io.deq.valid, "out_status")
+    genROReg(errCode.io.deq.bits, "out_errCode")
+
+    override def genHeader(base: BigInt, sb: StringBuilder) {
+      import CppGenerationUtils._
+      val headerWidgetName = getWName.toUpperCase
+      super.genHeader(base, sb)
+      sb.append(genConstStatic(s"${headerWidgetName}_message_count", UInt32((params.conditionInfo).size)))
+      sb.append(genArray(s"${headerWidgetName}_message_type", (params.conditionInfo).map(x => UInt32(if(x.isErr) 1 else 0))))
+      sb.append(genArray(s"${headerWidgetName}_message", (params.conditionInfo).map(x => CStrLit(x.message))))
+    }
+    genCRFile()
+  }
+}
--- a/sim/midas/src/main/cc/bridges/termination.cc
+++ b/sim/midas/src/main/cc/bridges/termination.cc
@@ -1,0 +1,66 @@
+#ifdef TERMINATIONBRIDGEMODULE_struct_guard
+
+#include "termination.h"
+#include <iostream>
+
+termination_t::termination_t(
+  simif_t* sim, 
+  std::vector<std::string> &args,
+  TERMINATIONBRIDGEMODULE_struct * mmio_addrs,
+  unsigned int num_messages,
+  unsigned int* isErr,
+  const char* const* msgs): 
+  bridge_driver_t(sim), 
+  mmio_addrs(mmio_addrs), 
+  num_messages(num_messages), 
+  isErr(isErr),
+  msgs(msgs)
+{
+  //tick-rate to decide sampling rate of MMIOs per number of ticks
+  this->mmio_addrs = mmio_addrs;
+  std::string tick_rate_arg = std::string("+tick-rate=");
+  for (auto &arg: args) {
+    if (arg.find(tick_rate_arg) == 0) {
+      char *str = const_cast<char*>(arg.c_str()) + tick_rate_arg.length();
+      uint64_t tick_period = atol(str);
+      this->tick_rate = tick_period;
+    }
+  }
+}
+
+termination_t::~termination_t() {
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs tick-rate
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_errCode);
+    	this->fail = this->isErr[msg_id];
+      test_done = true;
+			if(msg_id < this->num_messages) {
+    	  std::cerr << "Termination Bridge detected exit on cycle " << this->cycleCount()
+			  << " with message " << this->msgs[msg_id] << std::endl;
+			} else {
+			  std::cerr << "Unknown Termination message" << std::endl;
+			}
+    }
+    tick_counter = 0;
+  } else {
+  	tick_counter +=1;
+  }
+}
+
+const char* termination_t::exit_message() {
+  int msg_id = read(this->mmio_addrs->out_errCode);
+  return this->msgs[msg_id];
+}
+
+int termination_t::cycleCount() {
+  write(this->mmio_addrs->out_counter_latch, 1);
+  data_t cycle_l = read(this->mmio_addrs->out_counter_0);
+  data_t cycle_h = read(this->mmio_addrs->out_counter_1);
+  return (((uint64_t) cycle_h) << 32) | cycle_l;
+}
+
+#endif
--- a/sim/midas/src/main/cc/bridges/termination.cc
+++ b/sim/midas/src/main/cc/bridges/termination.cc
@@ -17,7 +17,7 @@ termination_t::termination_t(
   msgs(msgs)
 {
   //tick-rate to decide sampling rate of MMIOs per number of ticks
-  std::string tick_rate_arg = std::string("+tick-rate=");
+  std::string tick_rate_arg = std::string("+termination-bridge-tick-rate=");
   for (auto &arg: args) {
     if (arg.find(tick_rate_arg) == 0) {
       char *str = const_cast<char*>(arg.c_str()) + tick_rate_arg.length();
@@ -32,17 +32,14 @@ termination_t::~termination_t() {
 }
 
 void termination_t::tick() {  //reads the MMIOs at tick-rate
-  if(tick_counter == tick_rate) {
+  if (tick_counter == tick_rate) {
     if (read(this->mmio_addrs->out_status)) {
       int msg_id = read(this->mmio_addrs->out_terminationCode);
+      assert(msg_id < this->num_messages);
       this->fail = this->is_err[msg_id];
       test_done = true;
-      if(msg_id < this->num_messages) {
-        std::cerr << "Termination Bridge detected exit on cycle " << this->cycle_count()
-        << " with message " << this->msgs[msg_id] << std::endl;
-      } else {
-        std::cerr << "Unknown Termination message" << std::endl;
-      }
+      std::cerr << "Termination Bridge detected exit on cycle " << this->cycle_count()
+      << " with message " << this->msgs[msg_id] << std::endl;
     }
     tick_counter = 0;
   } else {
--- a/sim/midas/src/main/cc/bridges/termination.cc
+++ b/sim/midas/src/main/cc/bridges/termination.cc
@@ -8,16 +8,15 @@ termination_t::termination_t(
   std::vector<std::string> &args,
   TERMINATIONBRIDGEMODULE_struct * mmio_addrs,
   unsigned int num_messages,
-  unsigned int* isErr,
+  unsigned int* is_err,
   const char* const* msgs): 
   bridge_driver_t(sim), 
   mmio_addrs(mmio_addrs), 
   num_messages(num_messages), 
-  isErr(isErr),
+  is_err(is_err),
   msgs(msgs)
 {
   //tick-rate to decide sampling rate of MMIOs per number of ticks
-  this->mmio_addrs = mmio_addrs;
   std::string tick_rate_arg = std::string("+tick-rate=");
   for (auto &arg: args) {
     if (arg.find(tick_rate_arg) == 0) {
@@ -32,31 +31,31 @@ termination_t::~termination_t() {
   free(this->mmio_addrs);
 }
 
-void termination_t::tick() {  //reads the MMIOs tick-rate
+void termination_t::tick() {  //reads the MMIOs at tick-rate
   if(tick_counter == tick_rate) {
     if (read(this->mmio_addrs->out_status)) {
-      int msg_id = read(this->mmio_addrs->out_errCode);
-    	this->fail = this->isErr[msg_id];
+      int msg_id = read(this->mmio_addrs->out_terminationCode);
+      this->fail = this->is_err[msg_id];
       test_done = true;
-			if(msg_id < this->num_messages) {
-    	  std::cerr << "Termination Bridge detected exit on cycle " << this->cycleCount()
-			  << " with message " << this->msgs[msg_id] << std::endl;
-			} else {
-			  std::cerr << "Unknown Termination message" << std::endl;
-			}
+      if(msg_id < this->num_messages) {
+        std::cerr << "Termination Bridge detected exit on cycle " << this->cycle_count()
+        << " with message " << this->msgs[msg_id] << std::endl;
+      } else {
+        std::cerr << "Unknown Termination message" << std::endl;
+      }
     }
     tick_counter = 0;
   } else {
-  	tick_counter +=1;
+    tick_counter +=1;
   }
 }
 
 const char* termination_t::exit_message() {
-  int msg_id = read(this->mmio_addrs->out_errCode);
+  int msg_id = read(this->mmio_addrs->out_terminationCode);
   return this->msgs[msg_id];
 }
 
-int termination_t::cycleCount() {
+int termination_t::cycle_count() {
   write(this->mmio_addrs->out_counter_latch, 1);
   data_t cycle_l = read(this->mmio_addrs->out_counter_0);
   data_t cycle_h = read(this->mmio_addrs->out_counter_1);
--- a/sim/midas/src/main/cc/bridges/termination.h
+++ b/sim/midas/src/main/cc/bridges/termination.h
@@ -1,0 +1,38 @@
+
+#ifndef __TERMINATION_H
+#define __TERMINATION_H
+//See LICENSE for license details.
+
+#ifdef TERMINATIONBRIDGEMODULE_struct_guard
+
+#include "bridge_driver.h"
+
+class termination_t: public bridge_driver_t
+{
+  public:
+    termination_t(simif_t* sim, 
+    std::vector<std::string> &args,
+    TERMINATIONBRIDGEMODULE_struct * mmio_addrs,
+    unsigned int num_messages,
+    unsigned int* isErr,
+    const char* const* msgs);
+    ~termination_t();
+    virtual void init() {};
+    virtual void tick();
+    virtual void finish() {};
+    virtual bool terminate() { return test_done; };
+    virtual int exit_code() { return fail; };
+    const char* exit_message();
+    int cycleCount();
+  private:
+    TERMINATIONBRIDGEMODULE_struct * mmio_addrs;
+    bool test_done = false;
+    int fail = 0;
+    int tick_rate = 10;
+    int tick_counter = 0;
+    unsigned int num_messages;
+    unsigned int* isErr;
+    const char* const* msgs;
+};
+#endif // TerminationBridgeModule_struct_guard
+#endif //__TERMINATION_H
--- a/sim/midas/src/main/cc/bridges/termination.h
+++ b/sim/midas/src/main/cc/bridges/termination.h
@@ -14,7 +14,7 @@ class termination_t: public bridge_driver_t
     std::vector<std::string> &args,
     TERMINATIONBRIDGEMODULE_struct * mmio_addrs,
     unsigned int num_messages,
-    unsigned int* isErr,
+    unsigned int* is_err,
     const char* const* msgs);
     ~termination_t();
     virtual void init() {};
@@ -23,7 +23,7 @@ class termination_t: public bridge_driver_t
     virtual bool terminate() { return test_done; };
     virtual int exit_code() { return fail; };
     const char* exit_message();
-    int cycleCount();
+    int cycle_count();
   private:
     TERMINATIONBRIDGEMODULE_struct * mmio_addrs;
     bool test_done = false;
@@ -31,7 +31,7 @@ class termination_t: public bridge_driver_t
     int tick_rate = 10;
     int tick_counter = 0;
     unsigned int num_messages;
-    unsigned int* isErr;
+    unsigned int* is_err;
     const char* const* msgs;
 };
 #endif // TerminationBridgeModule_struct_guard
--- a/sim/src/main/cc/midasexamples/TerminationModule.h
+++ b/sim/src/main/cc/midasexamples/TerminationModule.h
@@ -39,9 +39,9 @@ public:
     poke(io_doneErrCode, termination_code);
     step(reset_length);
     poke(reset, 0);
-    if(termination_code %2 !=0) {
+    if (termination_code %2 !=0) {
       msgid = 0;
-    } else if(termination_code %4 != 0) {
+    } else if (termination_code %4 != 0) {
       msgid = 1;
     } else {
       msgid = 2;
--- a/sim/src/main/cc/midasexamples/TerminationModule.h
+++ b/sim/src/main/cc/midasexamples/TerminationModule.h
@@ -1,0 +1,61 @@
+//See LICENSE for license details.
+
+#include "simif_peek_poke.h"
+#include "bridges/termination.h"
+#include <iostream>
+#include <string.h>
+
+class TerminationModule_t: public simif_peek_poke_t
+{
+public:
+  termination_t * terminator;
+  TerminationModule_t(int argc, char** argv) {
+    TERMINATIONBRIDGEMODULE_0_substruct_create;
+    std::vector<std::string> args(argv + 1, argv + argc);
+    terminator = new termination_t(
+      this, args,
+      TERMINATIONBRIDGEMODULE_0_substruct,
+      TERMINATIONBRIDGEMODULE_0_message_count,
+      TERMINATIONBRIDGEMODULE_0_message_type,
+      TERMINATIONBRIDGEMODULE_0_message);
+  };
+  void run() {
+    poke(reset, 1);
+    //poke(io_globalReset, 1);
+    int terminations = 0;
+    int lv_validinCycle = 0;
+    int validinCycle = 0;
+    int lv_msginCycle = 0;
+    int msginCycle = 0;
+    int msgid = 0;
+    std::string failure_msg_list[3] = {"success 1", "success 2", "failure 3"};
+    int failure_cond_list[3] = { 0, 0, 1};
+    bool prematureExit = true;
+    validinCycle = rand_next(100);
+    msginCycle = rand_next(100);
+    int errCode = rand_next(8);
+    lv_validinCycle = lv_validinCycle + validinCycle;
+    lv_msginCycle = lv_msginCycle + msginCycle;
+    poke(io_validInCycle, lv_validinCycle);
+    poke(io_msgInCycle, lv_msginCycle);
+    poke(io_doneErrCode, errCode);
+    step(1);
+    //poke(io_globalReset, 0);
+    poke(reset, 0);
+    if(errCode %2 !=0) {
+      msgid = 0;
+    } else if(errCode %4 != 0) {
+      msgid = 1;
+    } else {
+      msgid = 2;
+    }
+    step(lv_validinCycle+2, false);
+    while (!terminator->terminate()) {
+      terminator->tick();
+    }
+    int str_match = terminator->exit_message() == failure_msg_list[msgid];
+    expect(terminator->cycleCount()==(lv_validinCycle+2), "Code Exits at precise time");
+    expect(terminator->exit_code()==failure_cond_list[msgid], "Error Code Verified");
+    expect(str_match, "Error Message Verified");
+  };
+};
--- a/sim/src/main/cc/midasexamples/TerminationModule.h
+++ b/sim/src/main/cc/midasexamples/TerminationModule.h
@@ -21,30 +21,27 @@ public:
   };
   void run() {
     poke(reset, 1);
-    //poke(io_globalReset, 1);
-    int terminations = 0;
     int lv_validinCycle = 0;
     int validinCycle = 0;
     int lv_msginCycle = 0;
     int msginCycle = 0;
     int msgid = 0;
+    int reset_length = 1;
     std::string failure_msg_list[3] = {"success 1", "success 2", "failure 3"};
     int failure_cond_list[3] = { 0, 0, 1};
-    bool prematureExit = true;
     validinCycle = rand_next(100);
     msginCycle = rand_next(100);
-    int errCode = rand_next(8);
+    int termination_code = rand_next(8);
     lv_validinCycle = lv_validinCycle + validinCycle;
     lv_msginCycle = lv_msginCycle + msginCycle;
     poke(io_validInCycle, lv_validinCycle);
     poke(io_msgInCycle, lv_msginCycle);
-    poke(io_doneErrCode, errCode);
-    step(1);
-    //poke(io_globalReset, 0);
+    poke(io_doneErrCode, termination_code);
+    step(reset_length);
     poke(reset, 0);
-    if(errCode %2 !=0) {
+    if(termination_code %2 !=0) {
       msgid = 0;
-    } else if(errCode %4 != 0) {
+    } else if(termination_code %4 != 0) {
       msgid = 1;
     } else {
       msgid = 2;
@@ -54,8 +51,8 @@ public:
       terminator->tick();
     }
     int str_match = terminator->exit_message() == failure_msg_list[msgid];
-    expect(terminator->cycleCount()==(lv_validinCycle+2), "Code Exits at precise time");
-    expect(terminator->exit_code()==failure_cond_list[msgid], "Error Code Verified");
+    expect(terminator->cycle_count() == (lv_validinCycle+reset_length), "Code Exits at precise time");
+    expect(terminator->exit_code() == failure_cond_list[msgid], "Error Code Verified");
     expect(str_match, "Error Message Verified");
   };
 };
--- a/sim/src/main/scala/midasexamples/Termination.scala
+++ b/sim/src/main/scala/midasexamples/Termination.scala
@@ -2,48 +2,30 @@
 //See LICENSE for license details
 package firesim.midasexamples
 
-import midas.widgets._
-
 import chisel3._
 import chisel3.util._
-import firesim.bridges._
-import midas.targetutils._
 import freechips.rocketchip.config.Parameters
+
+import midas.widgets._
 
 class TerminationModuleIO(val params: TerminationBridgeParams) extends Bundle {
   val msgInCycle = Input(UInt(16.W))  //Cycle when error is valid
   val validInCycle = Input(UInt(16.W)) //Cycle when the program finishes
   val doneErrCode = Input(UInt(((params.conditionInfo).size).W)) //Error code supplied by program
-  //val globalReset = Input(Bool())
 }
 
 class TerminationModuleDUT extends Module {
 
   val endMessage = Seq(TerminationCondition(false, "success 1"), TerminationCondition(false, "success 2"), TerminationCondition(true, "failure 3"))
   val params = TerminationBridgeParams(endMessage)
-  
   val io = IO(new TerminationModuleIO(params))
   val counter = RegInit(0.U(16.W))
 
-  //val globalReset = WireInit(true.B)
-
-  //globalReset := io.globalReset
-
-  //GlobalResetCondition(globalReset)
-
   counter := counter + 1.U 
-
-
-  val valid = (io.doneErrCode.asBools).map { _ && (counter > io.validInCycle)}
-
-
-  // When the target wants to send processed error code.
-  //val errCode = Mux(counter > io.msgInCycle, io.doneErrCode, 0.U)
+  val valid = (io.doneErrCode.asBools).map { _ && (counter >= io.validInCycle)}
 
   TerminationBridge(valid, params) 
 
- // When the target wants to send processed error code.
- //TerminationBridge(clock, valid, errCode, params) 
 }
 
 class TerminationModule(implicit p: Parameters) extends PeekPokeMidasExampleHarness(() => new TerminationModuleDUT)
--- a/sim/src/main/scala/midasexamples/Termination.scala
+++ b/sim/src/main/scala/midasexamples/Termination.scala
@@ -1,0 +1,49 @@
+
+//See LICENSE for license details
+package firesim.midasexamples
+
+import midas.widgets._
+
+import chisel3._
+import chisel3.util._
+import firesim.bridges._
+import midas.targetutils._
+import freechips.rocketchip.config.Parameters
+
+class TerminationModuleIO(val params: TerminationBridgeParams) extends Bundle {
+  val msgInCycle = Input(UInt(16.W))  //Cycle when error is valid
+  val validInCycle = Input(UInt(16.W)) //Cycle when the program finishes
+  val doneErrCode = Input(UInt(((params.conditionInfo).size).W)) //Error code supplied by program
+  //val globalReset = Input(Bool())
+}
+
+class TerminationModuleDUT extends Module {
+
+  val endMessage = Seq(TerminationCondition(false, "success 1"), TerminationCondition(false, "success 2"), TerminationCondition(true, "failure 3"))
+  val params = TerminationBridgeParams(endMessage)
+  
+  val io = IO(new TerminationModuleIO(params))
+  val counter = RegInit(0.U(16.W))
+
+  //val globalReset = WireInit(true.B)
+
+  //globalReset := io.globalReset
+
+  //GlobalResetCondition(globalReset)
+
+  counter := counter + 1.U 
+
+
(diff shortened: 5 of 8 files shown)


davidbiancolin (commented):
Lets call this something simpler, like: `TerminationBridge`, so 's/TestFinisher/Termination/g' should do the trick.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import chisel3._
+import chisel3.util._
+import freechips.rocketchip.config.Parameters
+import midas.widgets._
+import midas.widgets.PeekPokeBridge
+
+case class CodeElaborator(isErr: Boolean, message : String)
+
+case class TestIndicatorParams(
```
davidbiancolin, line 12: So per our discussions: 

```
/**
   * Defines a condition under which the simulator should halt. 
   * @param Message A string printed by the driver to indicate why the simulator is terminating.
   * @param isError When true, instructs the driver to return a non-zero value when exiting under this condition.
   * /
case class TerminationCondition(
  message: String,
  isError, Boolean)
  
/** 
   * Serializable bridge parameters for the TerminationBridge.
   * 
   * @param terminationConditions Enumerates all of the termination conditions this bridge instance can process.
   */ 
case class TerminationBridgeParams(terminationConditions: Seq[TerminationCondition])
```

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+
+class TestFinisherPortIO(val params: TestIndicatorParams) extends Bundle {
+  //Error code in case program terminates unsuccessfully
+  val errCode = UInt((params.codeBits).W)
+  val valid = Bool() 
+}
+
+class TestFinisherBridgeTargetIO(params: TestIndicatorParams) extends Bundle {
+  val clock = Input(Clock())
+  val testDone = Input(new TestFinisherPortIO(params))
```
davidbiancolin, line 28: Flatten as discussed.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+                              (private val targetIO : TestFinisherBridgeTargetIO = new TestFinisherBridgeTargetIO(params)) 
+                               extends Bundle with ChannelizedHostPortIO{
+  def targetClockRef = targetIO.clock 
+  //It is just inputs from the target to indicate test completion
+  //The HostPort annotation would not work with aggregates so creating individual channels.
+  val errCodeH = InputChannel(targetIO.testDone.errCode)
+  val validH = InputChannel(targetIO.testDone.valid)
+}
+
+class TestFinisherBridge(params: TestIndicatorParams)(implicit p: Parameters) extends BlackBox
```
davidbiancolin, line 41: Drop the implicit p: Parameters here. I don't think it's used.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+  val io = IO(new TestFinisherBridgeTargetIO(params))
+  val bridgeIO = new TestFinisherBridgeHostIO(params)(io)
+
+  val constructorArg = Some(params)
+
+  generateAnnotations()
+}
+
+object TestFinisherBridge {
+  def apply(clock: Clock, finisher: TestFinisherPortIO)(implicit p: Parameters): TestFinisherBridge = {
```
davidbiancolin, line 52: Ditto: remove the implicit.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+    val tfm = Module(new TestFinisherBridge(finisher.params))
+    tfm.io.testDone := finisher
+    tfm.io.clock := clock 
+    tfm  
+  }
```
davidbiancolin, line 57: Consider adding an apply method that takes a Seq[Bool] and a Seq[TerminationCondition] and does the encoding using a priority encoder.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+    // AXI4-lite for the simulation control bus, =
+    // AXI4 for DMA
+    val io = IO(new WidgetIO())
+
+
+
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TestFinisherBridgeHostIO(params)())
+    
+    val statusDone = RegInit(Bool(), false.B)
```
davidbiancolin, line 72: You should be able to write this as: 
```suggestion
    val statusDone = RegInit(false.B)
```

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+import midas.widgets._
+import midas.widgets.PeekPokeBridge
+
+case class CodeElaborator(isErr: Boolean, message : String)
+
+case class TestIndicatorParams(
+  nConcurrentTests: Int     = 1,
+  codeMessage: Seq[CodeElaborator]
+) {
+  def codeBits: Int    = 16
```
davidbiancolin, line 16: If you want to keep this around: 
```suggestion
  def codeBits: Int    = log2ceil(codeMessage.length)
```

sim/midas/src/main/cc/bridges/test_finisher.cc
```diff
@@ -0,0 +1,37 @@
...
+
+void test_finisher_t::tick() {  //reads the MMIOs every tick
+  if (read(this->mmio_addrs->out_status)) {
+	  msg_id = read(this->mmio_addrs->out_errCode);
+		if(this->isErr[msg_id]==0) {
+			fprintf(stderr, "*PASS* \n");
+		} else {
+			fprintf(stderr, "*FAIL* \n");
+		}
+	  fprintf(stderr, "Test completed with message:  ", msg_id);
```
davidbiancolin, line 32: Use (favor) the c++ std iostreams consistently.

sim/midas/src/main/cc/bridges/test_finisher.h
```diff
@@ -0,0 +1,32 @@
...
+				TESTFINISHERBRIDGEMODULE_struct * mmio_addrs,
+        unsigned int num_messages,
+        unsigned int* isErr,
+        const char* const* msgs);
+				~test_finisher_t();
+        virtual void init() {};
+        virtual void tick();
+        virtual void finish() {};
+        virtual bool terminate() { return test_done; };
+        virtual int exit_code() { return test_done ? 1 : 0; };
```
davidbiancolin, line 23: This needs to be driver by `isErr` in some way. You might as well set in tick as you've done with test_done.

sim/midas/src/main/cc/bridges/test_finisher.cc
```diff
@@ -0,0 +1,37 @@
...
+	msgs(msgs)
+{
+  this->mmio_addrs = mmio_addrs;
+}
+
+test_finisher_t::~test_finisher_t() {
+	free(this->mmio_addrs);
+}
+
+void test_finisher_t::tick() {  //reads the MMIOs every tick
```
davidbiancolin, line 24: Consider not polling on every tick invocation. This should eventually be on a +arg, but for now start by having some constant value called out in the header. Polling every 10 iterations will probably help improve FMR without meaningfully extending metasimulation runtimes.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+
+
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TestFinisherBridgeHostIO(params)())
+    
+    val statusDone = RegInit(Bool(), false.B)
+    
+    statusDone := hPort.validH.bits
+    val errCode = RegInit(0.U((params.codeBits).W))
+    when(statusDone) {
```
davidbiancolin, line 76: Does this not return the code from the cycle after hPort.validH.bits is first asserted?

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
```diff
@@ -0,0 +1,97 @@
...
+      errCode := hPort.errCodeH.bits
+    }
+    //Condition needed for to signal HostPort Channel can send the messages.
+    //So host is ready until the status is Done.
+    hPort.errCodeH.ready := !statusDone
```
davidbiancolin, line 81: So i made this harder for you than it needed to be by asking you to use ChannelizedHostPortIO (since i forgot it's not possible to mark aggregates as input channels). However, you should dequeue code and valid bits carefully to handle potential decoupling between them (e.g., you could do it concurrently). You can look at other bridges for examples of this, but the simplest way to do that is as follows: 

```
val helper = DecoupledHelper(hPort.errCodeH.valid, hPort.validH.valid,  hPort.validH.bits)
hPort.errCodeH.ready := helper.fire(hPort.errorCodeH.valid)
hPort.validH.ready := helper.fire(hPort.validH.valid)
```

If you want to be more fastidious you can latch the errorcode and valid first (making sure to handle potential decoupling between the channels) , and then use those to drive each channel's ready, instead of combinationally coupling the valid from opposite channels back to ready. This is closer to what you already have.

davidbiancolin (commented):
Cool to see this shaping up. We'll talk about retargeting this to main in a meeting.

Context for other devs: Rahul is writing a simple bridge that can be deployed in a target-independent manner that can call for simulation termination.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: Done.

jerryz123:
It would be cool if this hung off the `success` signal in the FireSim Chipyard "harness". This would:
- unify the termination mechanism between SW sim and FireSim. 
- implicitly test this block within the RocketChip regression tests
- extend support to other chipyard models which indicate completion through the harness success signal

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: [https://github.com/firesim/firesim/blob/2126f267f0b61b05d86d349675674ef391a6e572/sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala#L70-L91](url)
I decoupled it by latching.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: Assuming the `Seq[Bool]` as Seq of completion and priority encoder is applied on isErr of Seq[TerminationCondition]. Addressed it.

sim/firesim-lib/src/main/scala/bridges/TestFinisherBridge.scala
rahulb10: Done.

sim/midas/src/main/cc/bridges/test_finisher.cc
rahulb10: Yup.

rahulb10:
> Lets call this something simpler, like: `TerminationBridge`, so 's/TestFinisher/Termination/g' should do the trick.

I replaced the name in TestFinisherBridge.scala. Do we want to replace the name in other files and call it Terminator?

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,112 @@
...
+    with Bridge[TerminationBridgeHostIO, TerminationBridgeModule] {
+  val io = IO(new TerminationBridgeTargetIO(params))
+  val bridgeIO = new TerminationBridgeHostIO(params)(io)
+
+  val constructorArg = Some(params)
+
+  generateAnnotations()
+}
+
+object TerminationBridgeWithCode {
```
davidbiancolin, line 49: This should be in the same companion object `TerminationBridge`.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,112 @@
...
+  def apply(clock: Clock, finish: Bool, errCode: UInt, params: TerminationBridgeParams): TerminationBridge = {
+    val tbm = Module(new TerminationBridge(params))
```
davidbiancolin, line 51: Expand this name so it is more descriptive in the verilog.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,112 @@
...
+    tbm.io.valid := finish
+    tbm.io.clock := clock 
+    tbm  
+  }
+}
+
+object TerminationBridge {
+  def apply(clock: Clock, finish: Seq[Bool], errCode: Seq[TerminationCondition], params: TerminationBridgeParams): TerminationBridge = {
+    val finishAll = finish.reduce(_||_)
+    val errMessageID = PriorityEncoder((errCode.map(x => (x.isErr).B)))
```
davidbiancolin, line 62:
```suggestion
    val errMessageID = PriorityEncoder(finish)
```

This needs to find the first termination condition that is true, regardless of whether it is an error or not.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,112 @@
...
+
+
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TerminationBridgeHostIO(params)())
+    
+    //Resume the simulation after the target called for termination
+    val resume = WireInit(false.B)
+
+    val statusDone = Module(new Queue(Bool(), 1))
+    val errCode = Module(new Queue(UInt(log2Ceil((params.codeMessage).size).W), 1))
```
davidbiancolin, line 85: A DRY-er way to do this: 
```suggestion
    val errCode = Module(new Queue(hPort.errCodeH.bits.cloneType), 1))
```

Also there is a "Queue" companion object that more simply does this that you should have a look at: 
https://www.chisel-lang.org/api/latest/chisel3/util/Queue$.html

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,64 @@
...
+	unsigned int num_messages,
+  unsigned int* isErr,
+	const char* const* msgs): 
+	bridge_driver_t(sim), 
+	mmio_addrs(mmio_addrs), 
+	isErr(isErr),
+	msgs(msgs)
+{
+  this->mmio_addrs = mmio_addrs;
+  std::string tick_rate_arg = std::string("+tick-rate=");
```
davidbiancolin, line 19: Add comments explain the function of these plus args.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,64 @@
+#ifdef TERMINATIONBRIDGEMODULE_struct_guard
+
```
davidbiancolin, line 2: The indentation in this file looks a little weird.  Expand tabs and use 2 spaces consistently.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,64 @@
...
+		if (read(this->mmio_addrs->out_status)) {
+		  msg_id = read(this->mmio_addrs->out_errCode);
+			if(this->isErr[msg_id]==0) {
+			  fail = 0;
+				std::cerr << "*PASS*" << std::endl;
+			} else {
+			  fail = 1;
+				std::cerr << "*FAIL*" << std::endl;
+			}
+			std::cerr << "Test completed with message";
```
davidbiancolin, line 50: I think this would be better as: termination condition set on Cycle (return cycle). 

You'll need to add a cycle counter in your bridge, and then look at printf / assertion bridges to see how to get clock domain information.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,52 @@
...
+        std::vector<std::string> args(argv + 1, argv + argc);
+        terminator = new termination_t(
+          this, args,
+          TERMINATIONBRIDGEMODULE_0_substruct,
+          TERMINATIONBRIDGEMODULE_0_message_count,
+          TERMINATIONBRIDGEMODULE_0_message_type,
+          TERMINATIONBRIDGEMODULE_0_message);
+    };
+    void run() {
+        poke(reset, 1);
```
davidbiancolin, line 24: Again the indentation here is wonky. Likely need to expand tabs.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,52 @@
...
+				srand(time(NULL));
```
davidbiancolin, line 25: simif_peek_poke_t provides some barebones randomization utilities, use those so that seeding is done consistently and reproducibly.

davidbiancolin (commented):
At a high level what the fuzzer needs to do is check that when a specific error code is driven to the target bridge, the right message and return code is produced.

If you want to do this in the C++ driver, in you'll need to duplicate the strings and expected error codes. 

Alternatively, you could do this from ScalaTest, where you pass as command line arguments to the driver a randomly selected error code. Then post-process the output of the simulation to see the right string was produced.

sim/src/main/cc/midasexamples/TerminationModule.h
rahulb10: Got it. Done.

sim/midas/src/main/cc/bridges/termination.cc
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+    //val gReset = WireInit(false.B)
+    //GlobalResetCondition.sink(gReset)
+    val terminationBridgeTarget = Module(new TerminationBridge(params))
+    terminationBridgeTarget.io.errCode := errMessageID
+    terminationBridgeTarget.io.valid := finishAll && !lReset.asBool
+    terminationBridgeTarget.io.clock := clock 
+    terminationBridgeTarget  
+  }
+
+  def apply(clock: Clock, lReset: Reset, finish: Seq[Bool], errCode: Seq[TerminationCondition], params: TerminationBridgeParams): TerminationBridge = {
```
davidbiancolin, line 63: Just some naming suggestions on this interface: 
Maybe lReset -> reset, or expand it to "localReset". 

finish -> conditionBools
errCode -> conditionInfo

Also add `ScalaDoc` for both these apply methods, since i htink they are pretty close to being ready to ship.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+import freechips.rocketchip.util.{DecoupledHelper}
+
+/**
+  * Defines a condition under which the simulator should halt. 
+  * @param Message A string printed by the driver to indicate why the simulator is terminating.
+  * @param isError When true, instructs the driver to return a non-zero value when exiting under this condition.
+**/
+case class TerminationCondition(isErr: Boolean, message : String)
+
+//A seq of termination conditions one for each concurrently running program
```
davidbiancolin, line 18: Make this `ScalaDoc` as above. See: https://docs.scala-lang.org/style/scaladoc.html

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TerminationBridgeHostIO(params)())
+    
+    //Resume the simulation after the target called for termination
+    val resume = WireInit(false.B)
+
+    val statusDone = Module(new Queue(Bool(), 1))
+    val errCode = Module(new Queue(hPort.errCodeH.bits.cloneType, 1))
+
+    val cycleCountWidth = 16 //Data width is capped by nastikey
```
davidbiancolin, line 101: You should use a 64b counter for this. (The width is actually capped at 32b). 

Here's an example of how that looks: https://github.com/firesim/firesim/blob/366ceabe13ce27056ec1c53bf93d017ec182d85e/sim/midas/src/main/scala/midas/widgets/ClockBridge.scala#L172

davidbiancolin: And here's the C side: https://github.com/firesim/firesim/blob/366ceabe13ce27056ec1c53bf93d017ec182d85e/sim/midas/src/main/cc/simif.cc#L56

Notice the write to "latch" which provides a hacky way to do an "atomic" 64b read using two 32b reads.

davidbiancolin: All three of those addresses (_h, _l, _latch) will be autogenerated by the scala above.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+
+    val cycleCountWidth = 16 //Data width is capped by nastikey 
+
+    val tokenCounter = RegInit(0.U(cycleCountWidth.W))
+
+
+    statusDone.io.enq <> hPort.validH
+    errCode.io.enq <> hPort.errCodeH 
+
+    val terminateResume = !statusDone.io.deq.bits || resume
```
davidbiancolin, line 109: We decided to strip out the resume right?

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+  val bridgeIO = new TerminationBridgeHostIO(params)(io)
+
+  val constructorArg = Some(params)
+
+  generateAnnotations()
+}
+
+object TerminationBridge {
+
+  private def annotateTerminationBridge(clock: Clock, lReset: Reset, finish: Seq[Bool], errCode: Seq[TerminationCondition], params: TerminationBridgeParams): TerminationBridge = {
```
davidbiancolin, line 51: Remove errCode as it is unused.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+    terminationBridgeTarget.io.valid := finishAll && !lReset.asBool
+    terminationBridgeTarget.io.clock := clock 
+    terminationBridgeTarget  
+  }
+
+  def apply(clock: Clock, lReset: Reset, finish: Seq[Bool], errCode: Seq[TerminationCondition], params: TerminationBridgeParams): TerminationBridge = {
+    annotateTerminationBridge(clock, lReset, finish, errCode, params)
+  }
+
+  def apply(clock: Clock, lReset: Reset, finish: Bool, errCode: UInt, params: TerminationBridgeParams): TerminationBridge = {
```
davidbiancolin, line 67: I think for now (and for simplicity) we decided to just remove this method.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+  generateAnnotations()
+}
+
+object TerminationBridge {
+
+  private def annotateTerminationBridge(clock: Clock, lReset: Reset, finish: Seq[Bool], errCode: Seq[TerminationCondition], params: TerminationBridgeParams): TerminationBridge = {
+    val finishAll = finish.reduce(_||_)
+    val errMessageID = PriorityEncoder(finish)
+    //val gReset = WireInit(false.B)
+    //GlobalResetCondition.sink(gReset)
```
davidbiancolin, line 55: Add a require here that ensures that params.codeMessage.size == finish.size

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,136 @@
...
+/**
+  * Defines a condition under which the simulator should halt. 
+  * @param Message A string printed by the driver to indicate why the simulator is terminating.
+  * @param isError When true, instructs the driver to return a non-zero value when exiting under this condition.
+**/
+case class TerminationCondition(isErr: Boolean, message : String)
+
+//A seq of termination conditions one for each concurrently running program
+case class TerminationBridgeParams(
+  codeMessage: Seq[TerminationCondition]
```
davidbiancolin, line 20: Maybe `conditionInfo` is more descriptive.

davidbiancolin (commented):
Scala-side is getting close. 

The big thing is that `errCode` and `params` are redundant on many of the apply methods. I think maybe it's best to just remove errCode and use the full params everywhere.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,67 @@
...
+
+termination_t::~termination_t() {
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs every tick
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_errCode);
+    	fail = msg_id;
```
davidbiancolin, line 38: This is incorrect. It needs to be:

```suggestion
    	this->fail = isErr[msg_id] != 0; 
```

Your test should catch this.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,67 @@
...
+    	if(this->isErr[msg_id]==0) {
+    		std::cerr << "*PASS*" << std::endl;
+    	} else {
+    		std::cerr << "*FAIL*" << std::endl;
+    	}
+    	std::cerr << "Test completed with message ";
```
davidbiancolin, line 44:
```suggestion
    	std::cerr << "TerminationBridge detected exit condition on cycle " << this->cycleCount << ":" << std::endl;
```
Then print the message.

sim/midas/src/main/cc/bridges/termination.h
```diff
@@ -0,0 +1,38 @@
...
+    const char* const* msgs);
+    ~termination_t();
+    virtual void init() {};
+    virtual void tick();
+    virtual void finish() {};
+    virtual bool terminate() { return test_done; };
+    virtual int exit_code() { return fail; };
+    const char* exit_message();
+    int cycleCount();
+    void resume();
```
davidbiancolin, line 27: Remove.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      msgid = 2;
+    }
+    step(lv_validinCycle+2, false);
+    while (!terminator->terminate()) {
+      terminator->tick();
+    }
+    int str_match = strcmp(terminator->exit_message(),failure_list[msgid]);
+    expect(terminator->cycleCount()==(lv_validinCycle+2), "Code Exits at precise time");
+    expect(terminator->exit_code()==(msgid), "Error Code Verified");
```
davidbiancolin, line 60: The exit code should be either 0 or 1, i.e, it should be the result of `isErr[msg_id]`

If one day we want to better capture _why_ something failed we'll have to log it to a file with some more info about the bridge.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  //val globalReset = WireInit(true.B)
+
+  //globalReset := io.globalReset
+
+  //GlobalResetCondition(globalReset)
+
+  counter := counter + 1.U 
+
+  val vecValids = Seq(true.B, true.B, true.B)
```
davidbiancolin, line 36: Remove this.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  val valid = (vecValids zip io.doneErrCode.asBools).map { case (x,y) => x && (counter > io.validInCycle) && y}
+
+  val finishValids = valid.reduce(_||_)
```
davidbiancolin, line 40: And this.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,52 @@
...
+
+
+  // When the target wants to send processed error code.
+  //val errCode = Mux(counter > io.msgInCycle, io.doneErrCode, 0.U)
+
+  TerminationBridge(clock, reset, valid, endMessage, params)
```
davidbiancolin, line 46: You might as well use the implicit clock + reset form here.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,63 @@
...
+      msgid = 1;
+    } else {
+      msgid = 2;
+    }
+    step(lv_validinCycle+2, false);
+    while (!terminator->terminate()) {
+      terminator->tick();
+    }
+    int str_match = strcmp(terminator->exit_message(),failure_list[msgid]);
+    expect(terminator->cycleCount()==(lv_validinCycle+2), "Code Exits at precise time");
```
davidbiancolin, line 59: I think this should just +1. Basically the reset length + lv_validinCycle. Can you explain why it's two? 

I'd set the reset length somewhere in the C++, and use it to encode the the expected length.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,63 @@
...
+  void run() {
+    poke(reset, 1);
+    //poke(io_globalReset, 1);
+    int terminations = 0;
+    int lv_validinCycle = 0;
+    int validinCycle = 0;
+    int lv_msginCycle = 0;
+    int msginCycle = 0;
+    int msgid = 0;
+    char failure_list[3][10] = {"failure 1",
```
davidbiancolin, line 31: use std::string

davidbiancolin (commented):
More comments.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Yup. Will keep in mind the scaladoc style.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: I thought it will be useful to the user. Or is it only for testing purpose?

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Yup.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: We don't the use the `globalresetcondition` any more. I thought so not yet complex. Nonetheless we already have a apply method that gives full power to the users

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Okay.

sim/midas/src/main/cc/bridges/termination.h
rahulb10: There is also latch in the bridge module right?

sim/src/main/cc/midasexamples/TerminationModule.h
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done

sim/midas/src/main/cc/bridges/termination.cc
rahulb10: Yup. Redundant.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Done.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Same dilemma as mentioned above.

sim/midas/src/main/cc/bridges/termination.cc
rahulb10: Done

sim/src/main/scala/midasexamples/Termination.scala
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
rahulb10: Didn't extensively test it. So removed.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
davidbiancolin: I meant remove the _whole_ apply method, so there's one less thing you have to test.

sim/src/main/cc/midasexamples/TerminationModule.h
rahulb10: There is a latch right? In the TerminationBridgeModule.

davidbiancolin (commented):
Good progress.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,49 @@
...
+  //val globalReset = WireInit(true.B)
+
+  //globalReset := io.globalReset
+
+  //GlobalResetCondition(globalReset)
+
+  counter := counter + 1.U 
+
+
+  val valid = (io.doneErrCode.asBools).map { case x => x && (counter > io.validInCycle)}
```
davidbiancolin, line 37:
```suggestion
  val valid = (io.doneErrCode.asBools).map { _ && (counter > io.validInCycle)}
```
This is an easy place where you can do this anonymously.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,61 @@
...
+
+termination_t::~termination_t() {
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs every tick
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_errCode);
+    	this->fail = isErr[msg_id];
```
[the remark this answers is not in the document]
rahulb10: Definitely. My bad

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    val cycleCountWidth = 64 //Data width is capped by nastikey 
+
+    val tokenCounter = genWideRORegInit(0.U(cycleCountWidth.W), "out_counter")    
+
+
+    statusDone.io.enq <> hPort.validH
+    errCode.io.enq <> hPort.errCodeH 
+
+    val terminateResume = !statusDone.io.deq.bits
```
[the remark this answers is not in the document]
rahulb10: Done.

sim/src/main/scala/midasexamples/Termination.scala
rahulb10: Makes sense

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,61 @@
...
+}
+
+termination_t::~termination_t() {
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs every tick
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_errCode);
```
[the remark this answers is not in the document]
rahulb10: Done

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+    // AXI4-lite for the simulation control bus, =
+    // AXI4 for DMA
+    val io = IO(new WidgetIO())
+
+
+
+    // This creates the host-side interface of your TargetIO
+    val hPort = IO(new TerminationBridgeHostIO(params)())
+    
+    val statusDone = Module(new Queue(Bool(), 1))
```
davidbiancolin, line 98: Since you're just enqueuing the tokens directly into these, you should use the `Queue.apply` methods, and use two entry queues. 
```suggestion
    val statusDone = Queue(hPort.validH)
```

Then replace all instances of `statusDone.io.deq` with just statusDone.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+
+
+    statusDone.io.enq <> hPort.validH
+    errCode.io.enq <> hPort.errCodeH 
+
+    val noTermination = !statusDone.io.deq.bits
+
+    val tFireHelper = DecoupledHelper(errCode.io.deq.valid, statusDone.io.deq.valid, noTermination)
+
+    when(tFireHelper.fire()) {
```
davidbiancolin, line 113:
```suggestion
    when(tFireHelper.fire) {
```
Avoid parens on nullary functions. This for Scala 2.13 compatibility.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+      tokenCounter := tokenCounter + 1.U
+    }
+
+    statusDone.io.deq.ready := tFireHelper.fire(statusDone.io.deq.valid)
+    errCode.io.deq.ready := tFireHelper.fire(errCode.io.deq.valid)
+
+    genROReg(statusDone.io.deq.bits && statusDone.io.deq.valid, "out_status")
```
davidbiancolin, line 120:
```suggestion
    // When set, indicates the bridge has detected a termination condition
    genROReg(statusDone.io.deq.bits && statusDone.io.deq.valid, "out_status")
```
I think it's a good practice to comment the MMIO registers here to explain what they do.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+ */
+  def apply(conditionBools: Seq[Bool], params: TerminationBridgeParams): TerminationBridge = {
+    annotateTerminationBridge(Module.clock, Module.reset, conditionBools, params)
+  }
+
+}
+
+class TerminationBridgeModule(params: TerminationBridgeParams)(implicit p: Parameters) extends BridgeModule[TerminationBridgeHostIO]()(p) {
+  lazy val module = new BridgeModuleImp(this) {
+    // This creates the interfaces for all of the host-side transport
```
davidbiancolin, line 88: This comment is no longer accurate. I'd just say it instantiates the control bus.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  generateAnnotations()
+}
+
+object TerminationBridge {
+
+  private def annotateTerminationBridge(clock: Clock, reset: Reset, conditionBools: Seq[Bool], params: TerminationBridgeParams): TerminationBridge = {
+    val finish = conditionBools.reduce(_||_)
+    val errMessageID = PriorityEncoder(conditionBools)
+    //val gReset = WireInit(false.B)
```
davidbiancolin, line 55: Lets remove all commented out code for merge. You'll add this back in later.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
```diff
@@ -0,0 +1,133 @@
...
+}
+
+class TerminationBridgeModule(params: TerminationBridgeParams)(implicit p: Parameters) extends BridgeModule[TerminationBridgeHostIO]()(p) {
+  lazy val module = new BridgeModuleImp(this) {
+    // This creates the interfaces for all of the host-side transport
+    // AXI4-lite for the simulation control bus, =
+    // AXI4 for DMA
+    val io = IO(new WidgetIO())
+
+
```
davidbiancolin, line 93:
```suggestion
```

davidbiancolin: Nit: Slightly excessive whitespace.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,66 @@
...
+  unsigned int* isErr,
+  const char* const* msgs): 
+  bridge_driver_t(sim), 
+  mmio_addrs(mmio_addrs), 
+  num_messages(num_messages), 
+  isErr(isErr),
+  msgs(msgs)
+{
+  //tick-rate to decide sampling rate of MMIOs per number of ticks
+  this->mmio_addrs = mmio_addrs;
```
davidbiancolin, line 20:
```suggestion
```
You've already set mmio_addrs in the class's initializer list.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,66 @@
...
+
+termination_t::~termination_t() {
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs tick-rate
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_errCode);
+    	this->fail = this->isErr[msg_id];
```
davidbiancolin, line 39: The indentation in this file appears to be inconsistent still, judging by GH.com's rendering. Please fix this by replacing stray tabs before merge.

sim/midas/src/main/cc/bridges/termination.h
```diff
@@ -0,0 +1,38 @@
...
+    unsigned int* isErr,
+    const char* const* msgs);
+    ~termination_t();
+    virtual void init() {};
+    virtual void tick();
+    virtual void finish() {};
+    virtual bool terminate() { return test_done; };
+    virtual int exit_code() { return fail; };
+    const char* exit_message();
+    int cycleCount();
```
davidbiancolin, line 26:
```suggestion
    int cycle_count();
```

Snake case on class methods.

sim/midas/src/main/cc/bridges/termination.h
```diff
@@ -0,0 +1,38 @@
...
+  private:
+    TERMINATIONBRIDGEMODULE_struct * mmio_addrs;
+    bool test_done = false;
+    int fail = 0;
+    int tick_rate = 10;
+    int tick_counter = 0;
+    unsigned int num_messages;
+    unsigned int* isErr;
```
davidbiancolin, line 34:
```suggestion
    unsigned int* is_err;
```

And methods.

sim/src/main/cc/midasexamples/TerminationModule.h
davidbiancolin: If by latch you mean, the cycle required to increment the cycle counter, then yeah. I think i'd change the comparison to be `>=` instead of `>`, then this increment would be exactly equal to the reset length. 

Then you should use the reset length to encode this expect instead of using a magic number (the literal 2).

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      msgid = 2;
+    }
+    step(lv_validinCycle+2, false);
+    while (!terminator->terminate()) {
+      terminator->tick();
+    }
+    int str_match = terminator->exit_message() == failure_msg_list[msgid];
+    expect(terminator->cycleCount()==(lv_validinCycle+2), "Code Exits at precise time");
+    expect(terminator->exit_code()==failure_cond_list[msgid], "Error Code Verified");
```
davidbiancolin, line 58:
```suggestion
    expect(terminator->exit_code() == failure_cond_list[msgid], "Error Code Verified");
```
Spaces around these boolean operators.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,49 @@
...
+
+class TerminationModuleDUT extends Module {
+
+  val endMessage = Seq(TerminationCondition(false, "success 1"), TerminationCondition(false, "success 2"), TerminationCondition(true, "failure 3"))
+  val params = TerminationBridgeParams(endMessage)
+  
+  val io = IO(new TerminationModuleIO(params))
+  val counter = RegInit(0.U(16.W))
+
+  //val globalReset = WireInit(true.B)
```
davidbiancolin, line 28: Remove all the commented code for now.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  //GlobalResetCondition(globalReset)
+
+  counter := counter + 1.U 
+
+
+  val valid = (io.doneErrCode.asBools).map { _ && (counter > io.validInCycle)}
+
+
+  // When the target wants to send processed error code.
```
davidbiancolin, line 40: Remove these examples unless they are actually used during elaboration of the module.

sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -368,6 +368,11 @@ class ResetPulseBridgeActiveLowTest extends TutorialSuite(
     shouldPass = false)
 }
 
+class TerminationF1Test extends TutorialSuite("TerminationModule") {
+  1 to 10 foreach {x => runTest(backendSimulator, args = Seq("+tick-rate=10", s"+seed=${x}"), shouldPass = true)}
+  //runTest(backendSimulator, args = Seq("+tick-rate=10", s"+seed=10"), shouldPass = true)
```
davidbiancolin, line 373: Again, remove commented code.

sim/src/main/scala/midasexamples/Termination.scala
```diff
@@ -0,0 +1,49 @@
...
+  //val globalReset = WireInit(true.B)
+
+  //globalReset := io.globalReset
+
+  //GlobalResetCondition(globalReset)
+
+  counter := counter + 1.U 
+
+
+  val valid = (io.doneErrCode.asBools).map { _ && (counter > io.validInCycle)}
```
davidbiancolin, line 37:
```suggestion
  val valid = (io.doneErrCode.asBools).map { _ && (counter >= io.validInCycle)}
```

To simplify the expect logic so that the expected cycle is exactly equal to the reset length.

sim/src/main/cc/midasexamples/TerminationModule.h
```diff
@@ -0,0 +1,61 @@
...
+    //poke(io_globalReset, 1);
+    int terminations = 0;
+    int lv_validinCycle = 0;
+    int validinCycle = 0;
+    int lv_msginCycle = 0;
+    int msginCycle = 0;
+    int msgid = 0;
+    std::string failure_msg_list[3] = {"success 1", "success 2", "failure 3"};
+    int failure_cond_list[3] = { 0, 0, 1};
+    bool prematureExit = true;
```
davidbiancolin, line 33: This appears unused? Can you delete all dead or unused code to prepare for merge.

davidbiancolin (commented):
I'm almost ready to approve this. 

Please move code from `firesim-lib` to `midas` with the other bridges.
I'd like to see errCode renamed to terminationCode, because not all termination conditions are errors.

sim/firesim-lib/src/main/scala/bridges/TerminationBridge.scala
davidbiancolin: You can just delete it if you're unsure.

davidbiancolin:
Oh also, please update the PR title to reflect the name of the actual bridge -- this will go in the changelog.

Something like "Add a generic TerminationBridge to detect target-defined exit conditions"
Or something like that.

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,65 @@
...
+  free(this->mmio_addrs);
+}
+
+void termination_t::tick() {  //reads the MMIOs at tick-rate
+  if(tick_counter == tick_rate) {
+    if (read(this->mmio_addrs->out_status)) {
+      int msg_id = read(this->mmio_addrs->out_terminationCode);
+      this->fail = this->is_err[msg_id];
+      test_done = true;
+      if(msg_id < this->num_messages) {
```
davidbiancolin, line 40: Thanks for fixing the indentation. It's easier to review this. 

I think you should either make this comparison an assert, or change line 44 to say msg_id / error code was out of bounds, and call exit().

sim/midas/src/main/cc/bridges/termination.cc
```diff
@@ -0,0 +1,65 @@
...
+  unsigned int* is_err,
+  const char* const* msgs): 
+  bridge_driver_t(sim), 
+  mmio_addrs(mmio_addrs), 
+  num_messages(num_messages), 
+  is_err(is_err),
+  msgs(msgs)
+{
+  //tick-rate to decide sampling rate of MMIOs per number of ticks
+  std::string tick_rate_arg = std::string("+tick-rate=");
```
davidbiancolin, line 20: Maybe make this more verbose: `+termination-bridge-tick-rate`

sim/midas/src/main/cc/bridges/termination.cc
davidbiancolin: And you need to move the assert / bounds check above the index on line 38.

davidbiancolin (approved):
I think this can be merged once my final comments are addressed. 

There are some minor style things like consistently favoring `if (` over `if(` but i think we should just get this in.

PR #1033, opened 2022-04-20
CI:  Switch manager to use spot instances, cull instances more frequently 

davidbiancolin:
This is highest ROI change we can make to save EC2 credits without fundamentally changing the flow. I'm looking at intercepting a spot interruption notification from the workflow -- we could use that to potentially restart CI with on-demand instances. In the meantime devs should remain vigilant for potential spot-interruption-based failures.

#### UI / API Impact

None.

N/A

### Contributor Checklist

### DIFF
--- a/.github/scripts/launch-manager-instance.py
+++ b/.github/scripts/launch-manager-instance.py
@@ -24,6 +24,8 @@ def main():
     awstools.main([
         'launch',
         '--inst_type', 'z1d.2xlarge',
+        '--market', 'spot',
+        '--int_behavior', 'terminate',
         '--block_devices', str([{'DeviceName':'/dev/sda1','Ebs':{'VolumeSize':300,'VolumeType':'gp2'}}]),
         '--tags', str(get_manager_tag_dict(ci_commit_sha1, ci_workflow_run_id)),
         '--user_data_file', ci_workdir + "/scripts/machine-launch-script.sh"
--- a/.github/scripts/cull-old-ci-instances.py
+++ b/.github/scripts/cull-old-ci-instances.py
@@ -16,7 +16,7 @@ sys.path.append(ci_workdir + "/deploy/awstools")
 from awstools import get_instances_with_filter
 
 # The number of hours an instance may exist since its initial launch time
-INSTANCE_LIFETIME_LIMIT_HOURS = 24
+INSTANCE_LIFETIME_LIMIT_HOURS = 8
 
 def main():
     # Get a timezone-aware datetime instance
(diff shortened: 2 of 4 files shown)


.github/scripts/launch-manager-instance.py
```diff
@@ -24,6 +24,8 @@ def main():
     awstools.main([
         'launch',
         '--inst_type', 'z1d.2xlarge',
+        '--market', 'spot',
+        '--int_behavior', 'terminate',
```
davidbiancolin, line 28: This is the default, i just wanted to make it explicit.

abejgonzalez:
There is currently logic to have instances "stop" when there is a failure (I think this is mainly contained in the daemon/monitor). Are you planning on changing that as well?

davidbiancolin:
> There is currently logic to have instances "stop" when there is a failure (I think this is mainly contained in the daemon/monitor). Are you planning on changing that as well?

I wasn't planning to.

abejgonzalez:
> > There is currently logic to have instances "stop" when there is a failure (I think this is mainly contained in the daemon/monitor). Are you planning on changing that as well?
> 
> I wasn't planning to, was there a particular benefit you saw in doing so?

I think the logic should be changed to just terminate the instance on failure - whether or not you are using spot or ondemand instances (I thought in the past you added the stopping behavior so that you could debug things). I think its reasonable to just terminate because very few people have access to the CI AWS machines currently

davidbiancolin:
> > > There is currently logic to have instances "stop" when there is a failure (I think this is mainly contained in the daemon/monitor). Are you planning on changing that as well?
> > 
> > 
> > I wasn't planning to, was there a particular benefit you saw in doing so?
> 
> I think the logic should be changed to just terminate the instance on failure - whether or not you are using spot or ondemand instances (I thought in the past you added the stopping behavior so that you could debug things). I think its reasonable to just terminate because very few people have access to the CI AWS machines currently

Yeah, i wanted it to enable potentially restarting the instance to debug. You make a good point though. The only cost associated with leaving them stopped is the cost of the EBS volume before they are culled right? Your point is it should be consistent with the spot interruption behavior?

abejgonzalez:
> The only cost associated with leaving them stopped is the cost of the EBS volume before they are culled right? 

I think so.

> Your point is it should be consistent with the spot interruption behavior?

I'm advocating for never having the CI restart on a failure. In the case of a spot interrupt, we should terminate but make sure it is clear that the CI failed due to an interrupt so that the user can re-run the CI flow. In the case of the CI generally failing (presumably due to a user PR error), then the instance should terminate. At that point, the PR developer should fix the issue in their own AWS setup.

davidbiancolin:
The only reason i'm hesitant is that it's been useful for me in the past and it doesn't really cost anything to leave as is, but I'll change it.

davidbiancolin:
I'm going to play with schemes to detect spot interruptions in a follow up PR, its not clear to me what the best way to do that is yet.

abejgonzalez (approved):
LGTM.

PR #1037, opened 2022-04-22
awstools typing + small organization

abejgonzalez:
Taken from #1016.

- Adds initial Python typing to the `awstools.py` file (while not exhaustive it can be further refined / added to in the future).
- Splits up part of `aws_resource_names` functionality into a `get_localhost_instance_tags` function
- Adds new `get_localhost_instance_info` and `get_localhost_instance_id` functions
- This also does the following changes as recommended by @ timsynder-siv.
  - Renames "manager"-named functions to "localhost" - https://github.com/firesim/firesim/pull/1016/files#r854549774
  - Catches curl timeout errors - https://github.com/firesim/firesim/pull/1016/files#r854543162
  - Fails when unable to retrieve the instance id - https://github.com/firesim/firesim/pull/1016/files#r854522269
  - Fixes list(list(XYZ)) error - https://github.com/firesim/firesim/pull/1016/files#r854505356

#### Related PRs / Issues

#1016

### Contributor Checklist
- [ ] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [ ] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous prints/debugging code?
- [ ] Did you state the UI / API impact?
- [ ] Did you specify the Verilog / AGFI compatibility impact?

- [ ] If applicable, did you regenerate and publicly share default AGFIs?

- [ ] If applicable, did you apply the `ci:fpga-deploy` label?

- [ ] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1042, opened 2022-04-24
Add Python typing to all files + Cleanup based off it

abejgonzalez:
I started doing this in my "where to run" PR but figured it should probably be split out since static type checking catches a lot of potential bugs in the code (checking `None`s, improper member variable setup, etc).

This does the following:
- Adds types to all of `runtools` (some `Any` types could be further refined but I'll leave that for a future PR)
- Uses `__future__.annotations` everywhere (to make annotations strings - speeds things up) - Fixes bugs based on this in the `register_task` `firesim` decorator
- Avoids the use of `from ... import *` - Needed so that global variables aren't redefined (causing the type checker to not infer certain variables like `rootLogger`)
- Reorders the `__init__` of `FireSimTopology` to the top of the class (like other classes we write)
- Uses `f"..."` for the `basecommand` in `RuntimeConfig`
- Reworks the `add_simulation` and `add_switch` of the run farm instances (in `run_farm.py`) to not use `None`s. This should be the same functionality since the `None`s were filled in from 0 to MAX (i.e. slots were filled contiguously). Do we care about the other potential functionality of having the slots filled in a non-contiguous manner (slot 0, 3, 5, 6 filled)? IMO I don't think we care about this on F1, and in the future we can just override this on different run farm host types.

This depends on #1037 going in.

Sorry in advance for the big PR (much of it is pretty easy to understand).

**TODO**:
- [x] Fix `rootLogger` type inference
- [x] Misc. cleanup
- [x] Make sure there are no circular dependencies
- [ ] Run FPGA sim to confirm all is working (will run when this is about to be merged into main through CI)

### Contributor Checklist


deploy/runtools/firesim_topology_elements.py
```diff
@@ -495,15 +585,18 @@ def run_sim_start_command(self, slotno):
         if self.uplinks:
             all_shmemportnames = [self.uplinks[0].get_global_link_id()] + [self.supernode_get_sibling_shmemportname(x) for x in range(1, num_siblings)]
 
+        assert self.server_hardware_config is not None
```
abejgonzalez, line 588: General comment: Across all of this code I have to add these asserts so that the type checker doesn't error. These need to be analyzed much more to make sure they are in the right places (or maybe we can remove some extra `Optional` types).

deploy/runtools/utils.py
```diff
@@ -360,10 +362,11 @@ def get_local_shared_libraries(elf):
     ]
 
     libs = list()
-    rootLogger.debug(f"Identifying ldd dependencies for:{elf}")
+    rootLogger.debug(f"Identifying ldd dependencies for: {elf}")
     for dso in lddwrap.list_dependencies(Path(elf)):
         if dso.soname is None:
-            assert '/ld-linux' in fspath(dso.path), f"dynamic linker is only allowed no soname, not: {dso}"
+            if dso.path is not None:
```
abejgonzalez, line 368: @timsnyder-siv the type checker errors here since dso.path could potentially be `None`. What should this be?

deploy/runtools/firesim_topology_elements.py
```diff
@@ -248,6 +288,12 @@ def run_sim_start_command(self, slotno):
         all_bootbins = [self.get_bootbin_name()]
         all_shmemportnames = [shmemportname]
 
+        assert self.server_hardware_config is not None and isinstance(self.server_hardware_config, RuntimeHWConfig)
```
abejgonzalez, line 291: Probably can create a new function called something like `resolved_server_hw_cfg` that has these asserts built-in. Maybe we can have the `server_hw_cfg` be split into `server_hw_cfg` and `server_hw_cfg_name` so that we don't need this?

deploy/runtools/firesim_topology_elements.py
```diff
@@ -336,55 +382,66 @@ def copy_back_job_results_from_run(self, slotno):
...
 
-    def get_required_files_local_paths(self):
+    def get_required_files_local_paths(self) -> List[Tuple[str, str]]:
         """ Return local paths of all stuff needed to run this simulation as
         an array. """
         all_paths = []
 
-        if self.get_job().rootfs_path() is not None:
-            all_paths.append([self.get_job().rootfs_path(), self.get_rootfs_name()])
+        job_rootfs_path = self.get_job().rootfs_path()
```
abejgonzalez, line 397: IDK why but this was surprisingly hard to get `mypy` to cooperate hence why this is verbose now.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -442,68 +507,64 @@ def supernode_get_num_siblings_plus_one(self):
...
+        uplink_side = self.uplinks[0].get_uplink_side()
+        for index, servernode in enumerate(map(lambda x : x.get_downlink_side(), uplink_side.downlinks)):
             if self == servernode:
-                return self.uplinks[0].get_uplink_side().downlinks[index+siblingindex].get_downlink_side()
+                node = uplink_side.downlinks[index+siblingindex].get_downlink_side()
+                assert isinstance(node, FireSimDummyServerNode)
+                return node
+        assert False, "Should return supernode sibling"
 
-    def supernode_get_sibling_mac_address(self, siblingindex):
```
abejgonzalez, line 452: Figured this was too verbose? I just inlined these specific calls further down the file. I can revert this if people want.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -442,68 +507,64 @@ def supernode_get_num_siblings_plus_one(self):
...
-        all_linklatencies = [self.server_link_latency] + [self.supernode_get_sibling_link_latency(x) for x in range(1, num_siblings)]
-        all_maxbws = [self.server_bw_max] + [self.supernode_get_sibling_bw_max(x) for x in range(1, num_siblings)]
+        all_macs = [self.get_mac_address()]
+        all_rootfses = [self.get_rootfs_name()]
+        all_bootbins = [self.get_bootbin_name()]
+        all_linklatencies = [self.server_link_latency]
+        all_maxbws = [self.server_bw_max]
+
+        for x in range(1, num_siblings):
+            sibling = self.supernode_get_sibling(x)
```
abejgonzalez, line 534: Can probably convert this back into the `[self.supernode_get_sibling(x).FOO_TO_CALL]` syntax

deploy/runtools/firesim_topology_elements.py
```diff
@@ -442,68 +507,64 @@ def supernode_get_num_siblings_plus_one(self):
...
+            self.autocounter_readrate is not None and all_shmemportnames is not None and self.zerooutdram is not None and self.disable_asserts is not None and
+            self.print_start is not None and self.print_end is not None and self.print_cycle_prefix is not None)
 
         runcommand = self.server_hardware_config.get_boot_simulation_command(
             slotno, all_macs, all_rootfses, all_linklatencies, all_maxbws,
             self.server_profile_interval, all_bootbins, self.trace_enable,
             self.trace_select, self.trace_start, self.trace_end, self.trace_output_format,
-            self.autocounter_readrate, all_shmemportnames, self.zerooutdram)
+            self.autocounter_readrate, all_shmemportnames, self.zerooutdram, self.disable_asserts,
+            self.print_start, self.print_end, self.print_cycle_prefix)
```
abejgonzalez, line 563: Who tests supernode anyways?!?

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -226,15 +258,16 @@ def mapping_use_one_f1_16xlarge(self):
 
         for switch in switches:
             self.run_farm.f1_16s[f1_16s_used].add_switch(switch)
-            downlinknodes = map(lambda x: x.get_downlink_side(), switch.downlinks)
-            if all([isinstance(x, FireSimServerNode) for x in downlinknodes]):
+            alldownlinknodes = map(lambda x: x.get_downlink_side(), switch.downlinks)
+            if all([isinstance(x, FireSimServerNode) for x in alldownlinknodes]):
+                downlinknodes = cast(List[FireSimServerNode], alldownlinknodes)
```
abejgonzalez, line 263: Again, this is more verbose because `mypy` can't detect that `alldownlinknodes` are of type `FireSimServerNode` from this `if` statement. Maybe this is re-workable but this works for now.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -268,31 +300,34 @@ def pass_perform_host_node_mapping(self):
...
             3) In either case, call get_deploytriplet_for_config() once to
             make the API call and cache the result for the deploytriplet.
         """
         servers = self.firesimtopol.get_dfs_order_servers()
-        defaulthwconfig_obj = self.hwdb.get_runtimehwconfig_from_name(self.defaulthwconfig)
 
         for server in servers:
-            servhwconf = server.get_server_hardware_config()
-            if servhwconf is None:
+            hw_cfg = server.get_server_hardware_config()
```
abejgonzalez, line 317: I personally think this code is very confusing... in some places you expected the HW_CFG to be a `str` and in others you wanted a `RuntimeHWConfig` object. This keeps the same functionality but IMO this is ugly.

deploy/runtools/run_farm.py
```diff
@@ -102,63 +118,63 @@ def allocate_host_port(self):
...
         """ Get the number of fpga slots. """
         return self.FPGA_SLOTS
 
-    def get_num_fpga_slots_consumed(self):
+    def get_num_fpga_slots_consumed(self) -> int:
         """ Get the number of fpga slots. """
         return self.fpga_slots_consumed
 
-    def add_simulation(self, firesimservernode):
+    def add_simulation(self, firesimservernode: FireSimServerNode) -> None:
```
abejgonzalez, line 138: I reworked the `add_simulation` and `add_switch` to avoid having to use `Optional` types. This is still the same functionality as before.

deploy/runtools/user_topology.py
```diff
@@ -261,29 +273,32 @@ def supernode_flatten(arr):
             res = res + x
         return res
 
-    def supernode_example_6config(self):
+    def supernode_example_6config(self) -> None:
         self.roots = [FireSimSwitchNode()]
-        servers = [FireSimSuperNodeServerNode()] + [FireSimDummyServerNode() for x in range(5)]
```
abejgonzalez, line 266: This errored since a `List[SuperNode]` couldn't be added to a `List[DummyNode]`s.

deploy/runtools/user_topology.py
```diff
@@ -1,14 +1,27 @@
 """ Define your additional topologies here. The FireSimTopology class inherits
 from UserToplogies and thus can instantiate your topology. """
 
-from runtools.firesim_topology_elements import *
```
abejgonzalez, line 4: Having `import *` statements in the code caused things like the `rootLogger` global variable to be imported from other modules (and thus making mypy unable to infer it's type). Across the codebase I removed this and explicitly imported each object/class.

deploy/runtools/runtime_config.py
```diff
@@ -129,52 +139,26 @@ def array_to_lognames(values, prefix):
 
         command_bootbinaries = array_to_plusargs(all_bootbinaries, "+prog")
         zero_out_dram = "+zero-out-dram" if (enable_zerooutdram) else ""
-        disable_asserts_arg = "+disable-asserts" if (disable_asserts) else ""
+        disable_asserts = "+disable-asserts" if (disable_asserts_arg) else ""
```
abejgonzalez, line 142: Renamed this so that the `basecommand` line could be reduced to a `f"..."` string.

deploy/runtools/runtime_config.py
```diff
@@ -1,36 +1,44 @@
...
-import os
-import sys
+
+from typing import Optional, Dict, Any, List, Sequence, TYPE_CHECKING
+import argparse # this is not within a if TYPE_CHECKING: scope so the `register_task` in FireSim can evaluate it's annotation
+if TYPE_CHECKING:
+    from runtools.utils import MacAddress
 
 LOCAL_DRIVERS_BASE = "../sim/output/f1/"
-LOCAL_DRIVERS_GENERATED_SRC = "../sim/generated-src/f1/"
```
abejgonzalez, line 24: Unused code.

deploy/runtools/run_farm.py
```diff
@@ -60,40 +71,48 @@ def get_nbd_for_imagename(self, imagename):
...
+    def is_bound_to_real_instance(self) -> bool:
         return self.boto3_instance_object is not None
 
-    def get_private_ip(self):
+    def get_private_ip(self) -> str:
+        assert self.boto3_instance_object is not None
         return self.boto3_instance_object.private_ip_address
 
-    def add_switch(self, firesimswitchnode):
+    def add_switch(self, firesimswitchnode: FireSimSwitchNode) -> None:
```
abejgonzalez, line 104: See https://github.com/firesim/firesim/pull/1042/files#r857293207

deploy/runtools/run_farm.py
```diff
@@ -361,34 +390,34 @@ def terminate_run_farm(self, terminatesomef1_16, terminatesomef1_4, terminatesom
         else:
             rootLogger.critical("Termination cancelled.")
 
-    def get_all_host_nodes(self):
+    def get_all_host_nodes(self) -> List[EC2Inst]:
         """ Get objects for all host nodes in the run farm that are bound to
         a real instance. """
-        allinsts = self.f1_16s + self.f1_2s + self.f1_4s + self.m4_16s
-        return [inst for inst in allinsts if inst.boto3_instance_object is not None]
+        allinsts: List[EC2Inst] = [*self.f1_16s, *self.f1_2s, *self.f1_4s, *self.m4_16s]
```
abejgonzalez, line 396: This is hacky because lists of different types can't be added together. The type checker doesn't work in this scenario: 

```
class A(X)
class B(X)
class C(X)

all: List[X] = [A] + [B] + [C] # errors since A can't be added to B (doesn't resolve to X)
```

deploy/runtools/firesim_topology_elements.py
abejgonzalez: Ok I did this.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -181,21 +214,27 @@ def __init__(self, server_hardware_config=None, server_link_latency=None,
...
-    def set_server_hardware_config(self, server_hardware_config):
+    def set_server_hardware_config(self, server_hardware_config: RuntimeHWConfig) -> None:
         self.server_hardware_config = server_hardware_config
 
-    def get_server_hardware_config(self):
+    def get_server_hardware_config(self) -> Optional[Union[RuntimeHWConfig, str]]:
+        return self.server_hardware_config
+
+    def get_resolved_server_hardware_config(self) -> RuntimeHWConfig:
+        assert self.server_hardware_config is not None and isinstance(self.server_hardware_config, RuntimeHWConfig)
```
abejgonzalez, line 227:
```suggestion
        assert self.server_hardware_config is not None and not isinstance(self.server_hardware_config, str)
```

deploy/runtools/utils.py
timsnyder-siv: When I checkout your `runtools-mypy` branch and `./scripts/run-py-typecheck.sh`, I don't see an error.  What am I missing?

deploy/awstools/afitools.py
```diff
@@ -1,5 +1,7 @@
 """ Tools to help manage afis. """
 
+from __future__ import annotations
```
timsnyder-siv, line 3: Thank you for doing this across all files.  Since it changes the fundamental behavior of type annotations, it is good to be consistent in how we're using it.  It would be nice if mypy had a check (or something somewhere) that makes sure that we continue to have the import in all python files.

deploy/runtools/user_topology.py
timsnyder-siv: Even if you do this?
```
servers : List[FireSimServerNode] = [FireSimSuperNodeServerNode()] + [FireSimDummyServerNode() for x in range(5)]
```

from https://mypy.readthedocs.io/en/latest/common_issues.html#variance

timsnyder-siv: For my own edification, I confirmed that with this:
```
diff --git a/deploy/runtools/user_topology.py b/deploy/runtools/user_topology.py
index 4336b4b14..a7173fd37 100644
--- a/deploy/runtools/user_topology.py
+++ b/deploy/runtools/user_topology.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 from runtools.firesim_topology_elements import FireSimSwitchNode, FireSimServerNode, FireSimSuperNodeServerNode, FireSimDummyServerNode, FireSimNode
 
-from typing import Optional, Union, Callable, Sequence, TYPE_CHECKING
+from typing import List, Optional, Union, Callable, Sequence, TYPE_CHECKING
 if TYPE_CHECKING:
     from runtools.firesim_topology_with_passes import FireSimTopologyWithPasses
 
@@ -275,8 +275,8 @@ class UserTopologies:
 
     def supernode_example_6config(self) -> None:
         self.roots = [FireSimSwitchNode()]
-        self.roots[0].add_downlinks([FireSimSuperNodeServerNode()])
-        self.roots[0].add_downlinks([FireSimDummyServerNode() for x in range(5)])
+        servers : List[FireSimServerNode] = [FireSimSuperNodeServerNode()] + [FireSimDummyServerNode() for x in range(5)]
+        self.roots[0].add_downlinks(servers)
 
     def supernode_example_4config(self) -> None:
         self.roots = [FireSimSwitchNode()]
Submodule target-design/chipyard contains modified content
```

I still see:
```
\u276f ./scripts/run-py-typecheck.sh 
deploy/runtools/user_topology.py:278: error: Incompatible types in assignment (expression has type "List[FireSimSuperNodeServerNode]", variable has type "List[FireSimServerNode]")
deploy/runtools/user_topology.py:278: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
deploy/runtools/user_topology.py:278: note: Consider using "Sequence" instead, which is covariant
deploy/runtools/user_topology.py:278: error: List comprehension has incompatible type List[FireSimDummyServerNode]; expected List[FireSimSuperNodeServerNode]
Found 2 errors in 1 file (checked 22 source files)

```

timsnyder-siv: This works:
```
diff --git a/deploy/runtools/user_topology.py b/deploy/runtools/user_topology.py
index 4336b4b14..ce3479c1b 100644
--- a/deploy/runtools/user_topology.py
+++ b/deploy/runtools/user_topology.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 from runtools.firesim_topology_elements import FireSimSwitchNode, FireSimServerNode, FireSimSuperNodeServerNode, FireSimDummyServerNode, FireSimNode
 
-from typing import Optional, Union, Callable, Sequence, TYPE_CHECKING
+from typing import cast, List, Optional, Union, Callable, Sequence, TYPE_CHECKING
 if TYPE_CHECKING:
     from runtools.firesim_topology_with_passes import FireSimTopologyWithPasses
 
@@ -275,8 +275,8 @@ class UserTopologies:
 
     def supernode_example_6config(self) -> None:
         self.roots = [FireSimSwitchNode()]
-        self.roots[0].add_downlinks([FireSimSuperNodeServerNode()])
-        self.roots[0].add_downlinks([FireSimDummyServerNode() for x in range(5)])
+        servers = [cast(FireSimServerNode, FireSimSuperNodeServerNode())] + [cast(FireSimServerNode, FireSimDummyServerNode()) for x in range(5)]
+        self.roots[0].add_downlinks(servers)
 
     def supernode_example_4config(self) -> None:
         self.roots = [FireSimSwitchNode()]
Submodule target-design/chipyard contains modified content

```

But I like the way you rewrote it to avoid the problem better.  I really hate having to rewrite reasonable code for static checkers but whatever.

abejgonzalez:
@timsnyder-siv IDK why I can't comment on this: https://github.com/firesim/firesim/pull/1042#discussion_r859030911. But I added the extra `if dso.path is not None:` check. Is that correct?

deploy/runtools/utils.py
timsnyder-siv: > IDK why I can't comment on this: https://github.com/firesim/firesim/pull/1042#discussion_r859030911. But I added the extra if dso.path is not None: check. Is that correct?

I'm only able to comment in the thread if I go to the "Files Changed" tab, not in Conversations for some reason.

I looked back though lddwrap and I think I really meant 
```
       assert dso.path is not None and '/ld-linux' in fspath(dso.path), ...
```

I don't want it to silently continue past anything other than the dynamic linker in that case.

deploy/runtools/utils.py
```diff
@@ -360,10 +364,11 @@ def get_local_shared_libraries(elf):
...
+                assert '/ld-linux' in fspath(dso.path), f"dynamic linker is only allowed no soname, not: {dso}"
```
timsnyder-siv, line 371:
```suggestion
            assert dso.path is not None and '/ld-linux' in fspath(dso.path), f"dynamic linker is only allowed no soname, not: {dso}"
```

timsnyder-siv (changes requested):
Definitely should change this before merging.

deploy/runtools/runtime_config.py
```diff
@@ -129,52 +139,26 @@ def array_to_lognames(values, prefix):
...
-            tracefile=tracefile,
-            trace_output_format=trace_output_format,
-            dwarf_file_name=dwarf_file_name,
-            autocounterfile=autocounterfile,
-            autocounter_readrate=autocounter_readrate,
-            command_dromajo=command_dromajo,
-            print_cycle_prefix=print_cycle_prefix,
-            print_start=print_start,
-            print_end=print_end)
+        basecommand = f"""screen -S fsim{slotid} -d -m bash -c "script -f -c 'stty intr ^] && sudo ./{driver} +permissive $(sed \':a;N;$!ba;s/\\n/ /g\' {runtimeconf}) +slotid={slotid} +profile-interval={profile_interval} {zero_out_dram} {disable_asserts} {command_macs} {command_rootfses} {command_niclogs} {command_blkdev_logs}  {tracefile} +trace-select={trace_select} +trace-start={trace_start} +trace-end={trace_end} +trace-output-format={trace_output_format} {dwarf_file_name} +autocounter-readrate={autocounter_readrate} {autocounterfile} {command_dromajo} {print_cycle_prefix} +print-start={print_start} +print-end={print_end} {command_linklatencies} {command_netbws}  {command_shmemportnames} +permissive-off {command_bootbinaries} && stty intr ^c' uartlog"; sleep 1"""
```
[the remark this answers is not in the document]
abejgonzalez: I did this in the #1028 so I'll skip it in this PR for now.

deploy/runtools/run_farm.py
```diff
@@ -361,34 +376,34 @@ def terminate_run_farm(self, terminatesomef1_16, terminatesomef1_4, terminatesom
...
 
-    def lookup_by_ip_addr(self, ipaddr):
+    def lookup_by_ip_addr(self, ipaddr: str) -> EC2Inst:
         """ Get an instance object from its IP address. """
         for host_node in self.get_all_host_nodes():
             if host_node.get_private_ip() == ipaddr:
                 return host_node
-        return None
-
+        assert False, f"Unable to find host node by {ipaddr} host name"
```
abejgonzalez, line 390: This should be safe because you assign IP addresses instances before this call.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -541,6 +577,7 @@ def loop_logger(instancestates, terminateoncompletion):
             rootLogger.info("""FireSim Simulation Status @ {}""".format(str(datetime.datetime.utcnow())))
             rootLogger.info("-"*80)
             rootLogger.info("""This workload's output is located in:\n{}""".format(self.workload.job_results_dir))
+            assert isinstance(rootLogger.handlers[0], logging.FileHandler)
```
abejgonzalez, line 580: This is fine because it is assigned in the top-level `firesim` py script.

deploy/runtools/run_farm.py
```diff
@@ -489,59 +504,59 @@ def unload_xdma(self):
...
-            with StreamLogger('stdout'), StreamLogger('stderr'):
-                remote_kmsg("""about_to_check_clear_fpga{}""".format(slotno))
-                run("""until sudo fpga-describe-local-image -S {} -R -H | grep -q "cleared"; do  sleep 1;  done""".format(slotno))
-                remote_kmsg("""done_checking_clear_fpga{}""".format(slotno))
+            for slotno in range(self.parentnode.get_num_fpga_slots_max()):
+                self.instance_logger("""Checking for Cleared FPGA Slot {}.""".format(slotno))
+                with StreamLogger('stdout'), StreamLogger('stderr'):
+                    remote_kmsg("""about_to_check_clear_fpga{}""".format(slotno))
+                    run("""until sudo fpga-describe-local-image -S {} -R -H | grep -q "cleared"; do  sleep 1;  done""".format(slotno))
+                    remote_kmsg("""done_checking_clear_fpga{}""".format(slotno))
```
timsnyder-siv, line 522: How does this look once you have 'what-to-run' implemented?  I strongly dislike the current code silently does nothing `unless isinstance(self.parentnode, F1_Instance)` if for no other reason than we no longer "always clear ALL fpga slots" when `clear_fpgas` is called.  Should probably make this a `match` statement with default case that throws `NotImplemented` so that we don't open the door for really difficult to understand silent failure cases.

However, if 'where-to-run' takes care of this by abstracting this method behind a dispatcher that has to implement this method, I'm willing to look the other way here.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -620,8 +663,9 @@ def copy_back_switchlog_from_run(self, job_results_dir, switch_slot_no):
                 get(remote_path=remote_sim_run_dir + simoutputfile, local_path=job_dir)
 
 
-    def diagramstr(self):
-        msg = """{}:{}\n---------\ndownlinks: {}\nswitchingtable: {}""".format(
-            "FireSimSwitchNode", str(self.switch_id_internal), ", ".join(map(str, self.downlinkmacs)),
-            ", ".join(map(str, self.switch_table)))
+    def diagramstr(self) -> str:
+        msg =  f"FireSimSwitchNode: {self.switch_id_internal}\n"
```
sagark, line 667:
```suggestion
        msg =  f"FireSimSwitchNode:{self.switch_id_internal}\n"
```

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -252,47 +284,50 @@ def pass_perform_host_node_mapping(self):
...
         else:
             assert False, "IMPROPER MAPPING CONFIGURATION"
 
-    def pass_apply_default_hwconfig(self):
+    def pass_apply_default_hwconfig(self) -> None:
         """ This is the default mapping pass for hardware configurations - it
         does 3 things:
             1) If a node has a hardware config assigned (as a string), replace
-            it with the appropriate RuntimeHWConfig object.
+            it with the appropriate RuntimeHWConfig object. If it already a
```
sagark, line 306:
```suggestion
            it with the appropriate RuntimeHWConfig object. If it is already a
```

deploy/runtools/run_farm.py
abejgonzalez: These if's are really no-ops (with how things are currently used). In reality, these functions are called properly in `infrasetup` (https://github.com/firesim/firesim/blob/043905e14ddb304705226738df14dc8885faf436/deploy/runtools/run_farm.py#L676)(or other types of functions) with an `if` in that function doing the same thing. If the preference is adding an assert instead of silently skipping that's fine with me. I was just trying to avoid adding typing changes that affect the original code too much.

sagark (approved):
LGTM modulo small suggestions.

deploy/runtools/run_farm.py
abejgonzalez: @timsnyder-siv is this the only issue with the PR? If so I'll address this and I think we can merge this (since it doesn't change any functionality)

abejgonzalez:
@mergifyio rebase

deploy/runtools/run_farm.py
abejgonzalez: Solved offline. We will revert to using `assert`s

PR #1056, opened 2022-05-11
VSCode Integration for Scala Development

davidbiancolin:
I'm not one to mandate the use of an IDE on anyone, _however_: 
* VSCode is free, and has a super-rich set of plugins, including fantastic ones for git (gitlens) and scala (metals)
* VSCode likely has bindings for your editor and can slurp in your existing configurations files (mostly). 
* Getting a language server running under your editor correctly with FireSim and Chipyard is annoying. This PR does that for VSCode. 

Selfishly, we use VSCode at SiFive and i want my new hires to have a consistent development environment. I'll add more docs as i use VSCode in other parts of the codebase. 

One gotcha: we need to add bloop to Chipyard's sbt plugins for this to work. We'll have to do that manually for now, since I don't want to go through the whole rigamarole of a chipyard bump. There's a warning in the docs that we can remove when the bump happens eventually.

#### UI / API Impact
None.

NC

### Contributor Checklist


docs/Developer-Docs/VSCode-Integration.rst
```diff
@@ -0,0 +1,82 @@
...
+
+#. **No test task support for ScalaTests that use make.** Due to the way
+   FireSim's ScalaTest calls out to make to invoke the generator and Golden Gate, Metals's bloop instance
+   must initialized with ``env.sh`` sourced. This will be resolved in a future PR. 
+
+Other Notes
+###########
+
+Reliance on SBT multi-project builds breaks the default metals integration. To hide this, we've put workspace-specific settings for metals in
+sim/.vscode/setting.json which shouljd permit metals to run correctly out of
```
michael-etzkorn, line 77: small typo here "shouljd" -> "should"

davidbiancolin: hey @michael-etzkorn, feel free to suggest fixes via github cmd+g/ctrl+g or w.e, and i can easily merge them in.

michael-etzkorn: Ah, I'm on mobile and wasn't aware of that feature (and don't know the mobile equivalent). I'll suggest the change when I get the chance.

davidbiancolin: Np np. Thanks for catching it. I'll fix it now.

sim/.vscode/sbt-launch.sh
```diff
@@ -0,0 +1,12 @@
...
+# This script is used by VSCode only to launch sbt and run bloopInstall. It
+# performs no other tasks (compile, run, test, etc..), so I've omitted many
+# of the other command line flags we pass through the make system.
+
+# If we need these to agree we should consider moving stuff into .jvmopts and
+# .sbtopts and out of the Makefiles
+
+set -ex
+source ../env.sh
+java -jar /home/centos/firesim-global-reset/target-design/chipyard/generators/rocket-chip/sbt-launch.jar "$@"
```
abejgonzalez, line 12: Bug? 

If we are trying to avoid having to point to the `sbt-launch.jar` we can instead just use the `sbt` command that is provided by conda. Otherwise, I would just make this a relative path.

docs/Developer-Docs/VSCode-Integration.rst
```diff
@@ -0,0 +1,84 @@
...
+
+    Host ec2-manager
+        User centos
+        IdentityFile ~/.ssh/<your-firesim.pem>
+        Hostname <IP ADDR>
+    
+#. In VSCode, using the ``Remote Manager`` on the left sidebar, connect to your manager instance.
+#. Open a workspace in your FireSim clone under ``sim/``.
+#. First time per remote: install the Metals plugin on the *remote* machine.
+#. Metals will prompt you with the following: "New SBT Workspace Detected, would you like to import the build?". Click *Don't Show Again*.
```
abejgonzalez, line 57: I would clarify this a bit and say you don't want to import the build so you should click the "ignore" option. (Edit: Is this what you want? The next sentence makes it seem like you should import the build...)

docs/Developer-Docs/VSCode-Integration.rst
```diff
@@ -0,0 +1,84 @@
...
+
+#. **No test task support for ScalaTests that use make.** Due to the way
+   FireSim's ScalaTest calls out to make to invoke the generator and Golden Gate, Metals's bloop instance
+   must initialized with ``env.sh`` sourced. This will be resolved in a future PR. 
+
+Other Notes
+###########
+
+Reliance on SBT multi-project builds breaks the default metals integration. To hide this, we've put workspace-specific settings for metals in
+sim/.vscode/setting.json which should permit metals to run correctly out of
```
abejgonzalez, line 79:
```suggestion
``sim/.vscode/setting.json`` which should permit metals to run correctly out of
```

abejgonzalez (commented):
Seems good to me. Can you remind me why the default SBT integration doesn't work with this? Why do we need bloop?

sim/.vscode/sbt-launch.sh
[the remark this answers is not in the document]
davidbiancolin: sourcing env.sh should suffice as a check of that (since the file won't be generated otherwise).

docs/Developer-Docs/VSCode-Integration.rst
```diff
@@ -0,0 +1,84 @@
...
+running on your remote manager.
+
+
+General Setup
+=============
+
+#. Install VSCode. You can grab installers `here <https://code.visualstudio.com/download>`_.
+#. Open VSCode and install the ``Remote Developer Plugin``. See the `marketplace page <https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack>`_ for a complete description of its features.
+
+At this point, VSCode will slurp in your `.ssh/config`. Hosts you've listed
```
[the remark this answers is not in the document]
davidbiancolin: Here, read, basically. I'll be less floral.

docs/Developer-Docs/VSCode-Integration.rst
```diff
@@ -0,0 +1,84 @@
...
+
+.. warning:: Until Chipyard is bumped, you must add bloop to chipyard's plugins.sbt for this to work correctly. See :gh-file-ref:`sim/project/plugins.sbt` and copy the bloop installation into ``target-design/chipyard/project/plugins.sbt``.
+
+VSCode has rich support for Scala development, and the `Metals <https://scalameta.org/metals/docs/editors/vscode/>`_ plugin is really what makes the magic happen.
+
+How To Use (Remote Manager)
+###########################
+
+#. If you haven't already, clone FireSim and run build-setup.sh on your manager.
+#. Ensure your manager instance is listed as a host in your ``.ssh/config``. For example:
```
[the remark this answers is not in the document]
davidbiancolin: You need this so that the remote explorer shows available hosts neatly. 

Personally, i have multiple managers set up and they are named such that most of their settings derive from a single wildcard pattern.

docs/Developer-Docs/VSCode-Integration.rst
davidbiancolin: This will autoimport the build in the future. I'll try to clarify the text here.

timsnyder-siv:
When I search for something, with FireSim as top, I still see two copies of it.  One in the correct location and then another down in `target-design/Chipyard/sims/firesim`.  Can this be addressed?  For other IDE's I've always manually told them to ignore the `target-design/chipyard/sims/firesim` tree.  Why is that even populated when firesim is top?

davidbiancolin:
> When I search for something, with FireSim as top, I still see two copies of it. One in the correct location and then another down in `target-design/Chipyard/sims/firesim`. Can this be addressed? For other IDE's I've always manually told them to ignore the `target-design/chipyard/sims/firesim` tree. Why is that even populated when firesim is top?

I'll try to resolve this as part of this PR.

davidbiancolin:
> Seems good to me. Can you remind me why the default SBT integration doesn't work with this? Why do we need bloop?

Default SBT integration doesn't work because without the right env variables set, firesim `build.sbt` thinks it's running under chipyard and looks for sources in the wrong place. 

Bloop is the build server used by metals by default. Metals tries to install bloop without infecting the user's build sources, essentially by adding an auxilary project under `sim/`, but it doesn't know that it needs to do the same process under `target-design/chipyard` for this to work correctly.

abejgonzalez (approved):
LGTM

PR #992, opened 2022-03-18
Modifying AutoILA wiring pass for local FPGA support

jamesdunn:
#### UI / API Impact

No API impact; signal annotation remains the same.

#### Verilog / AGFI Compatibility

Yes, changes the generated Verilog.

Current ILA pass:

- Punches out annotated ports to F1Shell with a map of FpgaDebugAnnotation to TopWiringAnnotation.
- Emits TCL to generate ILA Verilog.
- Emits Verilog files for ports, module instantiation, and wiring the generated ILA module to top-level module.

This assumes something will wrap and instantiate the GoldenGate-emitted Verilog, which will not be the case moving forward for local FPGA support.

This pass:
- Uses string literal to generate the mostly static ILA Tcl definition.
- Runs TopWiring transform as before to route out ILA-annotated signals by creating ports and connect statements.
- Blackboxes ILA Verilog module within the top level.
- Removes the punched out ports from the top level, flips them, and uses them as inputs to the blackboxed module.
- Uses a map to re-wire the existing connect statements from FPGATop to ILA blackbox, instead of FPGATop to F1Shell

This is an incomplete draft for starting a review. Some of the above is implemented, with sketches for a few things that I have questions on.

### Contributor Checklist
- [ ] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [ ] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous prints/debugging code?
- [ ] Did you state the UI / API impact?
- [ ] Did you specify the Verilog / AGFI compatibility impact?

- [ ] If applicable, did you regenerate and publicly share default AGFIs?

- [ ] If applicable, did you apply the `ci:fpga-deploy` label?

- [ ] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1059, opened 2022-05-23
A host-portable AutoILA transform that instantiates the black box in IR

davidbiancolin:
This minimizes the amount of work required to bring up auto ILA on another Xilinx-based host, and supports having `PlatformShim`s that are the top-level (as in Vitis), as all the enclosing FPGA project needs to do is elaborate the ILA IP. 

I've moved keys and config fragments out of firesim-lib and into `midas` proper. Moving forward i hope to anything that isn't source-dependent on testchipip into midas.

h/t @jamesdunn For doing the initial work. 

#### Related PRs / Issues

#992 

#### UI / API Impact

Expanded. I made it possible to specify the number of comparators per-probe as a config option instead of being hardcoded to two.

I'll regenerate AGFIs because this touches FPGA project sources. 

### Contributor Checklist


davidbiancolin:
This is ready for review.

davidbiancolin:
This appears to work fine on EC2.

abejgonzalez (approved):
SGTM. Can you do a pass on the docs on the API to modify the ILADepth/Triggers. In the past, it was just adding a mixin (like `ILADepth8192` but I'm not sure that still applies)? Does it?

sim/midas/src/main/scala/midas/Config.scala
```diff
@@ -19,9 +19,23 @@ case object Platform extends Field[(Parameters) => PlatformShim]
...
+/** When set, [[targetutils.FPGADebugAnnotation]]s (signals labelled with FPGADebug())
+  * are automatically wired out to an ILA.
+  */
 case object EnableAutoILA extends Field[Boolean](false)
 
+/** Sets a per-probe buffer depth on the ILA. A greater value permits capturing a longer
+  * waveform, at the expense of FPGA reasources. See PG172
+  * ([200~https://docs.xilinx.com/v/u/en-US/pg172-ila) for more info.
+  */
+case object ILADepthKey extends Field[Int](1024)
```
abejgonzalez, line 31: Seems like this key is a duplicate of this key: https://github.com/firesim/firesim/blob/716b28d65438f94b1910c1ad658f5934fce75033/sim/firesim-lib/src/main/scala/util/Configs.scala#L16
https://github.com/firesim/firesim/blob/716b28d65438f94b1910c1ad658f5934fce75033/sim/firesim-lib/src/main/scala/util/Configs.scala#L9

Have you verified that this API still works?

davidbiancolin: Good catch. I've fixed this.

PR #1069, opened 2022-06-04
Have yes/no resolve to bool in Yaml

### LINKED ISSUE #1067: Yaml Parser Appears to Cast "yes"/"no" to Booleans, breaking config_runtime.yaml
### Background Work

- [X] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/firesim)
- [X] Yes, I searched [prior issues](https://github.com/firesim/firesim/issues)
- [X] Yes, I searched the [documentation](https://docs.fires.im/en/stable/)

### FireSim Version and Hash

Release: main

### OS Setup

Ex: Output of `uname -a` and `lsb_release -a`

### Other Setup

Ex: Prior steps taken / Documentation Followed / etc...

### Current Behavior

Setting `trace_enable: yes` does not enable tracing.

### Expected Behavior

It should enable tracing.

### Other Information

I suspect this is a problem in many places where Booleans are being compared to string literals (e.g, "yes")

abejgonzalez:
Prior to the YAML changes, enabling and disabling features was done with `yes` and `no` string matches. After the YAML change, these strings are interpreted as booleans and thus the `bool != string` matches would fail. This fixes this issue by changing the configuration logic to use booleans. This PR does not change the API of the `runtime.yaml` file (users can still write `yes` and `no` without quotes). In the future the API, can change to be `True/False` explicitly.

#### Related PRs / Issues

Fixes #1067 

#### UI / API Impact

N/A

N/A

### Contributor Checklist


deploy/runtools/runtime_config.py
```diff
@@ -316,7 +316,7 @@ def __init__(self, runtimeconfigfile: str, configoverridedata: str) -> None:
         self.print_cycle_prefix = True
 
         if 'tracing' in runtime_dict:
-            self.trace_enable = runtime_dict['tracing'].get('enable') == "yes"
+            self.trace_enable = runtime_dict['tracing'].get('enable', False)
```
davidbiancolin, line 319: So what happens if yes or no is misspelled? Does `get` have to return the same type as the default?

abejgonzalez: No it would return a string. I assumed that things would have to be spelled correctly.

davidbiancolin: Can you enforce that it returns a boolean type in this PR.

timsnyder-siv: This is an area where having a schema for the config files would be helpful.  Also note that if we're going to depend on PyYAML's implicit converson of unquoted yes/no, we're depending on YAML 1.1 and should probably document that. see https://stackoverflow.com/questions/36463531/pyyaml-automatically-converting-certain-keys-to-boolean-values

davidbiancolin: I too think we should have a schema for all of these files.

abejgonzalez: I'll make this TODO #123122232. But I'll also enforce that it returns a bool

davidbiancolin (approved):
This addresses the immediate problem. Can you scan through the other configuration yamls and ensure this pattern doesn't arise there too?

abejgonzalez:
> This addresses the immediate problem. Can you scan through the other configuration yamls and ensure this pattern doesn't arise there too?

I didn't see anything else.

PR #1090, opened 2022-06-15
check in CI if example outputs used in docs are up to date

sagark:
testing CI


sagark:
@abejgonzalez the continue-on-error failure was too subtle, so i just made it fail the entire pipeline for now.

this is ready for review

PR #1092, opened 2022-06-16
Add tutorial recipes + bump to chipyard main

sagark:
add tutorial recipes by default + bump chipyard (no more RTL changes in chipyard expected)

after this + vitis stuff goes in @abejgonzalez will re-generate default firesim AGFIs + tutorial AGFIs.


abejgonzalez (approved):
LGTM

PR #1108, opened 2022-06-23
ci: Move check docs generated components out of setup-manager

davidbiancolin:
This check should not bring down all of the CI jobs that depend on setting up the manager, so I've moved it into a separate job. This will unblock other PRs, which can be merged even with the failing check.

Aside: I think it might be better to do the generation during the docs build instead of relying on this mechanism, but I understand that is more work. 

### Contributor Checklist

### DIFF
--- a/.github/workflows/firesim-run-tests.yml
+++ b/.github/workflows/firesim-run-tests.yml
@@ -97,8 +97,6 @@ jobs:
         uses: ./.github/actions/initialize-manager
         with:
           max-runtime-hours: 10
-      - name: Check docs components that require manual re-generation (e.g. config_runtime.yaml example, help output)
-        run: ./.github/scripts/check-docs-generated-components.py
       - name: Initial Scala compilation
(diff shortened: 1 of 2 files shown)


timsnyder-siv (approved):
Looking at `.github/scripts/check-docs-generated-components.py` makes me wonder why we don't just generate those as part of the docs build but factoring this out seems like it isn't going to break things anymore than they are currently...

abejgonzalez (approved):
LGTM.

As for:

> Looking at `.github/scripts/check-docs-generated-components.py` makes me wonder why we don't just generate those as part of the docs build but factoring this out seems like it isn't going to break things anymore than they are currently...

I have some minor concerns that running `firesim managerinit` in the docs build area might be harder than it seems (we have to make sure that copy/update yaml files section of  `managerinit` works but can fail if there is no AWS access). We can push that to a later PR. 


Thanks David

davidbiancolin:
I'm going to merge this but the convo can continue.

PR #1109, opened 2022-06-24
autocounter.cc fixup split multiline string

timsnyder-siv:
gcc 4.8 on centos7 is confused by this construct.  I'm not sure if the newer gcc in conda has that problem or not.

We may want to add a config to CI that includes all of the possible bridges in the driver and increase our coverage of at least being able to compile all of the driver code.

#### UI / API Impact

No impact.

No impact.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/bridges/autocounter.cc
+++ b/sim/midas/src/main/cc/bridges/autocounter.cc
@@ -59,7 +59,7 @@ autocounter_t::autocounter_t(
             // TODO: Fix this in the bridge by not sampling with a fixed frequency
             if (this->clock_info.to_base_cycles(this->readrate) != this->readrate_base_clock) {
                 fprintf(stderr,
-"[AutoCounter] Warning: requested sample rate of %" PRIu64 " [base] cycles does not map to a whole number\n\
+"[AutoCounter] Warning: requested sample rate of %PRIu64 [base] cycles does not map to a whole number\n\
 of cycles in clock domain: %s, (%d/%d) of base clock.\n\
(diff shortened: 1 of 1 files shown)


timsnyder-siv:
@mergifyio, please rebase

timsnyder-siv:
I'm too polite...

@mergifyio rebase

davidbiancolin:
I think all of the bridges I am aware of are tested in CI. I think the conda's version of the compiler can handle this construct. 
I'm not sure why mergifyio is not rebasing for you though.

davidbiancolin (approved):

timsnyder-siv:
> I think the conda's version of the compiler can handle this construct.

Then perhaps this is just a transient issue I'm having with this code before I move forward to using the conda toolchain in my internal builds?

I'll try putting _only_ the command to mergify in a following comment.

timsnyder-siv:
@mergifyio rebase

timsnyder-siv:
Whelp, it looks like my change causes the diff in a couple of groupA tests to fail.  @davidbiancolin, I'm tempted to close this PR 'won't fix' since it only seems broken if you're using old tools and we're using much newer gcc here that can handle life.

davidbiancolin:
Maybe just use std::cerr and let C++ format the type as it sees fit?

timsnyder-siv:
Since this doesn't seem to be an issue with newer gcc, I'm going to close this unmerged.

I agree that changing the driver code to use io streams is a better longer-term solution.

PR #1117, opened 2022-07-02
Emit a fileset-grouped manifest from Golden Gate

davidbiancolin:
This should make it more clear what files are used where, allowing me to clean up some of the `replace-rtl` logic (we can drop per-platform junk), and easing integration into other build systems.

This output file manifest is rooted at `TargetDir`, all listed paths are relative to that directory. Note this does not include verilog blackboxes, and the `.f`. Collecting those files and including them here is going to be a chunk of work that might require duplicating a bunch of FIRRTL code since black box file emission is done ad-hoc does not reuse the CustomFileEmission machinery.

Here's an example emission:

```
{
  "BitstreamCompile":[
    "FireSim-generated.synthesis.xdc",
    "FireSim-generated.implementation.xdc",
    "FireSim-generated.env.tcl",
    "FireSim-generated.ila_firesim.ipgen.tcl",
    "FireSim-generated.defines.vh"
  ],
  "MetasimulatorCompile":[
    "FireSim-generated.const.h",
    "FireSim-generated.const.vh"
  ],
  "RuntimeDeployment":[
    "FireSim-generated.runtime.conf"
  ],
  "Unclassified":[
    
  ],
  "EmittedVerilogFiles":[
    "FireSim-generated.sv"
  ]
}
``` 

Some of the scala names are a little jank, i think `Fileset` is a better, shorter description of what i was trying to capture with DownstreamFlow / Dependency. 

### Contributor Checklist


timsnyder-siv (changes requested):
Overall, I think it is an improvement to have GG explicitly enumerate the optional artifacts that it is creating so that the makefile read that manifest and use it to do the right thing rather than having hardcoded artifact lists that are far removed from GG.  I also think it is a good idea to use a build-tool-agnostic representation of the manifest.  JSON is fine.

However, I think the current implementation isn't correct because it won't copy files that are added into the manifest by any GoldenGate run done by the current make invocation.  I think the solution to this is to:
1. create a tool that reads your JSON and spits out a generated makefrag representation of it that you explicitly `include` (note the intentional use of `include` instead of `-include`)
2. Add an explicit dependence from the makefrag to the manifest to the inputs of GoldenGate
3. Rethink how Distributed Elaboration works because 1&2 is definitely doing to break the 'cute' introspection of the `replace-rtl.sh` script I send to the distributed builders.  One possiblity might be to add  an additional cmdline argument like `--gen-manifest-only` to GoldenGate where it doesn't actually run and generate the 'real' outputs because those are computationally expensive (more like, my worst nightmare), but it does generate the manifest correctly (because hopefully that should be easy by comparison).

PR #1135, opened 2022-07-17
CI Runner Cleanup

abejgonzalez:
This PR updates the CI for the following:

- Deregister all runners that are `offline` when registering other runners. This fixes straggler runners that are not deregistered properly which might cause the CI to fail allocating new runners.
- Small cleanup of runner logic
- Remove all potentially conflicting runners before setting up new ones. Also remove the "global" FireSim repo before re-cloning.

### Contributor Checklist

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -1,6 +1,7 @@
 import sys
 import boto3
 import os
+import math
 from fabric.api import *
 import requests
 from ci_variables import ci_firesim_dir, local_fsim_dir, ci_gha_api_url, ci_repo_name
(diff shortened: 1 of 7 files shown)


abejgonzalez:
I tested the run-all jobs functionality with this and it seemed to work pretty well. I think this should be good to ship and it should be fine to mention that "run-all jobs" should work properly now.

.github/scripts/common.py
```diff
@@ -134,7 +148,8 @@ def change_workflow_instance_states(gh_token, tag_value, state_change, dryrun=Fa
 
     elif state_change == 'terminate':
         print("Terminating instances: {}".format(", ".join(instance_ids)))
-        deregister_runner_if_exists(gh_token, tag_value)
+        deregister_offline_runners(gh_token)
```
davidbiancolin, line 151: I know you are trying to be conservative here, but unless this is necessary (which seems like a bug), it's probably best if a workflow only touches its own runners. 

You can add more stuff to the periodic cleanup workflow for stuff you want to be workflow global.

abejgonzalez: SGTM. I'll add this to the "cull instances" workflow"

davidbiancolin:
Hey @abejgonzalez could you get this ready for merge this week. It would resolve stuff like: 
https://github.com/firesim/firesim/runs/7684767615?check_suite_focus=true#step:5:459

abejgonzalez:
Did some misc. cleanup. This should be good to merge assuming CI passes.

abejgonzalez:
@davidbiancolin Ping

davidbiancolin (approved):

PR #1136, opened 2022-07-17
Stricter Python type checking in the manager

abejgonzalez:
Ideally, all code that is added to the manager in the future has type hints. This PR helps to enforce this by erroring whenever `Any` is found in certain areas.

This modifies the type check in CI to be stricter for the following:

* Disallow functions without/insufficient type annotations
* Do type checking for interior of untyped functions (shouldn't be really used since the above check holds)
* Disallow having untyped imports that cause `Any` types

This is as strict as we can go without having to change all `Any` types to something useful (which is a lot of work). 

As part of this PR, I've updated typing across the files.

#### UI / API Impact

None. 

None.

### Contributor Checklist

### DIFF
--- a/deploy/awstools/afitools.py
+++ b/deploy/awstools/afitools.py
@@ -6,9 +6,11 @@ import logging
 import boto3
 from awstools.awstools import depaginated_boto_query
 
+from typing import List, Any, Dict, Optional
+
 rootLogger = logging.getLogger()
 
-def get_fpga_regions():
+def get_fpga_regions() -> List[str]:
     """ Get list of all regions with F1 support """
     fpga_regions = [
         'us-east-1',       # US East (N. Virginia)
@@ -18,11 +20,11 @@ def get_fpga_regions():
     ]
     return list(fpga_regions)
 
-def get_current_region():
+def get_current_region() -> str:
     boto_session = boto3.session.Session()
     return boto_session.region_name
 
-def get_afi_for_agfi(agfi_id, region=None):
+def get_afi_for_agfi(agfi_id: str, region: Optional[str] = None) -> str:
     """ Get the AFI for the AGFI in the specified region.
     region = None means use default region
     (AFIs are region specific, AGFIs are global).
@@ -46,7 +48,7 @@ def get_afi_for_agfi(agfi_id, region=None):
     rootLogger.debug(fpga_images_all)
     return fpga_images_all[0]['FpgaImageId']
 
-def copy_afi_to_all_regions(afi_id, starting_region=None):
+def copy_afi_to_all_regions(afi_id: str, starting_region: Optional[str] = None) -> None:
     """ Copies an AFI to all regions, excluding the specified region.
     starting_region=None makes starting region = to the default region. """
     starting_region = starting_region if starting_region is not None else get_current_region()
@@ -71,7 +73,7 @@ def copy_afi_to_all_regions(afi_id, starting_region=None):
         rootLogger.info("Copy result: " + str(result['FpgaImageId']))
 
 
-def share_afi_with_users(afi_id, region, useridlist):
+def share_afi_with_users(afi_id: str, region: str, useridlist: List[str]) -> None:
     """ share the AFI in Region region with users in userlist. """
     client = boto3.client('ec2', region_name=region)
     if "public" in useridlist:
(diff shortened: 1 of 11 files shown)


deploy/firesim
```diff
@@ -358,7 +358,7 @@ def shareagfi(build_config_file: BuildConfigFile) -> None:
     # share the image with all users
     for agfiname in build_config_file.agfistoshare:
         agfi = build_config_file.hwdb.get_runtimehwconfig_from_name(agfiname).agfi
-
+        assert agfi is not None
```
davidbiancolin, line 361: Provide a message here

deploy/firesim
```diff
@@ -257,7 +257,7 @@ def buildbitstream(build_config_file: BuildConfigFile) -> None:
         execute(build_config.bitbuilder.replace_rtl, hosts=['localhost'])
         execute(build_config.bitbuilder.build_driver, hosts=['localhost'])
 
-    def release_build_hosts_handler(sig, frame) -> None:
+    def release_build_hosts_handler(sig: Any, frame: Any) -> None:
```
davidbiancolin, line 260: I'm assuming there's no way to provide a more specific type annotation here?

davidbiancolin (approved):
This seems reasonable enough to me assuming all the cases of "Any" really cannot be made more specific.

deploy/firesim
abejgonzalez: Since there are alot of `Any`s left throughout the codebase, I figured this could just be done in a later PR / cleanup pass.

deploy/awstools/afitools.py
```diff
@@ -92,23 +94,23 @@ def share_afi_with_users(afi_id, region, useridlist):
...
         afi_id = get_afi_for_agfi(agfi_id, region)
         share_afi_with_users(afi_id, region, useridlist)
 
-def firesim_tags_to_description(buildtriplet, deploytriplet, commit):
+def firesim_tags_to_description(buildtriplet: str, deploytriplet: str, commit: str) -> str:
     """ Serialize the tags we want to set for storage in the AGFI description """
     return """firesim-buildtriplet:{},firesim-deploytriplet:{},firesim-commit:{}""".format(buildtriplet,deploytriplet,commit)
 
-def firesim_description_to_tags(description):
+def firesim_description_to_tags(description: str) -> Dict[Any, Any]:
```
timsnyder-siv, line 113: You probably mean `Dict[Hashable, Any]`.  I wonder if mypy would catch it for you if you tried to key the returndict with something that isn't hashable or if that will fall through to being a runtime error?

deploy/awstools/awstools.py
```diff
@@ -48,7 +48,7 @@ def __init__(self) -> None:
         MockBoto3Instance.base_ip += 1
         self.private_ip_address = ".".join([str((self.ip_addr_int >> (8*x)) & 0xFF) for x in [3, 2, 1, 0]])
 
-def depaginated_boto_query(client, operation, operation_params, return_key):
+def depaginated_boto_query(client: Any, operation: str, operation_params: Dict[Any, Any], return_key: str) -> List[Any]:
```
timsnyder-siv, line 51: Again, you probably want to use the generic `Mapping` for `operation_params` unless you spefically need a `Dict`.  Looking at the code, we only need `Mapping` because we are only unpacking it as kwargs to `paginate`

deploy/firesim
timsnyder-siv:
```suggestion
    def release_build_hosts_handler(sig: signal.Signals, frame: types.FrameType) -> None:
```
should work.

deploy/firesim
timsnyder-siv: Consider refactoring into a `no_longer_optional(agfi)` helper utility if mypy is smart enough.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -486,7 +486,7 @@ def run_workload_passes(self, use_mock_instances_for_testing: bool) -> None:
         self.boot_simulation_passes(False, skip_instance_binding=True)
 
         @parallel
-        def monitor_jobs_wrapper(run_farm, completed_jobs: List[str], teardown: bool, terminateoncompletion: bool, job_results_dir: str) -> Dict[str, Dict[str, bool]]:
+        def monitor_jobs_wrapper(run_farm: RunFarm, completed_jobs: List[str], teardown: bool, terminateoncompletion: bool, job_results_dir: str) -> Dict[str, Dict[str, bool]]:
```
timsnyder-siv, line 489: Consider using a more generic type for `completed_jobs`.

timsnyder-siv (commented):
Could you capture something about the motivation for making the type checking more strict in the PR description?  I'm assuming it is something like. "Require type checking in all manager code so that new code isn't written without including type hints" but perhaps I'm guessing incorrectly.  If you aren't worried about new code being introduced without type hints, then we should slow down and clean up more of the `Any` usage before turning this on.  Otherwise, it seems like you're sending a signal to others that it's okay to just annotate everything with `Any` and mypy starts to become something that adds verbosity but no useful meaning to the code.

This should probably be `changelog: changed` and make the tile more specific about python and that it is limited to the manager.

PR #1137, opened 2022-07-17
Fix mount files ownership

### LINKED ISSUE #1116: Copying files out of rootfs  will fail for sim_uid != real_uid
https://github.com/firesim/firesim/blob/29cd553986419772f1b8fad67650957b7f1450bb/deploy/runtools/firesim_topology_elements.py#L383

code above switches to making `sudo` optional for `chmod` calls.  Running `chmod` as non-root will fail unless the `chmod` doesn't really have to change the ownership of any files.  

I'm concerned that the `chmod` on Line 383 mentioned above will fail in many cases unless we modify the guestmount on https://github.com/firesim/firesim/blob/29cd553986419772f1b8fad67650957b7f1450bb/deploy/runtools/firesim_topology_elements.py#L346 to also use `-o uid=$(id -u) -o gid=$(id -g)` when mounting to map all uids and gids in the guest filesystem to the current user [as described in it's manpage.](https://libguestfs.org/guestmount.1.html#:~:text=extra%20memory%20usage.-,%2Do%20uid%3DN%20%2Do%20gid%3DN,-Use%20these%20options).

And then we should be able to condition the running of the recursive chown or move it into the block where the real mount is done using sudo.

I looked at real `mount` and the uid and gid options are dependent on the type of filesystem being mounted.

abejgonzalez:
Fixes #1116 

#### Related PRs / Issues

#1116

### Contributor Checklist

### DIFF
--- a/deploy/runtools/firesim_topology_elements.py
+++ b/deploy/runtools/firesim_topology_elements.py
@@ -342,8 +342,9 @@ class FireSimServerNode(FireSimNode):
         def mount(img: str, mnt: str, tmp_dir: str) -> None:
             if sudo:
                 run(f"sudo mount -o loop {img} {mnt}")
+                run(f"sudo chown -R $(whoami) {mnt}")
             else:
-                run(f"""screen -S guestmount-wait -dm bash -c "guestmount --pid-file {tmp_dir}/guestmount.pid -a {img} -m /dev/sda {mnt}; while true; do sleep 1; done;" """, pty=False)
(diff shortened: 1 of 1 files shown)


timsnyder-siv (approved):
LGTM

I reminded myself that `shell=True` is the default for `fabric.operations.run()` at https://docs.fabfile.org/en/1.14/api/core/operations.html#fabric.operations.run so the embedded `$(command substitution)` should work.

PR #1138, opened 2022-07-18
Pass Vitis FPGA platform from yaml

abejgonzalez:
Small PR that does the following:

- Cleans up the `build-bitstream.sh` scripts to parse arguments with proper errors. 
- Adds `device` to the `vitis.yaml` bit builder so that users can support different FPGAs than U250's (IIRC we only support U250 + U280s). Since I don't have a U280, this is untested. 

A couple of things to discuss:

- Do we want to add a `device` K/V pair to the HWDB to tell users what XCLBIN corresponds to what platform? I would imagine that the XCLBIN tells the FPGA platform (i.e. `device`) string if you "describe" it with `xclbinutil`.

This PR should help set up passing other args to the bitstream builds from the Python/manager side.

### Contributor Checklist

### DIFF
--- a/platforms/f1/build-bitstream.sh
+++ b/platforms/f1/build-bitstream.sh
@@ -1,9 +1,44 @@
 #!/bin/bash
 
-set -e
+# This script is called by FireSim's bitbuilder to create a xclbin
 
-CL_DIR=$1
-shift # get rid of $1 arg or else hdk_setup.sh will fail
+# exit script if any command fails
+set -e
+set -o pipefail
+
+usage() {
+    echo "usage: ${0} [OPTIONS]"
+    echo ""
+    echo "Options"
+    echo "   --cl_dir : Custom logic directory to build AWS F1 bitstream from"
+    echo "   --help   : Display this message"
+    exit "$1"
+}
+
+CL_DIR=""
+
+# getopts does not support long options, and is inflexible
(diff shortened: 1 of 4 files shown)


platforms/f1/build-bitstream.sh
```diff
@@ -1,9 +1,44 @@
 #!/bin/bash
```
davidbiancolin, line 1:
```suggestion
#!/usr/bin/env bash
```
Unless you have good reason for it to be `/bin/bash`

davidbiancolin (commented):
Mostly thinking out loud here, and i know i suggested having these scripts be the interface, but if the scripts are going to take different argument types (i.e., they are going to vary substantially from platform to platform) _maybe_ we should just push the complexity into the python and remove the build bitstream scripts. On the other hand, the bash is a little easier to read and more discoverable atm.

abejgonzalez:
> Mostly thinking out loud here, and i know i suggested having these scripts be the interface, but if the scripts are going to take different argument types (i.e., they are going to vary substantially from platform to platform) _maybe_ we should just push the complexity into the python and remove the build bitstream scripts. On the other hand, the bash is a little easier to read and more discoverable atm.

My impression was that we wanted to have the bash script so that users could call the script directly when they want to build a bitstream without the manager.

PR #1139, opened 2022-07-18
Enabled clang-format on C++ sources

nandor:
Ran clang-format on sources committed to the repository and added a CI job to ensure sources are properly formatted on each commit.
Clang-format can be installed using Linux package managers and the `git clang-format` command can be used to format all files that change with a commit.

#### Related PRs / Issues

N/A

#### UI / API Impact

No impact

No impact

### Contributor Checklist

### DIFF
--- a/.github/workflows/firesim-run-tests.yml
+++ b/.github/workflows/firesim-run-tests.yml
@@ -269,3 +269,38 @@ jobs:
           python3 -m sphinx.ext.intersphinx docs/_build/html/objects.inv
           cat /tmp/sphinx-err*.log
       - uses: ./.github/actions/job-end
+
+  clang-format:
+    name: clang-format
+    runs-on: ubuntu-20.04
+    steps:
+      # Clone the repository (shallow to save time).
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+          submodules: false
+
+      # Install clang-format
+      - name: Install clang-format
+        run: |
+          sudo apt update
+          sudo apt install clang-format -y
+
+      # Run 'clang-format', comparing against the base commit hash.
+      # If anything got reformatted, fail and output a patch.
+      - name: Run clang-format
+        run: |
+          git fetch --recurse-submodules=no origin ${{ github.base_ref }}
+          DIFF_COMMIT=$(git rev-parse origin/${{ github.base_ref }})
+          git clang-format $DIFF_COMMIT
+          git diff --ignore-submodules > clang-format.patch
+          if [ -s clang-format.patch ]; then
+            echo "error: clang-format had to fix the following files:"
+            git diff --ignore-submodules --name-only
+            echo "----- 8< ---- PATCH ----- 8< ----"
+            cat clang-format.patch
+            echo "----- 8< ---- PATCH ----- 8< ----"
+            git checkout .
+            exit 1
+          fi
--- a/docs/Developer-Docs/GoldenGate-and-Driver-Development.rst
+++ b/docs/Developer-Docs/GoldenGate-and-Driver-Development.rst
@@ -160,7 +160,6 @@ C/C++ guidelines
 
 The C++ sources are formatted using ``clang-format`` and all submitted pull-requests
 must be formatted prior to being accepted and merged. The sources follow the coding
-style defined `here <https://github.com/firesim/firesim/blob/main/.clang_format>`_, based
-on the `LLVM <https://llvm.org/docs/CodingStandards.html>`_ guidelines.
+style defined `here <https://github.com/firesim/firesim/blob/main/.clang_format>`_.
 
 ``git clang-format`` can be used before committing to ensure that files are properly formatted.
--- a/docs/Developer-Docs/GoldenGate-and-Driver-Development.rst
+++ b/docs/Developer-Docs/GoldenGate-and-Driver-Development.rst
@@ -154,3 +154,13 @@ Defining A New Test
 Extend the appropriate ScalaTest spec or base class, and
 place the file under the correct ``src/test/scala`` directory. They will be
 automatically enumerated by ScalaTest and will run in CI by default.
+
+C/C++ guidelines
+++++++++++++++++
+
+The C++ sources are formatted using ``clang-format`` and all submitted pull-requests
+must be formatted prior to being accepted and merged. The sources follow the coding
+style defined `here <https://github.com/firesim/firesim/blob/main/.clang_format>`_, based
+on the `LLVM <https://llvm.org/docs/CodingStandards.html>`_ guidelines.
+
+``git clang-format`` can be used before committing to ensure that files are properly formatted.
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -282,7 +282,7 @@ set -o pipefail
         screen \
     )
     # clang-format for driver coding style enforcement.
-    CONDA_PACKAGE_SPECS+=( clang-format )
+    CONDA_PACKAGE_SPECS+=( clang-format clang-tools )
 
     # python packages
     # While it is possible to install using pip after creating the
--- a/scripts/machine-launch-script.sh
+++ b/scripts/machine-launch-script.sh
@@ -281,6 +281,8 @@ set -o pipefail
         verilator==4.034 \
         screen \
     )
+    # clang-format for driver coding style enforcement.
+    CONDA_PACKAGE_SPECS+=( clang-format )
 
     # python packages
     # While it is possible to install using pip after creating the
--- a/sim/firesim-lib/src/main/cc/bridges/blockdev.h
+++ b/sim/firesim-lib/src/main/cc/bridges/blockdev.h
@@ -1,10 +1,10 @@
-//See LICENSE for license details
+// See LICENSE for license details
 #ifndef __BLOCKDEV_H
 #define __BLOCKDEV_H
 
-#include <vector>
 #include <queue>
 #include <stdio.h>
+#include <vector>
 
 #include "bridges/bridge_driver.h"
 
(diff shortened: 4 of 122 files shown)


.github/workflows/firesim-run-tests.yml
```diff
@@ -269,3 +269,44 @@ jobs:
           python3 -m sphinx.ext.intersphinx docs/_build/html/objects.inv
           cat /tmp/sphinx-err*.log
       - uses: ./.github/actions/job-end
+
+  lints:
+    name: Lints
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/circt/images/circt-ci-build:20210929220136
```
timsnyder-siv, line 277: I don't think we want to depend on another unrelated project's CI container to run clang-format and not provide the same version to developers to run locally via conda.

If we are going to adopt `clang-format`, we should better integrate this into our existing development environment before merging.

We should also name the step something more descriptive than `lints` unless we're thinking of cramming all of the lint checkers into this one step.  `clang-check` or `cppcheck` are more what I would consider lint checkers, this is just enforcing a formatting standard.

timsnyder-siv (changes requested):
In general, I'm all for having a well-defined C++ formatting standard and I don't have any issues with `clang-format`.  If we are going to adopt it for C++ we should add it to the conda-dependencies and document this in the dev docs before merging this.

.github/workflows/firesim-run-tests.yml
nandor: I changed this to use a default image. Will look into conda as well.

nandor:
I have added a small note in the driver development notes about coding style and amended the conda package list to include clang-format.

.github/workflows/firesim-run-tests.yml
```diff
@@ -269,3 +269,48 @@ jobs:
...
+      # is triggered on both pushes and pull requests, we have to determine if
+      # the pull request target branch is set (which it will only be on the PR
+      # triggered flow). If it's not, then compare against the last commit.
+      # Since we did a shallow fetch for this repo, we must fetch the commit
+      # upon which we be diffing
+      - name: Run clang-format
+        run: |
+          if [ -z "${{ github.base_ref }}" ]; then
+            DIFF_COMMIT=$(git rev-parse HEAD^)
+          else
```
davidbiancolin, line 302: So i'm a little confused here -- this workflow should only be run on open, non-draft PRs against main (and 1.[0-9].x), why _isn't_ the `base_ref` always set?

nandor: I've removed this check. The rule was adapted from another CI job which did not run only on pull events that expose base_ref. However, if you think the triggers of this job will change, we should revert.

sim/midas/src/main/cc/bridges/synthesized_prints.h
```diff
@@ -3,123 +3,114 @@
...
+public:
+  synthesized_prints_t(
+      simif_t *sim, std::vector<std::string> &args,
+      PRINTBRIDGEMODULE_struct *mmio_addrs, unsigned int print_count,
+      unsigned int token_bytes, unsigned int idle_cycles_mask,
+      const unsigned int *print_offsets, const char *const *format_strings,
+      const unsigned int *argument_counts, const unsigned int *argument_widths,
+      unsigned int stream_idx, unsigned int stream_depth,
+      const char *const clock_domain_name, const unsigned int clock_multiplier,
+      const unsigned int clock_divisor, int printno);
```
timsnyder-siv, line 53: I'll be the first to admit that the signature of this constructor is comically long but `clang-format` has made this almost unreadable by smashing collapsing it.

timsnyder-siv (commented):
I've taken a more serious look at the actual output of `clang-format` and have some thoughts for discussion.  In general, I'm really happy to see `clang-format` being used because it is going to make the code more consistent. Having a clearly defined format that is expected in the repo makes it possible to reformat it to anyone's individual liking while editing (and then reapply the official formatting).  Can be done either with git's smudge filters or with editor support (each has it's plusses and minuses).

fabianschuiki:
This is very nice! I love the idea of establishing _any_ `clang-format` style guidelines for the C++ here. I don't have any strong opinions on the individual style options. My suggestion would be to settle on _any_ initial formatting setup, and then iterate on the formatting details afterwards with subsequent PRs -- after all it just takes a run of clang-format across all sources in this repo to update to a new style.

sim/firesim-lib/src/main/cc/bridges/blockdev.h
```diff
@@ -14,76 +14,78 @@
...
 
 struct blkdev_request {
-    bool write;
-    uint32_t offset;
-    uint32_t len;
-    uint32_t tag;
+  bool write;
+  uint32_t offset;
+  uint32_t len;
+  uint32_t tag;
```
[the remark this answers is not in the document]
nandor: I strongly disagree with per-project policies instead of language-specific policies. For C++, 2 spaces and 80 characters are quite common and enforced in other related projects. I find Scala very hard to read since it exceeds 100 characters per line (which I think the limit might be in FireSim). Keeping indentation to 2 spaces and line length to 80 characters results, in my opinion, in a better overview of the code.

sim/midas/src/main/cc/bridges/synthesized_prints.h
nandor: I think we can adjust this in clang-format.

sim/firesim-lib/src/main/cc/bridges/dromajo.h
```diff
@@ -1,84 +1,74 @@
...
+  simif_t *_sim;
 
-        int invoke_dromajo(uint8_t* buf);
-        int beats_available_stable();
-        size_t process_tokens(int num_beats, size_t minimum_batch_beats);
-        void flush();
+  int invoke_dromajo(uint8_t *buf);
+  int beats_available_stable();
+  size_t process_tokens(int num_beats, size_t minimum_batch_beats);
+  void flush();
```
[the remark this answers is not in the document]
nandor: I'll keep this in mind if I have to change code involving trailing underscores in the future.

scripts/machine-launch-script.sh
```diff
@@ -281,6 +281,8 @@ set -o pipefail
         verilator==4.034 \
         screen \
     )
+    # clang-format for driver coding style enforcement.
+    CONDA_PACKAGE_SPECS+=( clang-format )
```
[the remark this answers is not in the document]
fabianschuiki: :+1: That sounds great!

sim/firesim-lib/src/main/cc/bridges/dromajo.cc
```diff
@@ -129,169 +121,190 @@ dromajo_t::~dromajo_t() {
...
 
-        // if int/excp was found in this set of commit traces... reset on next set of commit traces
-        if (this->_trace_idx == 0){
-            this->saw_int_excp = false;
-        }
+    // if int/excp was found in this set of commit traces... reset on next set
+    // of commit traces
+    if (this->_trace_idx == 0) {
+      this->saw_int_excp = false;
+    }
```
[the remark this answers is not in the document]
fabianschuiki: Totally agree. I've heard Dennis Ritchie being quoted as saying that the worst mistake he made in his entire life was making braces optional in C :smile:

sim/firesim-lib/src/main/cc/bridges/dromajo.cc
```diff
@@ -16,110 +16,102 @@
...
-            d_bin = const_cast<char*>(arg.c_str()) + dromajo_bin_arg.length();
-            this->dromajo_bin = std::string(d_bin);
-            bin_found = true;
-        }
+dromajo_t::dromajo_t(simif_t *sim, std::vector<std::string> &args,
+                     int iaddr_width, int insn_width, int wdata_width,
+                     int cause_width, int tval_width, int num_traces,
+                     DROMAJOBRIDGEMODULE_struct *mmio_addrs, int stream_idx,
+                     int stream_depth)
+    : bridge_driver_t(sim), stream_idx(stream_idx), stream_depth(stream_depth) {
```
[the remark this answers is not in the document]
nandor: I'd discourage multiple pointer declarations on a single line. In general, if a pointer declaration is not trivial, typedefs with comments should be inserted to clarify their meaning and then the location of the pointer or the reference is irrelevant. 

Regarding the Chromium style, I would very much prefer sticking to the LLVM style.

sim/firesim-lib/src/main/cc/bridges/dromajo.h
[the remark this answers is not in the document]
fabianschuiki: Should we spin this off into a dedicated tracking issue for making names more consistent? I like your suggestion of keeping this PR minimal and to the point of just adding consistent automated formatting.

sim/firesim-lib/src/main/cc/bridges/blockdev.h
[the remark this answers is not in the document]
fabianschuiki: My suggestion here would be to not necessarily condition one on the other. C++ can have a significant rightward drift if lambdas are used excessively, which would suggest 2 spaces of indentation. But I could also see other languages having 4 spaces as the idiomatic indentation, which is totally fine. And you might have languages like Make that rely on tabs, so unifying this across language boundaries seems like it's more trouble than benefit. We should definitely stay consistent within languages though.

sim/midas/src/main/cc/bridges/synthesized_prints.h
fabianschuiki: We could add
```
BinPackArguments: false
BinPackParameters: false
```
to ensure func args are either all on one line, or all on their own line.

sim/firesim-lib/src/main/cc/bridges/dromajo.cc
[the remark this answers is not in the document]
fabianschuiki: Totally agree with @timsnyder-siv here: we should probably just go with the `PointerAlignment: right` that the LLVM style uses by default. No need to switch to Chromium style; there's two simple bin packing options we can add to get the func args right.

sim/midas/src/main/cc/bridges/synthesized_prints.h
nandor: I tried these options and the output doesn't look great. We can either roll with what we have now or I can try to find some other options, although that would introduce further delays since I don't quite have time for that now.

sim/firesim-lib/src/main/cc/bridges/blockdev.h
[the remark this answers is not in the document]
timsnyder-siv: @davidbiancolin confirmed that the scala is using indent of 2 spaces.  so the indenting comment I made here is moot.  I was primarily concerned about Scala and C++ because it is most likely that one would be looking at both those at the same time in this project.  I should have been more precise and limited in my comment.

I'm fine with shifting it to 2 spaces.  I used to like 2 spaces more 20 years ago. As my eyes have gotten older, I like it less. I should probably suck it up and increase the font size on my display again or go to the eye doctor. 🤣 

Excessive use of lambdas should probably be avoided, as should signatures that take a page of code to represent without binpacking the arguments (or 10 lines with them packed...).

David also mentioned that the Scala column limit is currently 120.  That seems excessive to me but the coming adoption of scalafmt will make it easier to reflow that on the fly.

sim/midas/src/main/cc/bridges/synthesized_prints.h
timsnyder-siv: @nandor can you be more specific about what didn't look great?  When I scanned through the output with those options enabled, it looked like the output was only changing the arguments to be either all on one line or all on their own line.

davidbiancolin:
For reference, the scalaFmt is here: https://github.com/firesim/firesim/blob/main/sim/.scalafmt.conf#L210

davidbiancolin:
Really the only two things i'm willing to bike shed about are [the classic] column limit (maybe 100 instead of 80?), and parameter list formatting (see above). 

It's worth noting though, that most of the instances of long parameter + initializer lists are in bridge definitions will disappear once we make some of the driver modifications we've proposed, since these parameters will be captured in automatically constructed instances which will be passed to constructors their stead.

davidbiancolin:
Finally, I am a little reluctant to fiddle with the formatting if the pertubatios are going to introduce +/- 1000 line PRs that i'll have to deal with in our fork. I'd prefer to just pick something and run with it.

sim/midas/src/main/cc/bridges/synthesized_prints.h
nandor:
```
fprintf(stderr,
          "Some format %x\n",
          number);
```
is one example

nandor: We could bin-pack arguments to function calls, placing parameters on new lines.

timsnyder-siv: Thanks for providing the detail.

nandor:
Do you have a decision on parameter formatting and column widths?

scripts/machine-launch-script.sh
timsnyder-siv, line 285: Did some more looking and found it.  They put it in the `clang-tools` package.  I'll ask about splitting it out into the clang-format package so that we don't have to install the rest of clang but in the meantime, we could:
```suggestion
    CONDA_PACKAGE_SPECS+=( clang-format clang-tools )
```

davidbiancolin:
My final call is to add:
```
BinPackArguments: false
BinPackParameters: false
```
We leave the column limit at 80. We can circle back on the BinPackingX when we've cleaned up the header situation.

docs/Developer-Docs/GoldenGate-and-Driver-Development.rst
```diff
@@ -154,3 +154,10 @@ Defining A New Test
 Extend the appropriate ScalaTest spec or base class, and
 place the file under the correct ``src/test/scala`` directory. They will be
 automatically enumerated by ScalaTest and will run in CI by default.
+
+C/C++ guidelines
+++++++++++++++++
+
+The C++ sources follow the LLVM style guide and should be formatted accordingly
+using `clang-format`. `git clang-format` can be used before committing to ensure
+that files are properly formatted. Formatting is enforced through a CI job.
```
timsnyder-siv, line 163:
```suggestion
The C++ sources are formatted using ``clang-format`` and all submitted pull-requests 
must be formatted following `https://github.com/firesim/firesim/blob/main/.clang_format`_ 
prior to being accepted and merged.

``git clang-format`` can be used before committing to ensure that files are properly formatted. 
```

RST inline literals need to be surrounded in double backtics.

Should explain that we're using `clang-format` in a way that doesn't imply that we've adopted something called the "LLVM style guide" that lives somewhere like https://llvm.org/docs/CodingStandards.html or something else that a search engine might confuse someone with.

timsnyder-siv (approved):

davidbiancolin:
Im running out of time to play with this tonight, but i'll get to it tomorrow.

davidbiancolin:
I've decided that the argument packing to `fprintf` isn't triggering enough for me to go and add waivers to metasimulation harnesses. We should stop using `fprintf` in favor of C++ I/O anyways.

PR #1140, opened 2022-07-18
add xclbin URI support

timsnyder-siv:
use fsspec to enable xclbin's to be one of any URI protocol
supported by the library or an installed add-on

#### UI / API Impact

Enables xclbin in hwdb to be a URI supported by [fsspec](https://filesystem-spec.readthedocs.io/en/latest/index.html)

No impact to Verilog or AGFI.

### Contributor Checklist
- [x] Update the docs with 
  - list of protocols tested and known to work
  - pointer to fsspec config that can be used to inject auth stuff (TO ALL CALLS).  Since the auth args go to all calls when using the fsspec level config, it won't work in all cases but it might be useful as we start to use things and determine if/how we want to add features to the manager to specify auth.  In the limit, you might need different auth for every file you interact with but you don't want to have to repeat and maintain auth everywhere.  Just brainstorming, it seems like fsspec should allow it's config to do URI pattern matching to provide auth.
- [X] Add a test for `ssh://` in #1167 


deploy/runtools/runtime_config.py
```diff
@@ -40,7 +40,10 @@ class RuntimeHWConfig:
 
     # TODO: should be abstracted out between platforms with a URI
     agfi: Optional[str]
+    """User-specified, possibly-URI, path to xclbin"""
```
abejgonzalez, line 43: IMO we should just enforce that this is a URI (i.e. enforce `file:///path/to/file`)

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -499,11 +499,29 @@ key/value in the same recipe).
...
-This represents a path to a bitstream (FPGA Image) used by this hardware configuration.
-This path must be local to the run farm host that the simulation runs on.
-Only used in Vitis FireSim configurations (an ``agfi`` key/value cannot exist with this
-key/value in the same recipe)
-
+Indicates where the bitstream (FPGA Image) is located, may be one of:
+  * A Uniform Resource Identifier (URI) which specifies a protocol supported either `directly
+    by the fsspec library <https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations>`_ or
+    by `one of the many third party extension libraries which build on fsspec. <https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations>`_
+  * A filesystem path available to the run farm host.
```
abejgonzalez, line 506: Ditto.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -10,6 +10,9 @@
 from fabric.contrib.project import rsync_project # type: ignore
 import time
 from os.path import join as pjoin
+from os import PathLike, fspath
+from fsspec.core import url_to_fs
```
abejgonzalez, line 14: Shouldn't we add this to the `machine-launch-script.sh` even if it is pulled in from something else?

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+        """ download xclbin URI to remote node. """
+        assert slotno < len(self.parent_node.sim_slots)
+        serv = self.parent_node.sim_slots[slotno]
+        hwcfg = serv.get_resolved_server_hardware_config()
+        if re.match(_RFC_3986_PATTERN, hwcfg.xclbin):
+            remote_home_dir = self.parent_node.get_sim_dir()
+            remote_sim_dir = """{}/sim_slot_{}/""".format(remote_home_dir, slotno)
+            hwcfg.local_xclbin = './local.xclbin'
+
+            def get_xclbin(xclbin_uri:str, xclbin_lpath:PathLike) -> None:
```
abejgonzalez, line 674:
```suggestion
            def get_xclbin(xclbin_uri: str, xclbin_lpath: PathLike) -> None:
```

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+                # TODO instead of blowing up if the file exists, consider using fsspec
+                # filecache https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally
+                # so that multiple slots using the same xclbin only grab it once and
+                # we only download it if it has changed at the source.
+                # HOWEVER, 'filecache' isn't thread/process safe and I'm not sure whether
+                # this runs in @parallel for fabric
```
abejgonzalez, line 680: This code should run in parallel on multiple run hosts but once you are using a specific run host this is done serially: 

https://github.com/firesim/firesim/blob/7839d4fda5eb360841694833c40f50e3e9bf3387/deploy/runtools/firesim_topology_with_passes.py#L388-L393

https://github.com/firesim/firesim/blob/7839d4fda5eb360841694833c40f50e3e9bf3387/deploy/runtools/run_farm_deploy_managers.py#L598

abejgonzalez: I think using the file cache should be fine (and doesn't need a lock)

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+                lpath = Path(xclbin_lpath)
+                assert not lpath.exists, f"{lpath.resolve(strict=False)} already exists, refusing to overwrite"
+                rootLogger.debug("Downloading '%s' to '%s'", xclbin_uri, )
```
abejgonzalez, line 683:
```suggestion
                rootLogger.debug(f"Downloading {xclbin_uri} to {lpath}")
```

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+            hwcfg.local_xclbin = './local.xclbin'
+
+            def get_xclbin(xclbin_uri:str, xclbin_lpath:PathLike) -> None:
+                # TODO instead of blowing up if the file exists, consider using fsspec
+                # filecache https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally
+                # so that multiple slots using the same xclbin only grab it once and
+                # we only download it if it has changed at the source.
+                # HOWEVER, 'filecache' isn't thread/process safe and I'm not sure whether
+                # this runs in @parallel for fabric
+                lpath = Path(xclbin_lpath)
```
abejgonzalez, line 681: I hate to say this... but we should probably port to using `pathlib` everywhere. (just a comment)

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+                assert not lpath.exists, f"{lpath.resolve(strict=False)} already exists, refusing to overwrite"
+                rootLogger.debug("Downloading '%s' to '%s'", xclbin_uri, )
+                fs, rpath = url_to_fs(xclbin_uri)
+                fs.get_file(rpath, fspath(lpath)) # fsspec deals in strings, not PathLike
```
abejgonzalez, line 685: Whats the debug output if you can't obtain a file? Does this get caught by the `StreamLogger` (I forget how `StreamLogger` works exactly)

abejgonzalez (changes requested):
Seems pretty good to me so far minus some small nits.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Oh, yeah, if I didn't add fsspec to machine-launch-script, that was an oversight.  I'll add it.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Ok.  It's good to know that it should work now.  What happens if someone changes the code so that a lock is needed?  How do we detect that case and blow up?

Said differently, can we test to ensure that it should continue to be okay?

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Good question.  I think it throws an exception and I would suspect that the debug output quality varies based on which filesystem implementation is being used.  When I tried to `fs.ls()` a non-existent file (outside of FireSim and fabric), a stacktrace looked like :
```
FileNotFoundError                         Traceback (most recent call last)
/home/centos/orca_pr_220621/fsspec.ipynb Cell 13 in <cell line: 1>()
----> [1](vscode-notebook-cell://ssh-remote%2Bfs5/home/centos/orca_pr_220621/fsspec.ipynb#ch0000012vscode-remote?line=0) fs[0].ls(fs[1])

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/fsspec/implementations/sftp.py:115, in SFTPFileSystem.ls(self, path, detail)
    113 def ls(self, path, detail=False):
    114     logger.debug("Listing folder %s" % path)
--> 115     stats = [self._decode_stat(stat, path) for stat in self.ftp.listdir_iter(path)]
    116     if detail:
    117         return stats

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/fsspec/implementations/sftp.py:115, in <listcomp>(.0)
    113 def ls(self, path, detail=False):
    114     logger.debug("Listing folder %s" % path)
--> 115     stats = [self._decode_stat(stat, path) for stat in self.ftp.listdir_iter(path)]
    116     if detail:
    117         return stats

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/paramiko/sftp_client.py:278, in SFTPClient.listdir_iter(self, path, read_aheads)
    276 path = self._adjust_cwd(path)
    277 self._log(DEBUG, "listdir({!r})".format(path))
--> 278 t, msg = self._request(CMD_OPENDIR, path)
    280 if t != CMD_HANDLE:
    281     raise SFTPError("Expected handle")

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/paramiko/sftp_client.py:822, in SFTPClient._request(self, t, *arg)
    820 def _request(self, t, *arg):
    821     num = self._async_request(type(None), t, *arg)
--> 822     return self._read_response(num)

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/paramiko/sftp_client.py:874, in SFTPClient._read_response(self, waitfor)
    871 if num == waitfor:
    872     # synchronous
    873     if t == CMD_STATUS:
--> 874         self._convert_status(msg)
    875     return t, msg
    877 # can not rewrite this to deal with E721, either as a None check
    878 # nor as not an instance of None or NoneType

File ~/miniforge3/envs/jlab/lib/python3.10/site-packages/paramiko/sftp_client.py:903, in SFTPClient._convert_status(self, msg)
    900     raise EOFError(text)
    901 elif code == SFTP_NO_SUCH_FILE:
    902     # clever idea from john a. meinel: map the error codes to errno
--> 903     raise IOError(errno.ENOENT, text)
    904 elif code == SFTP_PERMISSION_DENIED:
    905     raise IOError(errno.EACCES, text)

FileNotFoundError: [Errno 2] No such file
```

I suspect that we will want to put this into  a `try/except` block that makes the debug output nicer than that big stacktrace.

deploy/runtools/runtime_config.py
timsnyder-siv: Probably.  The only thing we would be losing is the ability for someone to put a relative path in their hwdb (because file URIs need the full path).

While on this topic, I wonder how fsspec would handle `file://<ipaddress>/path/to/file` because that is a valid file URI.  https://en.wikipedia.org/wiki/File_URI_scheme. I had forgotten that you can give a hostname in a file URI because I never do.

abejgonzalez: I don't think that having a relative path is that important. I think everything using a URI would be clearer to users.

As for having a hostname in a URI, I assume it probably just does a `rsync` like operation to retrieve the file. Whats your concern?

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: I vote to just have a comment indicating that the code should not be run in parallel on the same host. I think because F1 requires that we do things serially, we will just enforce this in the future.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
                 with StreamLogger('stdout'), StreamLogger('stderr'):
                     run(f"xbutil reset -d {card_bdf} --force")
 
+    def localize_xclbin(self, slotno: int) -> None:
+        """ download xclbin URI to remote node. """
+        assert slotno < len(self.parent_node.sim_slots)
```
davidbiancolin, line 666: If you could provide a helpful assertion message, that might be useful.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
+
+            def get_xclbin(xclbin_uri:str, xclbin_lpath:PathLike) -> None:
+                # TODO instead of blowing up if the file exists, consider using fsspec
+                # filecache https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally
+                # so that multiple slots using the same xclbin only grab it once and
+                # we only download it if it has changed at the source.
+                # HOWEVER, 'filecache' isn't thread/process safe and I'm not sure whether
+                # this runs in @parallel for fabric
+                lpath = Path(xclbin_lpath)
+                assert not lpath.exists, f"{lpath.resolve(strict=False)} already exists, refusing to overwrite"
```
davidbiancolin, line 682: This is going to fail often since people rerun infrasetup before runworkload right? For all other files, the current behavior of infrasetup is to silently clobber stuff that exists so we should probably do the same here no?

However, we could have the manager backup the directory on re-invocations of infrasetup or ensure the directory does not exist, because the behavior here is probably more desirable.

deploy/runtools/runtime_config.py
davidbiancolin: authentication will be annoying depending on how it's implemented under the hood. 

I personally would like relative paths to be supported.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -24,6 +27,8 @@
 
 rootLogger = logging.getLogger()
 
+_RFC_3986_PATTERN = re.compile(r"^[A-Za-z][A-Za-z0-9+\-+.]*://")
```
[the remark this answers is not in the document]
timsnyder-siv: I swiped it from https://github.com/pandas-dev/pandas/blob/96b036cbcf7db5d3ba875aac28c4f6a678214bfb/pandas/io/common.py#L73. I'll add a comment that points there unless you were saying that you'd like a pointer to https://datatracker.ietf.org/doc/html/rfc3986

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Sure.  I'll come up with something and add it to the other methods that make this same assertion without a message.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Yeah.  You make a good point about this failing often.  The ideal thing to do would be using the FileCache so that fsspec will do a checksum and not re-download the file if you've already got it.  I'd just like to be confident that it is safe to use the FileCache (and that it remains safe).  

I really like the current behavior of infrasetup where the rsync is essentially a no-op if you need to rerun it.  I wouldn't want to change that (although I'm not sure that's what you were suggesting with your 'backup' idea).

deploy/runtools/runtime_config.py
timsnyder-siv: > Whats your concern?

That we would need to document it because users will need to know how to setup their authentication.

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: If you are concerned about this potentially breaking... I think grabbing a lock and releasing it should be fine (esp. since that wouldn't be the bottleneck).

timsnyder-siv: @abejgonzalez what kind of lock are you thinking?  `flock()`?  I typically stay away from `flock()` because it isn't implemented well in many of the NFSv3 environments I find myself having to work in.  I think I'll just change the code to overwrite the file for now and if we can revisit the FileCache in a future PR if needed.

timsnyder-siv:
I have some more commits that I want to push to this after rebasing on top of the fixed pytests (because they weren't really broken by this stuff)

deploy/runtools/runtime_config.py
abejgonzalez: If relative paths are supported, I think it would be nice to enforce that full paths are done using URIs while relative paths shouldn't (maybe `pathlib` has a way to enforce a `Path` is a relative path)? However, this isn't blocking so if that can't happen / is too much effort we can just do what currently exists.

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: I think this debug output is fine. I rather it be very verbose and descriptive without us having to do extra work to catch an exception.

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: I think this assertion should just be put into a function since this is done quite a bit.

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: SGTM. I didn't have a lock in mind.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -10,6 +10,9 @@
 from fabric.contrib.project import rsync_project # type: ignore
 import time
 from os.path import join as pjoin
+from os import PathLike, fspath
+from fsspec.core import url_to_fs # type: ignore
```
timsnyder-siv, line 14: Looks like fsspec doesn't have type annotations or hints available yet but they aren't opposed to discussing them if this becomes problematic. https://github.com/fsspec/filesystem_spec/issues/625

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv, line 30:
```suggestion
# from  https://github.com/pandas-dev/pandas/blob/96b036cbcf7db5d3ba875aac28c4f6a678214bfb/pandas/io/common.py#L73
_RFC_3986_PATTERN = re.compile(r"^[A-Za-z][A-Za-z0-9+\-+.]*://")
```

@davidbiancolin is this the kind of ref you are wanting?

deploy/runtools/runtime_config.py
timsnyder-siv: It wouldn't be hard to `raise` something if `localize_xclbin()` is given an absolute filepath.  I'm concerned that it would be better for this to be caught earlier like when the config is read and validated against a schema.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: added in https://github.com/firesim/firesim/pull/1140/commits/7e93452b96b7d2aa26d5c2ee5483303d580ccfec

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -656,6 +661,37 @@ def clear_fpgas(self) -> None:
...
                     run(f"xbutil reset -d {card_bdf} --force")
 
+    def localize_xclbin(self, slotno: int) -> None:
+        """ download xclbin URI to remote node. """
+        assert slotno < len(self.parent_node.sim_slots)
+        serv = self.parent_node.sim_slots[slotno]
+        hwcfg = serv.get_resolved_server_hardware_config()
+        if re.match(_RFC_3986_PATTERN, hwcfg.xclbin):
+            remote_home_dir = self.parent_node.get_sim_dir()
+            remote_sim_dir = """{}/sim_slot_{}/""".format(remote_home_dir, slotno)
```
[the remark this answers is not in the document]
timsnyder-siv: done. see https://github.com/firesim/firesim/pull/1140/commits/da27c75ed25b328b4cf7c68baf919f000fd37613

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: done. see https://github.com/firesim/firesim/pull/1140/commits/da27c75ed25b328b4cf7c68baf919f000fd37613

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: done. see https://github.com/firesim/firesim/pull/1140/commits/da27c75ed25b328b4cf7c68baf919f000fd37613

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: cool.  I won't worry about adding nicer error messaging for now then.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Yeah, I'm going to punt on using the FileCache for now.  I made `localize_xclbin()` overwrite with a logged message in https://github.com/firesim/firesim/pull/1140/commits/9bd9bce1d0880909a3676b967d3f967dfd368bfd

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -499,11 +499,29 @@ key/value in the same recipe).
...
+Documentation supplying credentials is provided by the individual protocol implementation.  For
+example:
+* `s3fs for AWS S3 <https://s3fs.readthedocs.io/en/latest/#credentials>`_
+* `gcfs for Google Cloud Services <https://gcsfs.readthedocs.io/en/latest/#credentials>`_
+
+For SSH, add any required keys to your ssh-agent.
+
+We currently support protocols that provide authentication via their own configuration files
+or environment variables.  In the future, we may provide a means for passing credentials that can
+only be passed to the fsspec method calls.
```
timsnyder-siv, line 524: Should tweak this to explain the fsspec config file mechanism for injecting `kwargs` for backends that require them.  Upstream doc is at https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Yeah, it might make sense.  I've started using it only because it makes `Path` methods available to tab-completion in whatever IDE I'm using and I'm too lazy to remember which functions need to be imported from `os.path` to do whatever I'm doing.

Note that there are some significant wrinkles with regards to `pathlib` and `fsspec` interoperability (see https://github.com/fsspec/filesystem_spec/issues/539).  Enough so that folks at QuantSight have developed https://github.com/fsspec/universal_pathlib as an adapter for `pathlib` usage with `fsspec`

abejgonzalez: Should we use `universal_pathlib` instead of `pathlib` everywhere then? Based on this it seems like it's a superset of functionality? https://github.com/fsspec/universal_pathlib/blob/main/notebooks/examples.ipynb.

I'm slightly concerned about future bugs when using `Path`s (or `UPath`s) with Fabric since `Path` objects are relative to localhost (i.e. `Path("/remote/path").mkdir()` will create a local directory when you meant to do this with Fabric). In any case, this discussion should happen in another PR.

timsnyder-siv: I agree that we should pull this out into another PR.  I captured #1148 for discussion.

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Or removed because letting Python raise IndexError is probably just as useful and reduces the amount of code in the manager.  I'll leave that to future cleanup PRs .

timsnyder-siv: Also, if we move it into another function, I wonder if mypy is smart enough to use the assert in it's type-checking.  I know this particular assert predates the type checking and this is more a curiosity that I'm voicing out-loud because we have quite a few mypy helper asserts that check for `None`ness of `Optional` types and I could envision asking the same thing about them.

deploy/runtools/run_farm_deploy_managers.py
davidbiancolin: Perfect.

timsnyder-siv:
I updated the checklist with the remaining tasks for this PR to be ready to merge.

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,64 @@
...
+import boto3
+
+pytest.mark.usefixtures("aws_test_credentials")
+
+
+@pytest.mark.parametrize(
+    'protocol_type,test_dest_file_path',
+    [
+        ('s3',Path("tests/s3_test_download_json.json")),
+        ('file',Path("tests/file_test_download_json.json")),
```
timsnyder-siv, line 17: Do we want to add a test for ssh?  The simplest way to do it would be to spin up a docker container that the test would try to pull things out of with ssh.  If docker isn't installed, the test would be skipped.

Alternately, we write a test for this that utilizes the multiple hosts in CI.  Or we could do both so that devs can test this locally (if they have docker installed) and we have a test in CI without having to install docker in the place where we're running pytests.

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    
+    logger_mock = mocker.patch("util.io.rootLogger", MagicMock())
+    test_file_path = Path("tests/fsspec_test_json.json")
+    
+    if protocol_type == 's3':
+        try:
+            test_bucket = "TestBucket"
+            test_bucket_key = "s3_blob.json"
+            mock_s3_client = boto3.client('s3')
+            mock_s3_client.create_bucket(Bucket="TestBucket")
```
filipstamenkovic-sifive, line 32:
```suggestion
            mock_s3_client = boto3.client('s3', region_name='us-west-2')
            mock_s3_client.create_bucket(
                Bucket="TestBucket",
                CreateBucketConfiguration={
                   'LocationConstraint': 'us-west-2',
                }
            )
```

scripts/machine-launch-script.sh
```diff
@@ -308,6 +308,7 @@ set -o pipefail
         boto3-stubs==1.21.6 \
         botocore-stubs==1.24.7 \
         mypy-boto3-s3==1.21.0 \
+        fsspec \
```
filipstamenkovic-sifive, line 311:
```suggestion
        fsspec \
        s3fs \
```

filipstamenkovic-sifive (commented):
@timsnyder-siv adding a few suggestions to fix the test issue in CI.

deploy/util/io.py
```diff
@@ -0,0 +1,29 @@
...
+    # we only download it if it has changed at the source.
+    # HOWEVER, 'filecache' isn't thread/process safe and I'm not sure whether
+    # this runs in @parallel for fabric
+    lpath = Path(local_dest_path)
+    if lpath.exists():
+        rootLogger.debug(f"Overwriting {lpath.resolve(strict=False)}")
+    rootLogger.debug(f"Downloading '{uri}' to '{lpath}'")
+    fs, rpath = url_to_fs(uri)
+    print(fs, type(fs))
+    print(rpath, type(rpath))
```
abejgonzalez, line 28:
```suggestion
```

deploy/runtools/run_farm_deploy_managers.py
abejgonzalez: Mypy is smart enough to do this. I can do quick cleanup of this when I get the chance

deploy/tests/test_utils.py
abejgonzalez: Couldn't we just test this by running `localhost` for the "it should work" case and some invalid IP for the "it shouldn't work" case? I'm not sure it's worth having docker installed for such a small test

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
timsnyder-siv: Done in https://github.com/firesim/firesim/pull/1140/commits/2268d442ee4d36e4f74f7487c5a45316d8169b5f

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: We decided to punt on FileCache for this PR.

deploy/tests/test_utils.py
timsnyder-siv: @abejgonzalez how do we differentiate whether someone doesn't have sshd configured correctly on localhost vs the test actually working?  I don't like having a dependency into sshd that the tests aren't actually controling the configuration for.

A lighter-weight option could be for a test fixture to spin up a Paramiko sshd on a non-system port or use something like [this gist](https://gist.github.com/vxgmichel/d0b7e4cc3caab32601051ee262ee7b31).   What do you think?

deploy/runtools/runtime_config.py
timsnyder-siv: @abejgonzalez, the more I think about this, the less I feel like enforcing where users use a `file://` URI vs a simple string path.  Please speak up if you feel strongly enough about this that it should block merging.

timsnyder-siv: @abejgonzalez , since you requested changes, I think I need you to approve.  I guess there are new mypy errors I need to cleanup also.

deploy/tests/test_utils.py
abejgonzalez: That seems better to me.

abejgonzalez (approved):
LGTM. Pending all CI passing.

deploy/tests/test_utils.py
timsnyder-siv: I've asked @filipstamenkovic-sifive to help me add a test.

timsnyder-siv: Captured in #1167  for followup.  I'd like to go ahead and get this merged if the CI passes this time.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -633,6 +640,25 @@ def clear_fpgas(self) -> None:
...
+        serv = self.parent_node.sim_slots[slotno]
+        hwcfg = serv.get_resolved_server_hardware_config()
+        assert hwcfg.xclbin is not None
+        if re.match(_RFC_3986_PATTERN, hwcfg.xclbin):
+            remote_home_dir = self.parent_node.get_sim_dir()
+            remote_sim_dir = f"{remote_home_dir}/sim_slot_{slotno}/"
+            hwcfg.local_xclbin = './local.xclbin'
+
+            with cd(remote_sim_dir):
+                run(downloadURI, hwcfg.xclbin, hwcfg.local_xclbin)
```
sifive-benjamin-morse, line 655: From my testing, this line doesn't compile. run() only accepts a string, which is a shell command, not a function. The 2nd and 3rd arguments are booleans, not arguments to the first function. Am I on the wrong version of fabric or something?  https://github.com/fabric/fabric/blob/1.14/fabric/operations.py#L970

timsnyder-siv: I think that we meant `execute()` instead of `run()`.

PR #1144, opened 2022-07-25
ci: support running under forks of firesim

davidbiancolin:
This PR makes it possible to run the FireSim's CI system under a fork of the repository. It:
- moves URLs that refer to `firesim/firesim` to `common.py`
  - it might be good to query this instead of having it hardcoded 
- breaks apart the initialize manager step into smaller sub-scripts
  - forks can insert a custom initialization procedure by running their own script. 
  - one could have a chipyard-as-top variant of the flow using this machinery (which is essentially what we're doing at SiFive)
 
#### UI / API Impact
N/C

N/C

### Contributor Checklist


.github/scripts/common.py
```diff
@@ -96,7 +103,7 @@ def deregister_runner_if_exists(gh_token, runner_name):
     runner_list = res_dict["runners"]
     for runner in runner_list:
         if runner_name in runner["name"]:
-            r = requests.delete("https://api.github.com/repos/firesim/firesim/actions/runners/{}".format(runner["id"]), headers=headers)
+            r = requests.delete("{}/{}".format(gha_runners_api_url, runner["id"]), headers=headers)
```
abejgonzalez, line 106:
```suggestion
            r = requests.delete(f"""{gha_runners_api_url}/{runner["id"]}""", headers=headers)
```

.github/scripts/workflow-monitor.py
```diff
@@ -36,11 +36,12 @@ def main(workflow_id, gha_ci_personal_token):
     state = None
     consecutive_failures = 0
     headers = {'Authorization': "token {}".format(gha_ci_personal_token.strip())}
+    gha_workflow_api_url = f"{gha_runs_api_url}/{workflow_id}"
```
abejgonzalez, line 39: Nit: I would just inline this into L44

.github/workflows/firesim-run-tests.yml
```diff
@@ -14,6 +14,7 @@ env:
   AWS-SECRET-ACCESS-KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
   AWS-DEFAULT-REGION: ${{ secrets.AWS_DEFAULT_REGION }}
   FIRESIM_PEM: ${{ secrets.FIRESIM_PEM }}
+  MANAGER_FIRESIM_LOCATION: "~/firesim"
```
abejgonzalez, line 17: I assume this being `/home/centos` creates problems w/ non-Centos based environments - what do you do on Azure? IIRC in the past there were some problems with having `~` or `$HOME` in scripts (though I can't find an example of it). If this is working, I'm fine with it.

.github/workflows/firesim-run-tests.yml
```diff
@@ -93,16 +94,20 @@ jobs:
     runs-on: ${{ github.run_id }}
     steps:
       - uses: actions/checkout@v2
+      - name: Install FireSim pem
+        run: ./.github/scripts/install-firesim-pem.py
       - name: Setup FireSim repo (.pem, build-setup.sh, AWS credentials, submodules) and CI daemons
-        uses: ./.github/actions/initialize-manager
+        run: ./.github/scripts/initialize-manager.py
```
abejgonzalez, line 100: I think we should rename this script to something like `initialize-repo`. Saying `init-manager` to me implies that the entire machine is getting setup.

.github/scripts/common.py
```diff
@@ -4,16 +4,23 @@
...
 import requests
 from local_flags import RUN_LOCAL, local_fsim_dir
+from ci_variables import ci_firesim_dir
 
 # Reuse manager utilities
 sys.path.append(local_fsim_dir + "/deploy/awstools")
 from awstools import get_instances_with_filter
 
+# Github URL related constants
+gh_repo_path        = "firesim/firesim"
```
abejgonzalez, line 14: This can be `$GITHUB_REPOSITORY`. See https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

.github/scripts/common.py
```diff
@@ -4,16 +4,23 @@
...
+gha_api_url         = f"https://api.github.com/repos/{gh_repo_path}/actions"
```
abejgonzalez, line 15: This can be constructed from `$GITHUB_API_URL`. See https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

abejgonzalez (commented):
Overall LGTM. A couple of updates/nits.

> one could have a chipyard-as-top variant of the flow using this machinery (which is essentially what we're doing at SiFive)

For my knowledge, how are you doing this? Are you just directly copying the workflow file, adding your script? I don't see any way to "inject" a script with these changes outside of breaking things up.

.github/workflows/firesim-run-tests.yml
davidbiancolin: I like that.

.github/scripts/common.py
davidbiancolin: I didn't want to introduce yet another ci_variable dep, because i'd have to sensibly handle the `run_local` case beyond hardcoding the string (which brings us back here). 

I could query the remotes and look for one named origin.

.github/scripts/common.py
davidbiancolin: See above run_local comment.

.github/workflows/firesim-run-tests.yml
davidbiancolin: We don't run on azure yet. I can switch it but this does seem to work fine.

davidbiancolin:
> For my knowledge, how are you doing this? Are you just directly copying the workflow file, adding your script? I don't see any way to "inject" a script with these changes outside of breaking things up.

I copied the workflow file, and swapped out part of the job that does the initialization. 

I could recollect the scripts into a single action which takes a "repo-initialization" script as an argument. That would be more DRY.

.github/scripts/common.py
abejgonzalez: We don't support running scripts locally that query the GH API (we set the `GITHUB_TOKEN/PERSONAL_ACCESS_TOKEN` to `0`), so it should be fine to add these in.

.github/workflows/firesim-run-tests.yml
abejgonzalez: This is fine with me.

.github/scripts/local_flags.py
```diff
@@ -1,4 +1,4 @@
 import os
 
 RUN_LOCAL = not os.environ.get('GITHUB_ACTIONS', False)
```
davidbiancolin, line 3: @abejgonzalez is there any reason this should be a separate file from ci_variables? 

It would be nice to move that there and better describe the semantics of local runs vs ci-hosted runs

abejgonzalez: Now that all the `RUN_LOCAL` usages would be in one file (`ci_variables.py`), I think it's fine to delete the `local_flags.py` file. I think there was a circular import issue but now looking at the code I don't see it. I think it would be nice if you could test this change locally by running `.github/scripts/run-mypy-typechecker.py` in your local FireSim clone just to be sure.

davidbiancolin: theres no circular dep, but mypy complains about all manner of things. I'll add the ci scripts directory to the type checking script

davidbiancolin: (in a follow up PR)

abejgonzalez: `mypy` is complaining about non-CI code or CI-code? I haven't added the CI to the typechecking.

davidbiancolin: CI code. Well the non-CI code too, but Tim fixed that.

davidbiancolin:
@mergifyio rebase

.github/scripts/common.py
timsnyder-siv: > We don't support running scripts locally that query the GH API

Why not?

davidbiancolin: I think the main barrier is providing an easy way to share the secrets (firesim.pem, and the personal access tokens) are the thorny ones really. Many of the other "secrets" , like aws region, needed be secret. 

I'd imagine it was the same for Abe, but when i initially developed some of this stuff i just set all the env variables manually.

.github/scripts/local_flags.py
timsnyder-siv: @davidbiancolin are you running mypy with the limited scope that https://github.com/firesim/firesim/blob/c240b2596f67ab6a3a3422565e38ada035715666/.github/scripts/run-mypy-typechecker.py#L11 uses?  You shouldn't see the CI scripts in the output of mypy or you're doing it wrong.

I'd honestly prefer that we switch to using `.mypy.ini` at the root of the repo instead of sticking the "right options" in a wrapper script that I'm always struggling to remember the name of.  However, https://mypy.readthedocs.io/en/stable/config_file.html#:~:text=then%20setup.cfg-,in%20the%20current%20directory,-%2C%20then%20%24XDG_CONFIG_HOME/mypy seems to imply that mypy doesn't look up the tree for a config file if it doesn't find one in the current working directory and I generally dislike that (yet understand why, it's inefficient).

something like https://pre-commit.com/ helps avoid this type of thing because it reduces the number of places stuff gets configured down to one for all the linters instead of one for each checker. (NOTE, I'm also assuming that it normalizes where you run from or at least refuses to run if you're not at the root of the repo but I haven't checked that yet).

davidbiancolin: I modified the bash script to include the ci dirs.

.github/scripts/workflow-monitor.py
timsnyder-siv: I disagree, headers is already hoisted up here.  This seems fine.

.github/scripts/local_flags.py
timsnyder-siv: Ok.  I was confused by what you were saying and thought you ran mypy directly maybe.

timsnyder-siv (commented):
This looks fine to me.  I'll ask Abe be the final word on merging it though.

.github/scripts/workflow-monitor.py
abejgonzalez: In other locations, `headers` is used in multiple places and thus is separated out (for ex:https://github.com/firesim/firesim/blob/b157da0b7ba889bf14d4a122aa04b044e849597a/.github/scripts/common.py#L87). Here ideally I think both `headers` and `gha_workflow_api_url` would be inlined (if they are used once inline). But like I said "nit" - very low priority.

abejgonzalez (approved):
LGTM

davidbiancolin:
I'm hitting this: https://github.com/firesim/firesim/issues/1147

Issue #1142, opened 2022-07-24
pytests failing on main@7839d4fd


timsnyder-siv:
If I checkout the main branch (currently https://github.com/firesim/firesim/commit/7839d4fda5eb360841694833c40f50e3e9bf3387)  and run pytest, there are failing tests. I expect that the pytests should be passing on main but it appears that something has snuck through causing them to fail.

I'm going to keep digging into this but wanted to file an issue so that others are aware.

timsnyder-siv:
One of the failures looks like:
```
    @wraps(func)
    def wrapper(self, *args, **kw):
        try:
            value = func(self, *args, **kw)
        except AssertionError as e:
>           raise AssertionError(e)
E           AssertionError: When calling b'main [firesim.py line 428]' the exception message does not match. Expected to match regex: 'INVALID_WORKLOAD'
E            against:
E            "[Errno 2] No such file or directory: 'config_build_recipes.yaml'"

../conda/envs/firesim/lib/python3.8/site-packages/sure/__init__.py:410: AssertionError
```

Thinking about this, it seems like it is being caused by manager support for metasim being added and wanting to read the `config_build_recipes.yaml` file when doing a "run" task.  That is new behavior of the manager added by metasim support and the `run_yamls` fixture in https://github.com/firesim/firesim/blob/7839d4fda5eb360841694833c40f50e3e9bf3387/deploy/tests/test_yaml_api.py#L169-L170 only writes `config_runtime.yaml` and `config_hwdb.yaml`.

I thought that #1076 may have caused them to start failing but looking at https://github.com/firesim/firesim/runs/6852034105?check_suite_focus=true#step:3:36 they passed. 

Rather than trying to track down where they didn't get run or started failing and were ignored, I'm going to bisect from where #1076 was merged into main to today's main and find the failing point.

timsnyder-siv:
It's weird that the CI passed on #1076 because when I checkout 0329104af77aa44af7d58218a4d4ce180a13361d where it was merged into main, the pytests fail.

timsnyder-siv:
And checking out the first parent of the merge of #1076 (previous value of main before merging it), the pytests pass.

timsnyder-siv:
Looking more into the CI run of #1076, I see it mentioning that it is running with SHA '4e5e0959b7d27e9d3f717c9558eeb926596eca3c' https://github.com/firesim/firesim/runs/6852034105?check_suite_focus=true#step:2:451

[...]

PR #1143, opened 2022-07-24
add config_build_recipes.yaml to run_yamls pytest fixture

timsnyder-siv:
Addition of config_build_recipes.yaml use by RuntimeConfig in #1076 broke
the pytests for config checking but somehow, it passed in the PR check.

#### Related PRs / Issues

Addresses broken tests portion of  for #1142 but doesn't do anything to address how this got through PR with 'passing' checks.

#### UI / API Impact

Isn't user visible, only updates pytests to reflect the need for `config_build_recipes.yaml` during `runworkload` and other "run" tasks.

#### Verilog / AGFI Compatibility

Doesn't impact Verilog.  Manager tests only.

### Contributor Checklist
- [X] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [X] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [X] Did you add at least one test demonstrating the PR?
- [X] Did you delete any extraneous prints/debugging code?
- [X] Did you state the UI / API impact?
- [X] Did you specify the Verilog / AGFI compatibility impact?

- [X] If applicable, did you regenerate and publicly share default AGFIs?

- [X] If applicable, did you apply the `ci:fpga-deploy` label?

- [X] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1145, opened 2022-07-25
ci: don't filter out python tests

davidbiancolin:
They run relatively quickly so there is little upside to selectively running them (see https://github.com/firesim/firesim/issues/1142). If we do want to filter on this we should be as permissive as possible. 

Somewhat relatedly, can we run these on a container instance? Python devs would get feedback sooner if we did that. Not a huge priority though. 

#### UI / API Impact

N/A

N/A

### Contributor Checklist

### DIFF
--- a/deploy/tests/test_yaml_api.py
+++ b/deploy/tests/test_yaml_api.py
@@ -175,7 +175,7 @@ def scy_build_recipes(tmp_path: Path, sample_backup_configs: Path) -> TmpYaml:
     return TmpYaml(tmp_path, sample_backup_configs / 'sample_config_build_recipes.yaml')
 
 @pytest.fixture()
-def build_yamls(scy_build, scy_build_recipes, scy_hwdb, non_existent_file) -> BuildTmpYamlSet:
+def build_yamls(scy_build: TmpYaml, scy_build_recipes: TmpYaml, scy_hwdb: TmpYaml, non_existent_file: Path) -> BuildTmpYamlSet:
     return BuildTmpYamlSet(scy_build, scy_build_recipes, scy_hwdb, non_existent_file)
 
 @pytest.fixture()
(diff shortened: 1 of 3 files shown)


timsnyder-siv:
@davidbiancolin , I don't understand how the pytests are passing in this PR without the changes from #1143

timsnyder-siv:
I think I see what's going on here.  If `firesim managerinit` has been run so that the sample config files have been moved into place, those get picked up by the pytests.  However, if `managerinit` hasn't been run, then there isn't a `config_build_recipes.yaml` for it to pick up and it fails after the metasimulation stuff was added to the manager.

timsnyder-siv:
I think that the pytests that are using the config files should either  cd into the temp_dir so that the default config files aren't picked up or pass non-existent file paths for configs that aren't being tested.  That will make them consistently pass or fail regardless of whether `managerinit` has been run.

timsnyder-siv:
I changed `test_yaml_api.py` to pass all config files to firesim so that it either passes or fails regardless of whether `managerinit` has been run.   I expect them to fail now and then we should merge #1143 into this PR to make them pass.  I'm going to go retarget that PR to this topic branch.

deploy/tests/test_yaml_api.py
```diff
@@ -169,14 +190,6 @@ def scy_runtime(tmp_path: Path, sample_backup_configs: Path) -> TmpYaml:
 def run_yamls(scy_hwdb: TmpYaml, scy_runtime: TmpYaml) -> RunTmpYamlSet:
     return RunTmpYamlSet(scy_hwdb, scy_runtime)
```
timsnyder-siv, line 191: Need to add the `non_existent_file` fixture to `run_yamls` fixture as well as `build_yamls`. 🤦‍♂️

timsnyder-siv: addressed in https://github.com/firesim/firesim/pull/1145/commits/eb55afb76e61a785dbff45e4712da1150f01d4c3

timsnyder-siv:
And after merging #1143  into this PR, pytests are passing again.  @abejgonzalez, could you give this stuff a review when you get a chance?  I'd like to get it merged

abejgonzalez (approved):
SGTM

davidbiancolin:
@timsnyder-siv You should green-button this one when you are ready.

PR #1154, opened 2022-08-03
PR Template: Explain how to run CI on PRs from forks.

davidbiancolin:
Not a great solution, but it'll have to do for now.


abejgonzalez (approved):
SFTM

PR #1155, opened 2022-08-04
add tests for downloadURI

filipstamenkovic-sifive:
@timsnyder I moved the function that fetches the URI from the VitisInstanceDeployManager class into it's own function. Then I wrote a simple test to download a SPEC result from the website twice so it overwrites an existing file on the file system, and assert the correct logging was done.

#### Related PRs / Issues

https://github.com/firesim/firesim/pull/1140

#### UI / API Impact

None

None

### Contributor Checklist


deploy/util/fsspec_utils.py
```diff
@@ -0,0 +1,27 @@
+import logging
```
timsnyder-siv, line 1: Let's rename this `io.py` so that it isn't specific to only fsspec

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+from unittest.mock import MagicMock, call
+from util.fsspec_utils import downloadURI
+from fsspec.implementations.local import LocalFileSystem
+
+
+def test_download_uri(mocker):
+    logger_mock = mocker.patch("util.fsspec_utils.rootLogger", MagicMock())
+    local_test_path = Path("tests/spec_results_download.csv")
+    test_uri = "https://www.spec.org/cpu2017/results/res2018q1/cpu2017-20171224-02028.csv"
```
timsnyder-siv, line 10: I don't want the tests to actually go out onto the internet to fetch resources from machines we have no control over while testing.  Do the fsspec tests themselves actually go out on the network or do they mock the network stuff out?

We should be able to use `moto` and and S3 URI and not require the actual internet to be working for the tests to pass.

timsnyder-siv: Refer to https://github.com/filipstamenkovic-sifive/firesim/blob/fsspec_for_bitstream/deploy/tests/awstools/test_awstools.py for examples of using moto to mock the ec2 API.

deploy/util/fsspec_utils.py
```diff
@@ -0,0 +1,27 @@
...
+from os import PathLike, fspath
+from fsspec.core import url_to_fs # type: ignore
+from pathlib import Path
+
+rootLogger = logging.getLogger()
+
+
+def downloadURI(uri: str, local_dest_path: PathLike):
```
timsnyder-siv, line 9:
```suggestion
def downloadURI(uri: str, local_dest_path: PathLike) -> None:
```
We've gotten in the habit of being very through with the type hints...

timsnyder-siv (changes requested):
This is a good start but we need the test to not fetch resources from the internet.

deploy/tests/test_utils.py
filipstamenkovic-sifive: Yeah that makes sense, I could see internet not working in for example a Github Actions environment blocking a CI pipeline because of this test. I'll change it so we mock S3 instead.

timsnyder-siv: Sounds good.  We're not really interested in testing http support but https://github.com/fsspec/filesystem_spec/blob/c99eca4d65eeede94be3f1eb1cd8d324571b821a/fsspec/tests/conftest.py#L146-L149 is the fixture fsspec uses to provide a local http server.

I think we want to have `file`, `ssh` and `s3` tests to start.  https://github.com/fsspec/filesystem_spec/blob/c99eca4d65eeede94be3f1eb1cd8d324571b821a/fsspec/implementations/tests/test_sftp.py#L20-L55 is where fsspec defines it's fixture for a local ssh server that runs inside a docker container (or skips the test if docker is not available).

@davidbiancolin and @abejgonzalez for testing that our integration of fsspec works, what do you think about depending on a nested docker container in CI?  Since we already spin up multiple servers in CI, perhaps we could:
1. write an integration test script at a higher level that knows about the other servers in CI
2. have the docker version in pytests for people who want to run it locally without other servers
3. skip it if docker isn't available.

filipstamenkovic-sifive: Unfortunately it seems like fsspec uses HTTP to interact with the S3 REST API and not boto3 for it's S3 integration. When I instantiate a file system using `url_to_fs, it's a `fsspec.implementations.http.HTTPFileSystem` instance, so using moto3 won't work to mock calls to AWS.

filipstamenkovic-sifive: To mock calls to S3, I think that I'll just have to mock HTTP calls out. I'll take a look at how fsspec does it in their tests.

filipstamenkovic-sifive: Nevermind, I learned how to properly generate an S3 URI for the test, so fsspec properly instantiates an `S3FileSystem` that wraps boto3... Now I can use moto to mock calls to s3.

deploy/tests/test_utils.py
```diff
@@ -1,37 +1,50 @@
 from pathlib import Path
 from unittest.mock import MagicMock, call
-from util.fsspec_utils import downloadURI
+from util.io import downloadURI
 from fsspec.implementations.local import LocalFileSystem
+from botocore.exceptions import ClientError
+import pytest
 
+pytest.mark.usefixtures("asw_test_credentials")
```
timsnyder-siv, line 8:
```suggestion
pytest.mark.usefixtures("aws_test_credentials")
```

deploy/tests/conftest.py
```diff
@@ -83,3 +86,14 @@ def patch(self, task_name: str, wrap_task=False, wrap_config=False) -> firesim.T
...
+#References: https://stackoverflow.com/questions/44718670/mock-s3-decorating-pytest-fixture
+@pytest.fixture()
+def mock_s3_client():
+    # setup: start moto server
+    mocks3 = mock_s3()
+    mocks3.start()
+    s3_client = boto3.client('s3')
+    yield s3_client
+    # teardown: stop moto server
+    mocks3.stop()
```
timsnyder-siv, line 99: Do you need to use `@pytest.yield_fixture` instead of the plain `pytest.fixture`?  I don't know the difference but I've also never written a fixture that yielded.

I have a preference for using `mock_s3` as a context manager as in https://github.com/spulec/moto/issues/620#issuecomment-613454360

I see how you need the fixture like this if you want to build up other fixtures that setup your testing state (i.e. create the bucket and 'upload' test collateral into it)

deploy/tests/test_utils.py
```diff
@@ -1,37 +1,50 @@
...
+        file_uri = f"s3://{test_bucket}/{test_bucket_key}"
+    except ClientError as e:
+        pytest.fail("Failed to mock an S3 client and upload a file.")
 
     fs = LocalFileSystem()
 
-    if local_test_path.exists():
-        fs.rm(str(local_test_path))
+    if test_dest_file_path.exists():
+        fs.rm(str(test_dest_file_path))
```
timsnyder-siv, line 28: nit: Since we're testing that the integration of ffspec still works for `downloadURI()`, it might be better to limit ourselves to stuff from the stdlib for doing this local filesystem manipulation.

deploy/tests/test_utils.py
```diff
@@ -1,37 +1,50 @@
...
 
     downloadURI(
```
timsnyder-siv, line 30: What is causing fsspec to use your `mock_s3_client`?  In the past, when testing things that don't explicitly take a client parameter, I've had to `patch.object` the `boto3._get_default_session` as in https://github.com/firesim/firesim/blob/e2e69dbb3bb6984c3b3017bae698c0d125205ec4/deploy/tests/awstools/test_awstools.py#L62-L71 to get the code under test to  use the mock client.  Perhaps my tests are doing something unnecessary though.

deploy/tests/conftest.py
filipstamenkovic-sifive: Not necessarily, a context manager will do just fine. I just wanted to add this fixture because it seemed like something that could be useful for other tests in the future.

But we should probably add it when it becomes necessary so we don't keep it around while it's unused.

deploy/tests/test_utils.py
filipstamenkovic-sifive: Agreed.

deploy/tests/conftest.py
```diff
@@ -83,3 +86,14 @@ def patch(self, task_name: str, wrap_task=False, wrap_config=False) -> firesim.T
 
     return TaskMocker(mocker)
 
+
+#References: https://stackoverflow.com/questions/44718670/mock-s3-decorating-pytest-fixture
+@pytest.fixture()
+def mock_s3_client():
+    # setup: start moto server
```
timsnyder-siv, line 93: This is another nitpicky thing.  The `mock_s3()` method isn't starting a server, is it?  I thought it's only swapping out the backend of boto to the moto mocks.  Starting a `moto_server` looks more like https://github.com/fsspec/s3fs/blob/18009c4ebdd19d688c3b10cd1dc7147dfaaae351/s3fs/tests/test_s3fs.py#L88

deploy/tests/test_utils.py
filipstamenkovic-sifive: I believe that when you use moto's mock_s3() as a decorator, it causes fsspec to also instantiate mocked s3 clients, basically the same client that's instantiated in setting up in the integration test.

filipstamenkovic-sifive: https://buildmedia.readthedocs.org/media/pdf/moto/latest/moto.pdf

I was reading these docs, and they give a good example on page 6 (page 10 of the pdf)

filipstamenkovic-sifive: I believe you've had to do an extra patch because you wanted to customize how your client behaves (return an authorization error).

deploy/tests/conftest.py
filipstamenkovic-sifive: Correct. I believe that the comment where I copied this code from is just worded poorly.

deploy/tests/test_utils.py
filipstamenkovic-sifive: I wonder how this didn't throw an exception....

timsnyder-siv: Yeah, me too.  Makes me think even more that we should just do the stuff in that fixture by default in conftest.py and swap the requirement to be so that people who want to really get out to AWS during pytest have to do work to undo the fake credentials.

deploy/tests/test_utils.py
```diff
@@ -1,37 +1,50 @@
 from pathlib import Path
 from unittest.mock import MagicMock, call
-from util.fsspec_utils import downloadURI
+from util.io import downloadURI
 from fsspec.implementations.local import LocalFileSystem
```
timsnyder-siv:
```suggestion
```

Don't need this anymore

deploy/tests/test_utils.py
```diff
@@ -1,37 +1,50 @@
 from pathlib import Path
 from unittest.mock import MagicMock, call
-from util.fsspec_utils import downloadURI
+from util.io import downloadURI
```
timsnyder-siv, line 3: Looking at the moto docs you sent me, I'm reminded that they recommend that you defer importing code that can reach out to AWS until after you have your mocks setup (and  you have the fake credentials installed). see http://docs.getmoto.org/en/latest/docs/getting_started.html#what-about-those-pesky-imports

I think that I'll open another PR that flattens aws_test_credentials into conftest.py. Or maybe you could after we're done with this one.

deploy/tests/test_utils.py
timsnyder-siv: Yeah, that's a good point.  I was misremembering why I needed  to do that.

deploy/tests/test_utils.py
filipstamenkovic-sifive: Agreed, if a boto3 is outside of a function, it'll be executed before there's a chance to mock it.

deploy/tests/conftest.py
```diff
@@ -5,6 +5,9 @@
 import os
 from os.path import dirname
 from pathlib import Path
+import boto3
+
+from moto import mock_s3
```
timsnyder-siv, line 10:
```suggestion
```
Should cleanup these imports since we removed the fixture.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -15,6 +15,7 @@
 from pathlib import Path
 
 from util.streamlogger import StreamLogger
+from util.fsspec_utils import downloadURI
```
timsnyder-siv, line 18:
```suggestion
from util.io import downloadURI
```
Need to update this import for the filename change.

timsnyder-siv (approved):
Thanks for doing this @filipstamenkovic-sifive, It's a great start to adding tests for fsspec integration into firesim.  I'm going to go ahead and merge this into my PR and I'll talk more with @abejgonzalez and @davidbiancolin this afternoon in the dev meeting.

PR #1166, opened 2022-08-09
Tag AGFI with sha of toplevel repo

timsnyder-siv:
Similar to how we changed repo_state_summary.sh to create repo_state file from
the toplevel git superproject, generate the commit tag for the AGFI from the toplevel
repo.

Add firesim_origin and top_origin tags to contain URL for those repos.

Add ispushed tags to indicate whether the commit existed on respective origin at the
time the AGFI was created.

#### Related PRs / Issues

Rewrite of #936.  h/t @filipstamenkovic-sifive for doing the rewrite!

#### UI / API Impact

Adds tags to AGFIs to aid in reproducing builds where Firesim is not the toplevel repo (Chipyard-as-top, etc).

Also adds tags to indicate whether the hash was pushed to a remote named `origin` or not at the time of build.

Adds `util.git` module to the firesim python for doing some of the tag introspection.

Doesn't 

### Contributor Checklist

### DIFF
--- a/deploy/tests/util/test_git.py
+++ b/deploy/tests/util/test_git.py
@@ -1,0 +1,165 @@
+import logging
+from os import fspath
+import pytest
+import re
+from subprocess import run, PIPE
+import sure
+
+from util.streamlogger import InfoStreamLogger
+
+
+rootLogger=logging.getLogger()
+
+class TestGitAssumptions:
+    """Test git features that we depend on"""
+
+    def test_git_fetch_pack_for_unadvertised_sha(self):
+        """in buildafi.git_origin_sha_is_pushed() we use git fetch-pack
+           to query the origin remote for the existence of a 40-digit sha
+           Unless the git remote server is configured to accept shas as arguments
+           (see the end of `git fetch-pack --help` for details) the code we use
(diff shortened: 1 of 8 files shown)


timsnyder-siv:
@Mergifyio rebase

deploy/tests/util/test_git.py
```diff
@@ -0,0 +1,159 @@
...
+
+from util.streamlogger import InfoStreamLogger
+
+
+rootLogger=logging.getLogger()
+
+class TestGitAssumptions:
+    """Test git features that we depend on"""
+
+    def test_git_fetch_pack_for_unadvertised_sha(self):
```
timsnyder-siv, line 16: @abejgonzalez can you help me understand why we need to create a second-level of clone in the CI scripts. See https://github.com/firesim/firesim/blob/0552050b95983be81bb71d03408241d0e3bfb783/.github/scripts/initialize-repo.py#L14-L15

This test is checking that the behavior of github (or whatever server the project is using to host the repo) but it assumes that we're in a clone that has origin set to the server we care about.

The test currently passes but it isn't testing anything because the clone it is running in was cloned from another directory on the CI host and not the actual github repository.

PR #1177, opened 2022-08-18
Fix various VCS metasimulation breakages

davidbiancolin:
This resolves three separate issues make VCS metasimulators unusable. We really need CI for this... 

1) 5e445b6: prevents driver-specific LD/CXX args from leaking into the metasimulator builds, and don't play nice with VCS in general

2) a8226d9: for some reason under VCS calling get on this pointer and dynamically casting it to mmio_t produces a pointer that no longer points the originating raw pointer. This doesn't seem like "right" fix i just put it here to start a conversation. 

3) 6679cdd: LDFLAGS brought it by conda, set `--as-needed` by default. The main issue is that VCS mangles the the order of objects and libraries such that libraries that are needed end up preceeding the objects that refer to them. I'm not sure how to fix the VCS invocation so i instead set `--no-as-needed`. 

The linkage of `mmio_t` `mm_t` instances between the driver context and the vcs/verilator context is a recurring problem. These are both global and `unique_ptr`s -- which seems like a gross misuse of `unique_ptr`. In SiFive's fork, i've switched these to be raw pointers because i was seeing instances of the runtime double-freeing the underlying data structure. 

I don't really see an easy way, short of re-engineeering how metasimulation is implemented (which is on the long list of things to do), to avoid using globals to share references to `mmio_t` and `mm_t`s. So it seems to me that either switching them to being `shared_ptr`s or raw pointers freed when the driver is torn down are the two best options. I'm open to alternatives. 

#### Related PRs / Issues

None

#### UI / API Impact

None

N/C

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/Makefile
+++ b/sim/midas/src/main/cc/Makefile
@@ -61,7 +61,7 @@ $(OUT_DIR)/$(DRIVER_NAME)-$(PLATFORM): $(design_h) $(lib) $(DRIVER) $(driver_h) 
 $(PLATFORM): $(OUT_DIR)/$(DRIVER_NAME)-$(PLATFORM)
 
 # Sources for building MIDAS-level simulators. Must be defined before sources VCS/Verilator Makefrags
-override CXXFLAGS += -include $(design_h)
+override CXXFLAGS += -std=c++17 -include $(design_h)
 
 # Models of FPGA primitives that are used in host-level sim, but not in FPGATop
 sim_fpga_resource_models := $(v_dir)/BUFGCE.v
--- a/sim/midas/src/main/cc/emul/mmio.cc
+++ b/sim/midas/src/main/cc/emul/mmio.cc
@@ -1,10 +1,6 @@
 #include "mmio.h"
-#include "mm.h"
-#include "mm_dramsim2.h"
 #include <cassert>
-#include <cmath>
-#include <iostream>
-#include <memory>
+#include <cstdlib>
 
 void mmio_t::read_req(uint64_t addr, size_t size, size_t len) {
   mmio_req_addr_t ar(0, addr, size, len);
@@ -85,26 +81,4 @@ bool mmio_t::write_resp() {
     write_inflight = false;
     return true;
   }
-}
-
-extern uint64_t main_time;
-extern std::unique_ptr<mmio_t> master;
-extern std::unique_ptr<mmio_t> dma;
-std::unique_ptr<mm_t> slave[MEM_NUM_CHANNELS];
-
-void init(uint64_t memsize, bool dramsim) {
-  master.reset(new mmio_t(CTRL_BEAT_BYTES));
-  dma.reset(new mmio_t(DMA_BEAT_BYTES));
-  for (int mem_channel_index = 0; mem_channel_index < MEM_NUM_CHANNELS;
-       mem_channel_index++) {
-    slave[mem_channel_index].reset(
-        dramsim ? (mm_t *)new mm_dramsim2_t(1 << MEM_ID_BITS)
-                : (mm_t *)new mm_magic_t);
-    slave[mem_channel_index]->init(memsize, MEM_BEAT_BYTES, 64);
-  }
-}
-
-void load_mems(const char *fname) {
-  slave[0]->load_mem(0, fname);
-  // TODO: allow file to be split across slaves
 }
--- a/sim/midas/src/main/cc/emul/vcs-harness.cc
+++ b/sim/midas/src/main/cc/emul/vcs-harness.cc
@@ -1,21 +1,15 @@
-#include "mm.h"
-#include "mm_dramsim2.h"
-#include "mmio.h"
+#include "simif_emul.h"
 #include <DirectC.h>
 #include <cassert>
 #include <cmath>
 #include <exception>
 #include <fesvr/context.h>
-#include <memory>
 #include <stdio.h>
 
 extern context_t *host;
 extern bool vcs_fin;
 extern bool vcs_rst;
 extern uint64_t main_time;
-extern std::unique_ptr<mmio_t> master;
-extern std::unique_ptr<mmio_t> dma;
-extern std::unique_ptr<mm_t> slave[MEM_NUM_CHANNELS];
 
 static const size_t CTRL_DATA_SIZE = CTRL_BEAT_BYTES / sizeof(uint32_t);
 static const size_t DMA_DATA_SIZE = DMA_BEAT_BYTES / sizeof(uint32_t);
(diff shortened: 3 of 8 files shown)


sim/target-agnostic.mk
```diff
@@ -39,18 +39,15 @@ PLATFORM ?=
...
 
-# Defined for each platform
-platforms_dir := $(abspath $(firesim_base_dir)/../platforms)
-vitis_CXX_FLAGS ?= -std=c++14 -idirafter ${CONDA_PREFIX}/include -idirafter /usr/include -idirafter $(XILINX_XRT)/include
-vitis_LD_FLAGS ?= -L${CONDA_PREFIX}/lib -Wl,-rpath-link=/usr/lib/x86_64-linux-gnu -L$(XILINX_XRT)/lib -luuid -lxrt_coreutil
-f1_CXX_FLAGS ?= -std=c++11 -I$(platforms_dir)/f1/aws-fpga/sdk/userspace/include
-f1_LD_FLAGS ?=
-
 # Target-specific CXX and LD flags for compiling the driver and meta-simulators
+# These should be platform independent should be governed by the target-specific makefrag
```
timsnyder-siv, line 43: I'm confused by this comment because it looks like you're moving what we're doing here into the targets defined in `target-agnostic.mk`, not into the target-specific makefrags.

sim/target-agnostic.mk
```diff
@@ -188,9 +186,11 @@ $(f1): $(header) $(DRIVER_CC) $(DRIVER_H) $(midas_cc) $(midas_h)
 	GEN_DIR=$(OUTPUT_DIR)/build OUT_DIR=$(OUTPUT_DIR) DRIVER="$(DRIVER_CC)" \
 	TOP_DIR=$(chipyard_dir)
 
-$(vitis): export CXXFLAGS := $(CXXFLAGS) $(common_cxx_flags) $(DRIVER_CXXOPTS)
+$(vitis): export CXXFLAGS := $(CXXFLAGS) $(common_cxx_flags) $(DRIVER_CXXOPTS) \
+	-std=c++14 -idirafter ${CONDA_PREFIX}/include -idirafter /usr/include -idirafter $(XILINX_XRT)/include
 # Statically link libfesvr to make it easier to distribute drivers to f1 instances
-$(vitis): export LDFLAGS := $(LDFLAGS) $(common_ld_flags) -Wl,-rpath='$$$$ORIGIN'
+$(vitis): export LDFLAGS := $(LDFLAGS) $(common_ld_flags) -Wl,-rpath='$$$$ORIGIN' \
+	-L${CONDA_PREFIX}/lib -Wl,-rpath-link=/usr/lib/x86_64-linux-gnu -L$(XILINX_XRT)/lib -luuid -lxrt_coreutil
```
timsnyder-siv, line 193: Just a note, all of the search-path arguments here can be removed once we switch to using the conda-packaged XRT libraries.  This was a temporary hack to make it so that things will compile & link correctly against the system-provided XRT on Ubuntu.  The paths need to be different on Centos.

sim/target-agnostic.mk
```diff
@@ -150,7 +147,7 @@ vcs = $(GENERATED_DIR)/$(DESIGN)
 vcs_debug = $(GENERATED_DIR)/$(DESIGN)-debug
 
 $(vcs) $(vcs_debug): export CXXFLAGS := $(CXXFLAGS) $(common_cxx_flags) $(VCS_CXXOPTS) -I$(VCS_HOME)/include -D RTLSIM
-$(vcs) $(vcs_debug): export LDFLAGS := $(LDFLAGS) $(common_ld_flags) -Wl,-rpath='$$$$ORIGIN'
+$(vcs) $(vcs_debug): export LDFLAGS := $(LDFLAGS) -Wl,--no-as-needed $(common_ld_flags) -Wl,-rpath='$$$$ORIGIN'
```
timsnyder-siv, line 150: I'm surprised by the need for this because we shouldn't have conda injecting any FLAGS variables anymore.  Perhaps we should put a specific test for that behavior in CI or in env.sh.

sim/midas/src/main/cc/emul/vcs-harness.cc
```diff
@@ -218,40 +218,39 @@ void tick(vc_handle reset,
           vc_handle mem_3_b_bits_resp,
           vc_handle mem_3_b_bits_id) {
   try {
-    mmio_t *m, *d;
-    assert(m = dynamic_cast<mmio_t *>(master.get()));
-    assert(d = dynamic_cast<mmio_t *>(dma.get()));
```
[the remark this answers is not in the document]
timsnyder-siv: It helps if I read the whole description of the PR carefully before going to do the review...   I see now where you say:
> The linkage of mmio_t mm_t instances between the driver context and the vcs/verilator context is a recurring problem. These are both global and unique_ptrs -- which seems like a gross misuse of unique_ptr. In SiFive's fork, i've switched these to be raw pointers because i was seeing instances of the runtime double-freeing the underlying data structure.
> 
> I don't really see an easy way, short of re-engineeering how metasimulation is implemented (which is on the long list of things to do), to avoid using globals to share references to mmio_t and mm_ts. So it seems to me that either switching them to being shared_ptrs or raw pointers freed when the driver is torn down are the two best options. I'm open to alternatives.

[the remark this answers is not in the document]
timsnyder-siv: Ok, after looking at how `master` and `dma` are [declared](https://github.com/firesim/firesim/blob/f9b0f1cf40e56942c6e5831eaacd15d7d94c8bea/sim/midas/src/main/cc/simif_emul.cc#L18-L19) and [initialized](https://github.com/firesim/firesim/blob/f9b0f1cf40e56942c6e5831eaacd15d7d94c8bea/sim/midas/src/main/cc/emul/mmio.cc#L96-L97). If I ignore whether it is good design to use globals for those pointers, I see that your change is effectively dropping the assertion that those pointers have been initialized  before this `tick()` is called.

At a minimum we should put the assertions back in:
```
assert(master.get() != nullptr)
assert(dma.get() != nullptr)
```
(if we change them to raw pointers, we should not call `.get()` on them but we still should have the assertions.

sim/target-agnostic.mk
davidbiancolin, line 150: Add a comment explaining this.

sim/target-agnostic.mk
davidbiancolin: What's the state of the XRT conda package? Do we just need someone to actually do the switch over?

sim/midas/src/main/cc/emul/vcs-harness.cc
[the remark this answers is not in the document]
timsnyder-siv: From our offline discussion, I agree that we should just use raw pointers here until we rearchitect the code.

sim/target-agnostic.mk
timsnyder-siv: We need to be able to bump to using 2022.1 and then start using it.  (Or, I could build older versions but I've been trying to avoid doing that).

I still don't see 2022.1 VM images in the Azure Portal.  @abejgonzalez mentioned trying 2022.1 on firesim1 or something at one point.

sim/target-agnostic.mk
davidbiancolin: I'm missing an "and" here. These were intended to give a mechanism for target-specific makefrags to glom on settings that are then passed to all driver/metasimulator targets.

sim/midas/src/main/cc/emul/mmio.cc
```diff
@@ -86,25 +82,3 @@ bool mmio_t::write_resp() {
     return true;
   }
 }
-
-extern uint64_t main_time;
```
davidbiancolin, line 90: This has been moved into simif_emul_t, which has a more functionally consistent scope.

sim/midas/src/main/cc/simif_emul.cc
```diff
@@ -15,8 +15,10 @@
 #include "bridges/cpu_managed_stream.h"
 
 uint64_t main_time = 0;
-std::unique_ptr<mmio_t> master;
-std::unique_ptr<mmio_t> dma;
+
+mmio_t *simif_emul_t::master = nullptr;
+mmio_t *simif_emul_t::dma = nullptr;
+mm_t *simif_emul_t::slave[MEM_NUM_CHANNELS] = {nullptr};
```
timsnyder-siv, line 21: I know you can initialize `master` and `dma` to something real here and remove the new and delete calls for them in the constructor/destructor.  That would make the memory handling thread safe because they will be allocated before `mainI()` is called.  There's more logic for handling the `slave` channels, so I don't think that will work here.

timsnyder-siv (commented):
As we discussed in sideband, I don't think you want to be calling delete on the static members in the destructor.

sim/midas/src/main/cc/Makefile
```diff
@@ -61,7 +61,7 @@ $(OUT_DIR)/$(DRIVER_NAME)-$(PLATFORM): $(design_h) $(lib) $(DRIVER) $(driver_h)
 $(PLATFORM): $(OUT_DIR)/$(DRIVER_NAME)-$(PLATFORM)
 
 # Sources for building MIDAS-level simulators. Must be defined before sources VCS/Verilator Makefrags
-override CXXFLAGS += -include $(design_h)
+override CXXFLAGS += -std=c++17 -include $(design_h)
```
davidbiancolin, line 64: Note i switched to C++17 here:
- to harmonize it across all builds
- because it was previously requested
- i wanted to use `inline` on static members. 

It might be out of scope for this change and i can kick it down the road.

sim/midas/src/main/cc/emul/verilator-harness.cc
```diff
@@ -1,141 +1,148 @@
...
-  top->dma_ar_bits_size = d->ar_size();
-  top->dma_ar_bits_len = d->ar_len();
-
-  top->dma_w_valid = d->w_valid();
-  top->dma_w_bits_strb = d->w_strb();
-  top->dma_w_bits_last = d->w_last();
-
-  top->dma_r_ready = d->r_ready();
-  top->dma_b_ready = d->b_ready();
+  top->ctrl_aw_valid = simif_emul_t::simif_emul_t::master->aw_valid();
```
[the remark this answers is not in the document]
timsnyder-siv: @davidbiancolin this still looks like it is wrong.  I'm surprised that this compiles
```suggestion
  top->ctrl_aw_valid = simif_emul_t::master->aw_valid();
```

sim/target-agnostic.mk
abejgonzalez: I tried it at one point but until everyone can bump to 2022.1 successfully I held back from bumping everything (Bumping to 2022.1 breaks AWS builds supposedly @sagark )

sim/midas/src/main/cc/emul/verilator-harness.cc
```diff
@@ -1,141 +1,148 @@
...
+
+  top->dma_r_ready = simif_emul_t::dma->r_ready();
+  top->dma_b_ready = simif_emul_t::dma->b_ready();
 #if DMA_DATA_BITS > 64
-  memcpy(top->dma_w_bits_data, d->w_data(), DMA_BEAT_BYTES);
+  memcpy(top->dma_w_bits_data, simif_emul_t::dma->w_data(), DMA_BEAT_BYTES);
 #else
-  memcpy(&top->dma_w_bits_data, d->w_data(), DMA_BEAT_BYTES);
+  memcpy(&top->dma_w_bits_data, simif_emul_t::dma->w_data(), DMA_BEAT_BYTES);
 #endif
```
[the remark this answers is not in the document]
timsnyder-siv: @davidbiancolin explained offline that both VCS and Verilator have a bitwidth where the underlying type changes and the memcpy calls change from pass by value to pass by reference.

We both agree that it would be nice to avoid this with some abstraction layer in the rewrite, possibly DPI.

timsnyder-siv (approved):

PR #1160, opened 2022-08-06
Use libelf and libdwarf from conda

timsnyder-siv:
Changes dependencies for libelf and libdwarf to come from conda packages instead of submodules that are compiled
from source.

#### UI / API Impact

Needs to update driver to newer APIs of libelf and libdwarf but shouldn't change firesim-user facing API/UI 

#### Verilog / AGFI Compatibility

Doesn't touch Verilog or AGFI.

### Contributor Checklist
- [X] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [X] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [X] Did you add at least one test demonstrating the PR?
- [X] Did you delete any extraneous prints/debugging code?
- [X] Did you state the UI / API impact?
- [X] Did you specify the Verilog / AGFI compatibility impact?

- [X] If applicable, did you regenerate and publicly share default AGFIs?

- [X] If applicable, did you apply the `ci:fpga-deploy` label?

- [X] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1180, opened 2022-08-24
refactor most of machine-launch-script.sh into build-setup.sh

timsnyder-siv:
use `scripts/conda-requirements.yaml` to create `./.conda-env/` and activate that environment in `env.sh`

#### Related PRs / Issues

Is based on #1160 and currently targeting being merged into that branch instead of main, only to minimize the differences to what is being proposed here.

#### UI / API Impact

Moves the conda environment from one 'named' `firesim` to `.conda-env` in the root of the firesim clone.  This is only created when not using 'library' mode or 'skip toolchain'.  In this way, Chipyard can create it's own conda env.

No change.

### Contributor Checklist


timsnyder-siv:
As I use this, I'm noticing that the conda PS1 settings seem to be saying that `base` environment is activated even after activating this.  I'm not sure how it will impact other prompts that 'know' how to display which conda environment is activated.  It may need to be tweaked for that.

build-setup-nolog.sh
```diff
@@ -173,36 +173,10 @@ fi
...
-        cd "$target_chipyard_dir"
-        if [ "$FASTINSTALL" = "true" ] ; then
-            MAKE=make ./scripts/build-toolchains.sh ec2fast
-        else
-            MAKE=make ./scripts/build-toolchains.sh
-        fi
-    )
-    source "$target_chipyard_dir/env.sh"
-    env_append "source $target_chipyard_dir/env.sh"
+    conda env create -f=./scripts/conda-requirements.yaml -p=./.conda-env
```
timsnyder-siv, line 176:
```suggestion
    (set -x; conda env create -f=./scripts/conda-requirements.yaml -p=./.conda-env)
```
echo the conda environment creation command so that it is logged and it makes the output of the script a bit more understandable for the user as well.

build-setup-nolog.sh
```diff
@@ -173,36 +173,10 @@ fi
...
+    source ./.conda-env/etc/profile.d/conda.sh
+    conda activate ./.conda-env
+    env_append "conda activate ./.conda-env"
```
timsnyder-siv, line 179: I think the prevailing wisdom is to install conda separately from creating environments with it, hence, I don't add the  `source ./.conda-env/etc/profile.d/conda.sh` into `env.sh`.  Since `conda` is implemented as a shell-function, any script that wants to use `conda activate` must bootstrap conda into the environment by sourcing the equivalent of what `conda init` puts into the shell init scripts.

scripts/machine-launch-script.sh
```diff
@@ -212,108 +212,15 @@ set -o pipefail
...
-
-    # python packages
-    # While it is possible to install using pip after creating the
-    # conda environment, pip's dependency resolution can conflict with
-    # conda and create broken environments.  It's best to use the conda
-    # packages so that the environment is consistent
-    CONDA_PACKAGE_SPECS+=( \
-        boto3==1.20.21 \
-        colorama==0.4.3 \
         argcomplete==1.12.3 \
```
timsnyder-siv, line 223: I left bash-completion and argcomplete in machine-launch-script.sh because I didn't want to think through how to handle them being moved into build-setup.sh.  It may be that they should or need to move or not move for proper function.

timsnyder-siv:
With what's in here so far, I hit:
```
make[1]: *** No rule to make target '/home/centos/src/project_data/<my_firesim_top_clone>/firesim/target-design/chipyard/tools/dromajo/dromajo-src/src/libdromajo_cosim.a', needed by '/home/centos/src/project_data/<my_top>/firesim/sim/output/f1/<blah blah blah>/FireSim-f1'.  Stop.
```

I suspect that the path to `libdromajo_cosim.a` needs to be updated somewhere in the FireSim makefiles.

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,120 @@
...
+    #
+    # pin to gcc=10 until we get that fixed.
+
+    - gcc =10 
+    - gxx =10 
+    - binutils 
+    - conda-gcc-specs
+
+    - dromajo
+    - riscv-tools
```
abejgonzalez, line 43: I think we should install the tools separately in the `build-setup...` script from the `conda-requirements.yaml`. Either that or we remove the option in the setup script to use your own toolchain and we document that to choose another toolchain you have to uninstall `riscv-tools` from the env and install `esp-tools` (or use your own toolchain). I think the former option is a bit better (installing the toolchain separately) / less documentation.

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,120 @@
...
+    - libusb1
+    - ncurses
+    - cmake
+    - graphviz
+    - expect
+    - dtc
+    - verilator==4.224
+    - screen
+    - elfutils
+    - libdwarf-dev==0.0.0.20190110_28_ga81397fc4
```
abejgonzalez, line 84: Now that the toolchain build is handled separately from here, I think we can remove a lot of these extra packages. I would look at the minimal set I created in https://github.com/ucb-bar/chipyard/pull/1163

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,120 @@
...
+    - clang-tools
+
+    # python packages
+    # While it is possible to install using pip after creating the
+    # conda environment, pip's dependency resolution can conflict with
+    # conda and create broken environments.  It's best to use the conda
+    # packages so that the environment is consistent
+    - boto3==1.20.21
+    - colorama==0.4.3
+    - argcomplete==1.12.3
```
abejgonzalez, line 97: (Low priority) I'm concerned about duplication between `machine-launch-script.sh` and this, specifically listing out needed packages and their versions. They are installed in separate environments but updating things in two places might cause unforeseen bugs later on (although minor).

abejgonzalez (changes requested):

build-setup-nolog.sh
```diff
@@ -173,36 +173,10 @@ fi
 # When FireSim is being used as a library, the user is expected to build their
 # own toolchain. For FireSim-as-top, call out to Chipyard's toolchain scripts.
 if [ "$SKIP_TOOLCHAIN" != true ]; then
```
davidbiancolin, line 175: Would we not want to create this environment whenever `IS_LIBRARY` is set false, regardless of what we are doing with the toolchain? 

Does it even make sense to have the skip-toolchain functionality if conda is bringing it in fairly cheapily?

abejgonzalez: Agreed. The environment should always be created. However, I think we should still have the skip toolchain option for power users who need their own toolchain

timsnyder-siv: Because I think that Chipyard is going to want to create a conda environment when they are "top" and it seemed to make sense that firesim would cede it's ownership of the environment when it isn't "top"

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,120 @@
...
+    #   filterable list of all conda-forge packages
+    # https://conda-forge.org/#contribute
+    #   instructions on adding a recipe
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/pkg-specs.html#package-match-specifications
+    #   documentation on package_spec syntax for constraining versions
+
+
+    # handy tool for introspecting package relationships and file ownership
+    # see https://github.com/rvalieris/conda-tree
+    - conda-tree
```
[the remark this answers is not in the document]
timsnyder-siv: I use it all the time but it isn't required by anything in the flow.

scripts/conda-requirements.yaml
timsnyder-siv: PRs welcome!  I was starting this PR as a concrete talking point for your request to try out the toolchain, not as an attempt to minimize the package set.  `conda-tree` is really good at helping someone come up with the minimal set of requirements if that's something you want to do.

scripts/conda-requirements.yaml
timsnyder-siv: I was going to mention that best-practice is to create the conda env from a single list of requirements so that the solver can take all of the depencencies into account.  Installing things incrementally is possible but can be problematic.

I tried looking for a decent writeup of "why" it is best to install everything at once and I'm not finding it right now.  I'll look more later.

timsnyder-siv: I've thought it would probably be best to have things specified in multiple `conda-requirements.yaml` files that are appropriately collected by whatever helper script and given to a single `conda create env` command.  It can take multiple `-f` arguments to solve and install at once.

abejgonzalez: I think that having multiple req. files would be fine. I just think there should be some sort of mechanism for users to choose which toolchain they want (probably by having the `build.sh` combine the main requirements + the toolchain requirements for them as you mentioned).

scripts/conda-requirements.yaml
abejgonzalez: SGTM. I can help with this. We can just merge this PR in 1st (when its ready) then there can be a follow PR to clean up things.

build-setup-nolog.sh
davidbiancolin: To be more clear: FireSim as top implies IS_LIBRARY = false, it does _not_ imply SKIP_TOOLCHAIN = true currently. 

For instance, i often skip building the toolchain, like on mill where i can't use the prebuilt one, when using FireSim as top currently.

PR #1196, opened 2022-09-01
Metasimulation: remove dependency on fesvr for ucontext

davidbiancolin:
This is part 1 of simplifying FireSim's dependencies on external libraries for metasimulation. Metasimulation support for VCS has a dependence on FESVR for setting up a coroutine for VCS execution. We lazily brought that in with the RISC-V toolchain. 

Instead of trying to make this dependency on spike/fesvr more explicit, here i copy the two source files we care about in tree as we have previously discussed. 

----

As part of removing dramsim2, i discovered that the stack that was being allocated (in the heap) for the VCS context was only 64K. I suspect modern versions of VCS need a larger stack: the segfaults i was seeing previously were due to VCS's stack overflowing into previously allocated structures on the heap.

In the long run, we're going to switch to DPI and remove the threading here. For now we need to make this less dangerous:
1) Just allocate much larger stack dynamically (as i have done initially)
2) Allocate a larger stack statically; somewhere else with less risk of overflow / or more obvious error. 
3) Just use the pthreads implementation of fesvr/context. I'm not sure we really need the light weight context switching ucontext gives us. 

The main issue is i don't know how much stack VCS needs. Suggestions? 

#### Related PRs / Issues

#1177 

#### UI / API Impact

None

N/C

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/emul/context.cc
+++ b/sim/midas/src/main/cc/emul/context.cc
@@ -50,8 +50,8 @@ void context_t::init(void (*f)(void*), void* a)
 #ifdef USE_UCONTEXT
   getcontext(context.get());
   context->uc_link = creator->context.get();
-  // Note this is 1024x larger than what is found upstream in spike
-  context->uc_stack.ss_size = 64* 1024 * 1024;
+  // Note this is larger than what is upstream in spike to support using VCS as a child context
+  context->uc_stack.ss_size = 64 * 1024 * 1024;
   context->uc_stack.ss_sp = new void*[context->uc_stack.ss_size/sizeof(void*)];
 #ifndef GLIBC_64BIT_PTR_BUG
   makecontext(context.get(), (void(*)(void))&context_t::wrapper, 1, this);
--- a/sim/midas/src/main/cc/emul/context.cc
+++ b/sim/midas/src/main/cc/emul/context.cc
@@ -1,38 +1,34 @@
-// clang-format off
 #include "context.h"
 #include <assert.h>
 #include <sched.h>
 #include <stdlib.h>
 
-static __thread context_t* cur;
+static __thread context_t *cur;
 
 context_t::context_t()
-  : creator(NULL), func(NULL), arg(NULL),
+    : creator(NULL), func(NULL), arg(NULL),
 #ifndef USE_UCONTEXT
-    mutex(PTHREAD_MUTEX_INITIALIZER),
-    cond(PTHREAD_COND_INITIALIZER), flag(0)
+      mutex(PTHREAD_MUTEX_INITIALIZER), cond(PTHREAD_COND_INITIALIZER), flag(0)
 #else
-    context(new ucontext_t)
+      context(new ucontext_t)
 #endif
 {
 }
 
 #ifdef USE_UCONTEXT
 #ifndef GLIBC_64BIT_PTR_BUG
-void context_t::wrapper(context_t* ctx)
-{
+void context_t::wrapper(context_t *ctx) {
 #else
-void context_t::wrapper(unsigned int hi, unsigned int lo)
-{
-  context_t* ctx = reinterpret_cast<context_t*>(static_cast<unsigned long>(lo) | (static_cast<unsigned long>(hi) << 32));
+void context_t::wrapper(unsigned int hi, unsigned int lo) {
+  context_t *ctx = reinterpret_cast<context_t *>(
+      static_cast<unsigned long>(lo) | (static_cast<unsigned long>(hi) << 32));
 #endif
   ctx->creator->switch_to();
   ctx->func(ctx->arg);
 }
 #else
-void* context_t::wrapper(void* a)
-{
-  context_t* ctx = static_cast<context_t*>(a);
+void *context_t::wrapper(void *a) {
+  context_t *ctx = static_cast<context_t *>(a);
   cur = ctx;
   ctx->creator->switch_to();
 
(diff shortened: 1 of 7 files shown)


sim/midas/src/main/cc/emul/context.h
```diff
@@ -0,0 +1,59 @@
...
+#endif
+#endif /* ULONG_MAX > UINT_MAX */
+
+#endif
+
+class context_t
+{
+ public:
+  context_t();
+  ~context_t();
```
nandor, line 34: If this is not a final class, please make the destructor `virtual` otherwise it will not be called. If it is a final class, please use `class context_t final`.

sim/midas/src/main/cc/emul/context.h
```diff
@@ -0,0 +1,59 @@
...
+# undef USE_UCONTEXT
+# define USE_UCONTEXT
+# include <ucontext.h>
+# include <memory>
+#include <limits.h>
+
+#if (ULONG_MAX > UINT_MAX) // 64-bit systems only
+#if (100*GLIB_MAJOR_VERSION+GLIB_MINOR_VERSION < 208)
+#define GLIBC_64BIT_PTR_BUG
+static_assert (sizeof(unsigned int)  == 4, "uint size doesn't match expected 32bit");
```
nandor, line 22: What is the source of this problem? Why noi use `uint32_t` for 4-byte values, `uint64_t` for 8-byte values and `uintptr_t` for pointers cast to integers?

nandor (commented):
Are we not using clang-format already?

I left some comments on `context.h` header file. I know it's copied code and if we are to replace `ucontext` with something that is not deprecated please ignore them. Maybe we should look for an alternative library implementing coroutines.

sim/midas/src/main/cc/emul/context.h
davidbiancolin: I'm not entirely sure what the context was, but i'd rather not fiddle too much with this code.

davidbiancolin:
Do you have any suggestions for an alternate co-routine library? I don't think C++20 coroutines will suffice in this context.

Longer term we were planning on moving the driver behind a DPI interface so that it would be easy to metasimulate stuff using other simulators beyond VCS / verilator. Unfortunately, modern VCS really wants to be main.   Anyways, this would remove the need for the co-routine library and delete a bunch of per-simulator harness code.

sim/midas/src/main/cc/emul/context.h
```diff
@@ -0,0 +1,59 @@
...
+static_assert (sizeof(unsigned long) == 8, "ulong size doesn't match expected 64bit");
+static_assert (sizeof(void*)         == 8, "ptr size doesn't match expected 64bit");
+#endif
+#endif /* ULONG_MAX > UINT_MAX */
+
+#endif
+
+class context_t
```
davidbiancolin, line 30:
```suggestion
class context_t final
```

sim/midas/src/main/cc/emul/context.cc
```diff
@@ -0,0 +1,118 @@
...
+void context_t::init(void (*f)(void*), void* a)
+{
+  func = f;
+  arg = a;
+  creator = current();
+
+#ifdef USE_UCONTEXT
+  getcontext(context.get());
+  context->uc_link = creator->context.get();
+  // Note this is 1024x larger than what is found upstream in spike
```
davidbiancolin, line 53:
```suggestion
  // Note this is larger than what is upstream in spike to support using VCS as a child context
```

sim/midas/src/main/cc/emul/context.cc
```diff
@@ -0,0 +1,118 @@
...
+  context->uc_stack.ss_size = 64* 1024 * 1024;
```
davidbiancolin, line 54:
```suggestion
  context->uc_stack.ss_size = 64 * 1024 * 1024;
```

davidbiancolin:
I disabled clang-format on these files only to make the diff easier to grok with upstream spike. I can remove that if you want.

sim/midas/src/main/cc/emul/context.cc
```diff
@@ -0,0 +1,118 @@
...
+{
+  func = f;
+  arg = a;
+  creator = current();
+
+#ifdef USE_UCONTEXT
+  getcontext(context.get());
+  context->uc_link = creator->context.get();
+  // Note this is larger than what is upstream in spike to support using VCS as a child context
+  context->uc_stack.ss_size = 64 * 1024 * 1024;
```
davidbiancolin, line 54:
```suggestion
  context->uc_stack.ss_size = 8 * 1024 * 1024;
```

sim/midas/src/main/cc/emul/context.cc
```diff
@@ -0,0 +1,118 @@
...
+void context_t::init(void (*f)(void*), void* a)
+{
+  func = f;
+  arg = a;
+  creator = current();
+
+#ifdef USE_UCONTEXT
+  getcontext(context.get());
+  context->uc_link = creator->context.get();
+  // Note this is larger than what is upstream in spike to support using VCS as a child context
```
davidbiancolin, line 53:
```suggestion
  // Note this is larger than what was historically upstream in spike to support using
  // VCS as a child context. 8 MiB was chosen as it is a typical linux max stack size. 
```

nandor (approved):
I don't know any decent substitutes for ucontext. Maybe simply using pthread could work? Also, I think we should keep clang-format enabled here.

PR #1197, opened 2022-09-01
Metasimulation: remove dramsim2 and copy host memory model sources in-tree

davidbiancolin:
1) Removes DRAMSim2, libmidas.a compilation. Switches over to using `mm_magic_t`
2) Copies in `mm.h` and `mm.cc` back from testchipip. 

This is in service of making it easier to deploy FireSim as a library, and should make it a little easier to do a redesign the driver + build system. 

The downside is now we only have one host-memory system model. I wrote a fuzzing implementation of `mm` long ago, but i think it might be easiest to use some of diplomacy's fuzzing utilities (which can then be backed by mm_magic_t) instead. That would more productively verify host-memory system changes vs changing between `mm_magic_t` and dramsim. 

#### Related PRs / Issues

#1196 

#### UI / API Impact

+dramsim now does nothing when passed to metasimulation. 

No change, as this feature is used only in metasim.
 

### Contributor Checklist


nandor:
This looks good to me - thanks for doing this change. Could you follow this up with a PR to set identical warning-related flags in all `Makefrag` files? `-Wall -Wextra -Werror` would be great. There were some nasty warnings in DRAMSim from `-Wextra`, but now I think that the remaining issues could be fixed.

sim/midas/src/main/cc/rtlsim/Makefrag-vcs
```diff
@@ -35,6 +35,7 @@ override VCS_FLAGS := -quiet -timescale=1ns/1ps +v2k +rad +vcs+initreg+random +v
 vcs_v := $(emul_v) $(vcs_wrapper_v)
 vcs_cc := $(emul_cc) $(vcs_harness)
 
+$(info $(emul_h))
```
davidbiancolin, line 38:
```suggestion
```

davidbiancolin:
> This looks good to me - thanks for doing this change. Could you follow this up with a PR to set identical warning-related flags in all `Makefrag` files? `-Wall -Wextra -Werror` would be great. There were some nasty warnings in DRAMSim from `-Wextra`, but now I think that the remaining issues could be fixed.

Yeah, i will do that next week.

nandor (approved):
LGTM

PR #1202, opened 2022-09-08
Removed the Fame1Instances transformation

nandor:
The `Fame1Instances` transformation was used solely to clock gate the model within `FasedMemoryTimingModel`. Instead of the transformation which individually altered all memories and registers within the hierarchy, now a clock gate is used on the clock signal passed to the model.

Altered `HostSpecialization` for Xilinx to replace all modules which have `AbstractClockGate` in their name since the elaboration of the simulation wrapper generates the `AbstractClockGate_0` name to avoid duplication.

#### Related PRs / Issues

N/A

#### UI / API Impact

N/A

N/A

### Contributor Checklist


sim/midas/src/main/scala/midas/passes/xilinx/package.scala
```diff
@@ -15,7 +15,7 @@ package object xilinx {
 
   object ReplaceAbstractClockGates extends Transform with NoAnalysisPass {
     val transformer = StatementTransformer {
-      case wi: WDefInstance if wi.module == DefineAbstractClockGate.blackboxName =>
+      case wi: WDefInstance if wi.module.contains(DefineAbstractClockGate.blackboxName) =>
```
fabianschuiki, line 18: Would it make sense to check the extmodule's `defname` instead? That can be picked pretty freely, and in case multiple extmodule declarations exist with deduplicated names, the `defname` would still be identical among all of them.

I guess the downside is that you'd have to go lookup the extmodule definition here instead of being able to tell from the `WDefInstance` that the module is a clock gate :thinking:

fabianschuiki: Cool!

davidbiancolin (approved):
I think this looks much better, though for now can you put the blackbox stub in passes with DefineAbstractClockGate, and add scaladoc explaining it can be used to clock gate bridge logic? I'll think of a better long term home for it.

nandor:
I didn't move the class to `passes` since `chisel` names would then conflict with names imported from `firrtl` for the transformations. I moved it to the Utils file and avoided hardcoding the module name by referring to the one in DefineAbstractClockGates.

PR #1235, opened 2022-09-27
Add XDCAnnotation type hints, end-to-end test, empty arg list support

fabianschuiki:
This commit does three things:

- Add a `CustomAnnotations` test to the midasexamples test suite which is basically the shift register test with additional uses of the `XDC(...)` Chisel API. The test checks whether the expected files get generated and contain the expected lines.

- Add an early out to `WriteXDCFile.formatArguments` in case the format argument list is empty. This allows the user to specify an XDC constraint that contains no format arguments, which until now would cause an exception in `argumentList.head`. The end-to-end test checks for this.

- Add type hints to the `XDCAnnotation` such that the `XDCDestinationFile` case objects get serialized properly.

#### Related PRs / Issues

None

#### UI / API Impact

No impact

No impact

### Contributor Checklist

### DIFF
--- a/sim/src/main/scala/midasexamples/CustomConstraints.scala
+++ b/sim/src/main/scala/midasexamples/CustomConstraints.scala
@@ -1,0 +1,16 @@
+//See LICENSE for license details.
+
+package firesim.midasexamples
+
+import chisel3._
+import freechips.rocketchip.config.Parameters
+import midas.targetutils.xdc._
+
+class CustomConstraintsDUT extends ShiftRegisterDUT {
+  XDC(XDCFiles.Synthesis, "constrain_synth1")
+  XDC(XDCFiles.Synthesis, "constrain_synth2 [reg {}]", r0)
+  XDC(XDCFiles.Implementation, "constrain_impl1")
+  XDC(XDCFiles.Implementation, "constrain_impl2 [reg {}]", r1)
+}
+
+class CustomConstraints(implicit p: Parameters) extends PeekPokeMidasExampleHarness(() => new CustomConstraintsDUT)
--- a/sim/src/test/scala/midasexamples/TutorialSuite.scala
+++ b/sim/src/test/scala/midasexamples/TutorialSuite.scala
@@ -377,6 +377,23 @@ class TerminationF1Test extends TutorialSuite("TerminationModule") {
   1 to 10 foreach {x => runTest(backendSimulator, args = Seq("+termination-bridge-tick-rate=10", s"+seed=${x}"), shouldPass = true)}
 }
 
+class CustomConstraintsF1Test extends TutorialSuite("CustomConstraints") {
+  def readLines(filename: String): List[String] = {
+    val file = new File(genDir, s"/${filename}")
+    Source.fromFile(file).getLines.toList
+  }
+  it should s"generate synthesis XDC file" in {
+    val xdc = readLines("FireSim-generated.synthesis.xdc")
+    xdc should contain("constrain_synth1")
+    atLeast (1, xdc) should fullyMatch regex "constrain_synth2 \\[reg firesim_top/.*/dut/r0\\]".r
+  }
+  it should s"generate implementation XDC file" in {
+    val xdc = readLines("FireSim-generated.implementation.xdc")
+    xdc should contain("constrain_impl1")
+    atLeast (1, xdc) should fullyMatch regex "constrain_impl2 \\[reg WRAPPER_INST/CL/firesim_top/.*/dut/r1]".r
+  }
+}
+
 // Suite Collections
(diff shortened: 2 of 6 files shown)


sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -377,6 +377,23 @@ class TerminationF1Test extends TutorialSuite("TerminationModule") {
...
 
+class CustomConstraintsF1Test extends TutorialSuite("CustomConstraints") {
+  def readLines(filename: String): List[String] = {
+    val file = new File(genDir, s"/${filename}")
+    Source.fromFile(file).getLines.toList
+  }
+  it should s"generate synthesis XDC file" in {
+    val xdc = readLines("FireSim-generated.synthesis.xdc")
+    xdc should contain("constrain_synth1")
+    xdc should contain("constrain_synth2 [reg firesim_top/top/sim/target/CustomConstraints_/dut/r0]")
```
davidbiancolin, line 388: Do you have any ideas for ways we could avoid hardcoding a module path here? These are liable to change.

sim/src/test/scala/midasexamples/TutorialSuite.scala
```diff
@@ -377,6 +377,23 @@ class TerminationF1Test extends TutorialSuite("TerminationModule") {
...
+  }
+  it should s"generate implementation XDC file" in {
+    val xdc = readLines("FireSim-generated.implementation.xdc")
+    xdc should contain("constrain_impl1")
+    xdc should contain("constrain_impl2 [reg WRAPPER_INST/CL/firesim_top/top/sim/target/CustomConstraints_/dut/r1]")
+  }
+}
+
 // Suite Collections
 class ChiselExampleDesigns extends Suites(
```
davidbiancolin, line 398: You should add your constraint test to one of the CI groups below.

davidbiancolin (commented):
This is a nice improvement, thanks for fixing the 0-argument case. 

There are a lot of hardcoded strings in the test, but this is consistent with what i've done elsewhere.

sim/src/test/scala/midasexamples/TutorialSuite.scala
fabianschuiki: Yeah the hardcoded strings also make me sad.

Theoretically you could try to throw a custom annotation into the input and then check what the target of the anno got transformed into at the end of the flow, and then check whether that also shows up here in the XDC output file. Alternatively, some form of metadata could be useful, like the ability to assign a tag to something in the IR and then at the end of the compilation asking the compiler to dig up the concrete final path that tag landed on. But that's basically what the annotations already do.

sim/src/main/scala/midasexamples/CustomConstraints.scala
```diff
@@ -0,0 +1,25 @@
+//See LICENSE for license details.
+
+package firesim.midasexamples
+
+import chisel3._
+import freechips.rocketchip.config.Parameters
+import midas.targetutils.xdc._
+
+class CustomConstraintsDUT extends Module {
```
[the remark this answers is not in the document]
fabianschuiki: Yeah good point. Let me do that.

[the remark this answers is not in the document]
fabianschuiki: Done

sim/src/test/scala/midasexamples/TutorialSuite.scala
fabianschuiki: Done

fabianschuiki:
Applied your suggestions @davidbiancolin, thanks a lot :+1:

sim/src/test/scala/midasexamples/TutorialSuite.scala
davidbiancolin: Yeah... While not perfect, I think both of those ideas are pretty reasonable. 

In the short term you can probably just wildcard the middle component of the path? The pass-level test in GoldenGate can check for complete path correctness.

fabianschuiki: Yeah I like that approach. We only care about the prefix and suffix of the path. Let me tweak this.

fabianschuiki: Finally figured out how to do this. So many DSLs :see_no_evil:

davidbiancolin (commented):
This looks good to me, but need to hold off merging until Sagar cuts a release for the FireSim tutorial.

fabianschuiki:
Sounds great, thanks for the reviews and suggestions :tada:!

davidbiancolin (approved):

PR #1262, opened 2022-10-13
Adding support for Azure CI

t14916:
Added support for Azure CI and refactored much of the library code in the CI system to better support multiple platforms.

The one issue with this PR: there are way too many secrets right now to support running Azure because of the number of env variables needed (you can do it, it just looks annoying). I talked it over with David, eventually I'll look into implementing a workflow generation script to generate a workflow based off a template, and in the script populate these environment variables that are stored in the script / in a config file. For now, I'll comment these out in this PR but I'll fix it when time permits. The current changes still have a lot of QOL which makes the code more readable, gets rid of enviornment variables in a bunch of places, and adds type annotations.

#### Related PRs / Issues

N/A

#### UI / API Impact

N/A

N/A

### Contributor Checklist


.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+
+from typing import Any, Callable, Dict, List
+
+# Reuse manager utilities
+# Note: ci_firesim_dir must not be used here because the persistent clone my not be initialized yet.
+sys.path.append(local_fsim_dir + "/deploy")
+from awstools.awstools import get_instances_with_filter
+
+# This tag is common to all instances launched as part of a given workflow
+unique_tag_key = 'ci_workflow_id'
```
abejgonzalez, line 25: I've always not liked the naming of this var. Can we rename it to something like "workflow_id_tag"?

.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+    Instances - a more general term, but can refer to runfarm / buildfarm instances run by CI as well
+    """
+
+    @abc.abstractmethod
+    def get_filter(self, tag_value: str) -> Dict:
+        """ Returns a filter that returns all instances associated with workflow """        
+        raise NotImplementedError
+    
+    @abc.abstractmethod
+    def get_manager_tag_dict(self, sha: str, tag_value: str) -> Dict[str, str]:
```
abejgonzalez, line 64: I would change the `tag_value` input name to be more descriptive (it will always be the workflow ID).

.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+    Note that there are 2 terms used to refer to machines running CI: Managers and Instances
+
+    Managers - refer to a manager instance, this only really matters for AWS, but it's the main machine
+    running CI
+
+    Instances - a more general term, but can refer to runfarm / buildfarm instances run by CI as well
+    """
+
+    @abc.abstractmethod
+    def get_filter(self, tag_value: str) -> Dict:
```
abejgonzalez, line 59: Similar to the other method with `tag_value`, I think it should be renamed to `workflow_id`

.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def get_platform_enum(self) -> None:
+        return Platform.AWS
+
+    def get_manager_metadata_string(self, tag_value: str) -> str:
+        return self.instance_metadata_str(self.find_manager(tag_value))
+
+    def instance_metadata_str(self, instance) -> str:
+        """ Pretty prints instance info, including ID, state, and public IP """
+        static_md = """    Instance ID: {}
```
abejgonzalez, line 222: Can you switch to `f""` string formatting.

abejgonzalez: And do this everywhere (in new code).

.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+        else:
+            return None
+
+    def find_all_workflow_instances(self, tag_value: str) -> List:
+        """ Grabs a list of all instance dicts sharing the CI workflow run's unique tag """
+        return get_instances_with_filter([self.get_filter(tag_value)])
+
+    def find_all_ci_instances(self) -> List:
+        """ Grabs a list of all instances across all CI using the CI unique tag key"""
+        all_ci_instances_filter = {'Name': 'tag:' + unique_tag_key, 'Values' : ['*']}
```
abejgonzalez, line 160: You can just reuse the `get_filter` here with `"*"` arg.

.github/scripts/platform_lib.py
abejgonzalez: Im sensing this trend with all the `tag_value` arg. names. Is there a case where the `tag_value` would not be the workflow ID?

.github/scripts/platform_lib.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def get_manager_ip(self, tag_value: str) -> str:
+        """
+        Looks up the AWS manager IP using the CI workflow run's unique tag
+        """
+        aws_manager = self.find_manager(tag_value)
+        if aws_manager is None:
+            raise Exception("No AWS manager instance running with tag matching the assigned workflow id\n")
+    
+        return aws_manager['PublicIpAddress']
```
abejgonzalez, line 172: IIRC this value is only populated if the instance is properly started. Do you ensure the instance is properly running somewhere else? Or do you validate the IP is a valid IP somewhere else?

.github/scripts/setup-workflow-monitor.py
```diff
@@ -4,28 +4,48 @@
...
         instances should have before it is stopped. This serves as a redundant check
         in case the workflow-monitor is brought down for some reason.
+    
+    platform: Enum that indicates either 'aws' or 'azure' currently. Describes the current platform 
+        from which CI is being run from.
     """
     with cd(manager_ci_dir):
+        # This generates a file that can be sourced to get all the right keys / ids to run any of the
+        # Azure jobs. This is always generated and sourced, but is completely unused by the AWS platform.
+        generate_azure_credited_env()
```
abejgonzalez, line 31: Can you just use a `with prefix(...)` call around the `run` to setup the right environment? Wouldn't that then get passed to the subshell that runs the screen (removing the need for a file)?

.github/scripts/setup-workflow-monitor.py
```diff
@@ -4,28 +4,48 @@
...
         # Instances will be terminated (since they are spot requests) and cleaned up in a nightly job.
 
         # Setting pty=False is required to stop the screen from being
         # culled when the SSH session associated with the run command ends.
-        run("screen -S ttl -dm bash -c \'sleep {}; ./change-workflow-instance-states.py {} terminate {}\'"
-            .format(int(max_runtime) * 3600, ci_workflow_run_id, ci_personal_api_token), pty=False)
-        run("screen -S workflow-monitor -L -dm ./workflow-monitor.py {} {}"
-            .format(ci_workflow_run_id, ci_personal_api_token), pty=False)
 
+        run("echo 'zombie kr' >> ~/.screenrc") # for testing purposes, keep the screen on even after it dies
```
abejgonzalez, line 39: Can you elaborate on why this was needed in the past?

.github/scripts/setup-workflow-monitor.py
```diff
@@ -4,28 +4,48 @@
...
+        run("screen -S ttl -dm bash -c \'source ~/azure_env.sh; sleep {}; ./change-workflow-instance-states.py {} {} terminate {}\'"
+            .format(int(max_runtime) * 3600, platform, ci_workflow_run_id, ci_personal_api_token), pty=False)
+        run("screen -S workflow-monitor -L -dm bash -c \'source ~/azure_env.sh; ./workflow-monitor.py {} {} {}\'"
+            .format(platform, ci_workflow_run_id, ci_personal_api_token), pty=False)
 
 if __name__ == "__main__":
     set_fabric_firesim_pem()
-    max_runtime = sys.argv[1]
-    execute(setup_workflow_monitor, max_runtime, hosts=["localhost"])
+    platform = get_platform_enum(sys.argv[1])
```
abejgonzalez, line 47: You add asserts in other files for number of args, can you add it here as well.

.github/scripts/workflow-monitor.py
```diff
@@ -17,10 +17,10 @@
 import sys
 import requests
 
-from common import terminate_workflow_instances, stop_workflow_instances, gha_runs_api_url
-
+from common import get_platform_lib, gha_runs_api_url
+from platform_lib import Platform, get_platform_enum
 # Time between HTTPS requests to github
-POLLING_INTERVAL_SECONDS = 60
+POLLING_INTERVAL_SECONDS = 5
```
abejgonzalez, line 23: Just checking, does this have any noticeable impact on CI performance? It shouldn't but we were very conservative in the past.

.github/scripts/change-workflow-instance-states.py
```diff
@@ -6,17 +6,24 @@
...
     parser.add_argument('state_change',
                        choices = ['terminate', 'stop', 'start'],
                        help = "The state transition to initiate on workflow instances.")
     parser.add_argument('github_api_token',
                        help = "API token to modify self-hosted runner state.")
-
     args = parser.parse_args()
-    change_workflow_instance_states(args.github_api_token, args.tag_value, args.state_change)
+    platform = get_platform_enum(args.platform)
+    if platform == Platform.AWS or platform == Platform.ALL:
```
abejgonzalez, line 26: When do you expect to use `Platform.ALL`? I wonder if we can just remove this functionality / potential confusion.

.github/scripts/change-workflow-instance-states.py
```diff
@@ -6,17 +6,24 @@
...
 import argparse
 
-from common import change_workflow_instance_states
+from platform_lib import Platform, get_platform_enum
+from common import get_platform_lib
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
+    parser.add_argument('platform',
+                        choices = ['aws', 'azure', 'all'],
```
abejgonzalez, line 15: Nit: This choice list should be derived from a list given in the `platform_lib`

.github/scripts/common.py
```diff
@@ -37,60 +31,14 @@ def set_fabric_firesim_pem():
     env.key_filename = manager_fsim_pem
 
 # This tag is common to all instances launched as part of a given workflow
-unique_tag_key = 'ci-workflow-id'
+unique_tag_key = 'ci_workflow_id'
 
 # Produce a filter that return all instances associated with this workflow
 def get_ci_filter(tag_value):
     return {'Name': 'tag:' + unique_tag_key, 'Values' : [tag_value]}
```
abejgonzalez, line 38: Should this be deleted?

.github/scripts/setup-workflow-monitor.py
t14916: When I tested this in CI, running exports / running with a prefix did not result in the screen job inheriting the environment variables, which is unfortunately the only (feasible) way for us internally to pass credentials to CI.

.github/scripts/change-workflow-instance-states.py
t14916: Internally I use it in my `setup-self-hosted-runner` job to catch the failed instances (since I setup both the Azure / AWS runner in the same step in the workflow). This just prevents duplication of the that step in the job. Ostensibly I can also just do this in 2 separate steps to avoid this, but that leads to a lot more copy-pasta. There are some small things like this that can probably be cleaned up after I get around to workflow generation like you suggested, but I need to get that working first.

.github/scripts/platform_lib.py
t14916: refactored it to `workflow_tag`, since technically the `workflow_id` will have either `aws` and `azure` prepended to it now.

.github/scripts/platform_lib.py
t14916: I didn't change any of the logic for AWS, and I don't think we ever checked before. We usually print out the metadata information, including public ip, at launch, and it is usually populated there (although the print code does handle the case where it isn't). If you feel strongly about this, however, I can add an error check here. This case should never occur for Azure however.

scripts/machine-launch-script.sh
```diff
@@ -257,8 +257,13 @@ set -o pipefail
     # prefer creating the environment with a single invocation of
     # conda
     PIP_PKGS=( \
-        fab-classic \
-        mypy-boto3-ec2==1.12.9 \
+        fab-classic==1.19.1 \
+        mypy-boto3-ec2==1.21.9 \
```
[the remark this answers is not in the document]
abejgonzalez: We want to move to unpinning as much as possible (and having the conda lockfile determining pinned versions) so this is fine as long as the mypy tests pass.

abejgonzalez (approved):
LGTM!

PR #1274, opened 2022-10-18
Put instPath before label in AutoCounter output

timsnyder-siv:
In a design with many autocounters being output, it is often the case that the instPath can be:
* quite long
* have common substrings across many of the counters

To simplify post-processing, it is strongly preferred that the instPath is first so that common prefix algorithms can be used to shorten the autocounter labels.

I think @davidbiancolin has other people who have complained about the consistency of name ordering in other places as well.

#### UI / API Impact

This will change the naming of AutoCounters such that the instPath will come at the beginning of the names, followed by the label.  Downstream AutoCounter post-processing scripts will need to be updated if they manipulate the names with the old assumptions.

No impact to Verilog or AGFI.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/scala/midas/passes/AutoCounterTransform.scala
+++ b/sim/midas/src/main/scala/midas/passes/AutoCounterTransform.scala
@@ -183,7 +183,7 @@ class AutoCounterTransform extends Transform with AutoCounterConsts {
         val eventWidth = portWidthMap(anno.topSink.ref)
         EventMetadata(
           anno.topSink.ref,
-          (pathlessLabel +: instPath).mkString("_"),
+          (instPath :+ pathlessLabel).mkString("_"),
           autoCounterAnno.description,
           eventWidth,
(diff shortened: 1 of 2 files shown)


davidbiancolin:
@nandor IIRC you had an opinion about this right?

davidbiancolin:
Alternatively we could put the path information in a new column/columns, but that would be more work

davidbiancolin:
You're going to need to update the tests, here: https://github.com/firesim/firesim/blob/fabb54bdaba8de68d5850a4bfb4cc2d048c2dba6/sim/src/main/scala/midasexamples/AutoCounterValidator.scala#L133

So the reference output matches the new name encoding.

timsnyder-siv:
Thanks for the pointer to the tests. I'll clean them up.

abejgonzalez (approved):
SGTM

davidbiancolin (approved):

PR #1269, opened 2022-10-18
Add PyTest docs

abejgonzalez:
First pass on pytest documentation.

### Contributor Checklist
- [ ] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [ ] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous prints/debugging code?
- [ ] Did you state the UI / API impact?
- [ ] Did you specify the Verilog / AGFI compatibility impact?

- [ ] If applicable, did you regenerate and publicly share default AGFIs?

- [ ] If applicable, did you apply the `ci:fpga-deploy` label?

- [ ] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
Note: to run CI on PRs from forks, comment `@Mergifyio copy main` and manage the change from the new PR.
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1299, opened 2022-11-12
Changed "firesim infrasetup" to deploy using a tarball. 

sifive-benjamin-morse:
#### UI / API Impact
* The `firesim infrasetup` stage now uses a tarball to move the driver files into `sim_slot_x`.
* In addition if `driver_tar_uri` key is provided in `config_hwdb.yaml` the tarball found at that path is used, and the driver build is skipped and the found tarball is used

* no change

### Contributor Checklist


russell-horvath:
@sifive-benjamin-morse Github won't let me comment on the submodule  changes but it looks like you pushed something to `aws-fpga` did you mean to do that? Cause I am not able to see the changes if so.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -474,9 +474,37 @@ def get_sim_kill_command(self, slotno: int) -> str:
         """
         return self.get_resolved_server_hardware_config().get_kill_simulation_command()
 
+    def get_tarball_files_paths(self) -> List[Tuple[str, str]]:
+        """ Return local and remote paths of all stuff destined for the driver tarball 
+        The returned paths in the tuple are [local_path, remote_path]. """
+        all_paths = []
+
+        driver_path = self.get_resolved_server_hardware_config().get_local_driver_path()
+        all_paths.append((driver_path, ''))
```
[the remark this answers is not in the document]
sifive-benjamin-morse: The empty string thing will put the file with the same name on the remote. This is a feature of rsync. as far as documentation I can add a comment about it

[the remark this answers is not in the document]
davidbiancolin: Yeah if you could do that in the docstring that would be great.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -148,6 +148,11 @@ def disconnect_all_nbds_instance(self) -> None:
 
                 run("; ".join(fullcmd))
 
+    def get_remote_sim_dir_for_slot(self, slotno: int) -> str:
```
davidbiancolin, line 151: docstring

deploy/runtools/firesim_topology_elements.py
```diff
@@ -474,9 +474,37 @@ def get_sim_kill_command(self, slotno: int) -> str:
...
+        all_paths += self.get_resolved_server_hardware_config().get_additional_required_sim_files()
+
+        all_paths += self.get_job().get_siminputs()
+        return all_paths
+    
+    def get_tarball_path_pair(self) -> List[Tuple[str, str]]:
+        """ Return local and remote paths of the actual driver tarball, not files inside"""
+        all_paths = []
+
+        if not isinstance(self.server_hardware_config, str) and self.server_hardware_config is not None and self.server_hardware_config.driver_tar_uri is not None:
```
davidbiancolin, line 497: This might be a question for @abejgonzalez , but can you explain is this `isinstance(self.server_hardware_config, str)` here.  Under what conditions can it be a string?

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -172,6 +175,18 @@ def copy_sim_slot_infrastructure(self, slotno: int) -> None:
...
+    def extract_driver_tarball(self, slotno: int) -> None:
+        """ extract tarball that already exists on the remote node. """
+        if self.instance_assigned_simulations():
+            assert slotno < len(self.parent_node.sim_slots)
+            serv = self.parent_node.sim_slots[slotno]
+
+            remote_sim_dir = self.get_remote_sim_dir_for_slot(slotno)
+
+            options = "-zxf"
+            cmd = f"cd {remote_sim_dir} && tar {options} {serv.get_tar_name()}"
```
davidbiancolin, line 187: So for `cd` fabric provides a context manager you can use instead. And then, nit: I'd then just put the run command inline 
i.e., `run(f"tar {options} {serv.get_tar_name()")` as it is done generally throughout the codebase. I find it more readable because of the consistency, but i'll let you decide.

deploy/runtools/runtime_config.py
```diff
@@ -117,6 +123,22 @@ def get_local_driver_path(self) -> str:
...
 
+    def tarball_creation_foldername(self) -> str:
+        """ return the name of the folder that the tarball is created from, no path """
+        return 'tarball'
+
+    def local_tarball_creation_path(self) -> str:
+        """ return the absolute path of the folder used to create the tarball """
+        triplet = self.get_deploytriplet_for_config()
+        dirname = self.tarball_creation_foldername()
+        return f'{get_deploy_dir()}/../sim/output/f1/{triplet}/{dirname}'
```
davidbiancolin, line 134: By your estimation, how hard would it be to start using pathlib for these?

deploy/runtools/runtime_config.py
```diff
@@ -257,14 +285,53 @@ def build_sim_driver(self) -> None:
...
+        must run before this function.  """
+        if self.tarball_built:
+            # we already built it
+            return
+
+        builddir = self.local_tarball_creation_path()
+
+        with InfoStreamLogger('stdout'):
+            cmd = f"rm -rf {builddir}"
+
```
davidbiancolin, line 303: I'm a little wary of putting `rm -rfs` in the manager. What if builddir path get's mangled by an errant commit?

davidbiancolin (commented):
Mostly nits aside from one question: is my understanding correct that you're copying the files to the remote and then tarring there? 

As an intermediate step it seems to me it would make more sense to tar things up locally. That would open the door to having a tar-only step without needing the remote.

Could you add pytests for this feature? Frankly, i'm not totally sure how this works now, but abe created a skeleton docs page here: https://github.com/firesim/firesim/pull/1269/files but apparenlty that's not getting published correctly.

deploy/runtools/runtime_config.py
sifive-benjamin-morse: I can try should be easy

deploy/runtools/runtime_config.py
sifive-benjamin-morse: `builddir` gets created and then all of the rsync (source, dest) pairs run and copy files into it. If files are removed from source, there is no way to catch that using rsync. The will remain at dest.  There is a flag (`--delete                 delete extraneous files from dest dirs`) but I believe this won't work because there are multiple pairs. The second pair would delete all the desired files from the first run.

If you believe files being removed from source and abandoned in dest isn't an issue, then I can remove this. If there is some other trick to delete these files in a more clean way please suggest

deploy/runtools/run_farm_deploy_managers.py
sifive-benjamin-morse: I am taring things locally. If I may ask, which part led you to think that the tar was being done remotely ?

deploy/runtools/runtime_config.py
```diff
@@ -257,14 +285,53 @@ def build_sim_driver(self) -> None:
...
+        with InfoStreamLogger('stdout'), prefix(f'cd {get_deploy_dir()}'), \
+            prefix(f'mkdir -p {builddir}'):
+            for local_path, remote_path in paths:
+                # The `rsync_project()` function does not allow
+                # copying between two local directories.
+                # This uses the same option flags but operates rsync in local->local mode
+                options = '-pthrvz -L'
+                local_dir=local_path
+                remote_dir=pjoin(builddir, remote_path)
+                cmd = "rsync %s %s %s" % (options, local_dir, remote_dir)
```
davidbiancolin, line 316:
```suggestion
                cmd = f"rsync {options} {local_dir} {remote_dir}"
```

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -553,6 +553,19 @@ to the relative name of the config. For example,
 ``custom_runtime_config: GREATCONFIG.conf``.
 
 
+``driver_tar_uri``
+"""""""""""""""""""""""""""""
+
+This is an advanced feature - under normal conditions, omit this key.
```
davidbiancolin, line 559:
```suggestion
```

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -553,6 +553,19 @@ to the relative name of the config. For example,
...
+When this key is present, the local driver will not build from source.
+Instead, during the `firesim infrasetup`, this `.tar.gz` file will be
+deployed and extracted into the `sim_slot_X` folder. The purpose of this
+feature is to enable advanced CI configurations where the driver build step
```
davidbiancolin, line 563: I think you can sell this feature a little harder, because it's cool and very useful: 

This feature allows sharing simulator drivers alongside their bitstreams (AGFIs or XCLBINs) in a HWDB entry. This eases sharing simulators with other users by avoiding both long rebuilds of the driver and potential driver-bitstream incompatibilities.

deploy/runtools/runtime_config.py
```diff
@@ -117,6 +123,22 @@ def get_local_driver_path(self) -> str:
...
+        """ return the absolute path of the folder used to create the tarball """
+        triplet = self.get_deploytriplet_for_config()
+        dirname = self.tarball_creation_foldername()
+        return f'{get_deploy_dir()}/../sim/output/f1/{triplet}/{dirname}'
+
+    def local_tarball_path(self, name: str) -> str:
+        """ return the local path of the tarball """
+        triplet = self.get_deploytriplet_for_config()
+        dirname = self.tarball_creation_foldername()
+        return f'{get_deploy_dir()}/../sim/output/f1/{triplet}/{name}'
```
davidbiancolin, line 140: The `f1` in this path needs to be the `PLATFORM` environment variable for this to work correctly under vitis.

deploy/runtools/runtime_config.py
davidbiancolin: The f1 in this path needs to be the PLATFORM environment variable for this to work correctly under vitis.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -526,6 +547,10 @@ def get_rootfs_name(self) -> Optional[str]:
             # cases
             return self.get_job_name() + "-" + rootfs_path.split("/")[-1]
 
+    def get_tar_name(self) -> str:
+        """ Get the name of the tarball on the run host"""
+        return "driver-bundle.tar.gz"
```
davidbiancolin, line 552: Maybe this should be hoisted to the top of the file and put in all caps, per PEP8. I'll defer to @timsnyder

deploy/runtools/run_farm_deploy_managers.py
davidbiancolin: I didn't read your in-source documentation correctly. 

I think it was the confluence of seeing:
1) `run` commands, which generally imply running on a worker node (as opposed to `local`) and forgetting that we've switched to using `run` + localhost in favor of `local` generally.
2) See rsync + "remote_dir" in places where the remote dir is just another dir under output. Again, if i read the code more carefully that would've been clear.

deploy/runtools/firesim_topology_elements.py
abejgonzalez: In the user topologies, you can assign each server node of a simulation to a specific HW config by giving a string (https://github.com/firesim/firesim/blob/7aa7ee22f0ab91d29340d1842f34acb7952c20bc/deploy/runtools/user_topology.py#L407-L414). This was an old API to let people do heterogeneous simulations with different HW cfgs.

deploy/runtools/runtime_config.py
sifive-benjamin-morse: I've addressed this comment by making a new folder each time. the new folder incorporates the epoc time in seconds as well as a few chars of a uuid. I've removed the `rm -rf`

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -148,6 +148,13 @@ def disconnect_all_nbds_instance(self) -> None:
 
                 run("; ".join(fullcmd))
 
+    def get_remote_sim_dir_for_slot(self, slotno: int) -> str:
+        """ Returns the path on the remote for a given slot number. This function
+        must return a path that includes the trailing slash."""
+        remote_home_dir = self.parent_node.get_sim_dir()
+        remote_sim_dir = """{}/sim_slot_{}/""".format(remote_home_dir, slotno)
```
timsnyder, line 155: Rather than stating the assumption about the return value in the docstring, it makes sense to put in an assertion here.  Also, use formatstrings instead of `.format()`
```suggestion
        """ Returns the path on the remote for a given slot number. """
        
        remote_home_dir = self.parent_node.get_sim_dir()
        remote_sim_dir = f"{remote_home_dir}/sim_slot_{slotno}/"
        assert remote_sim_dir[-1] == '/', 
            f"Return value of get_remote_sim_dir_for_slot({remote_home_dir}, {slotno}) must end with '/'."
        # so that callers can supply the return value to rsync and have well-defined semantics
```

NOTE: I'm not sure that I have legal indenting in my suggestion and I'm just taking a guess at why the trailing slash matters but it would be good to document why we need it, possibly in a comment if the assertion message becomes too unweildly.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -172,6 +177,18 @@ def copy_sim_slot_infrastructure(self, slotno: int) -> None:
...
             run("""cp -r {}/* {}/""".format(remote_sim_rsync_dir, remote_sim_dir), shell=True)
 
+    def extract_driver_tarball(self, slotno: int) -> None:
+        """ extract tarball that already exists on the remote node. """
+        if self.instance_assigned_simulations():
+            assert slotno < len(self.parent_node.sim_slots)
+            serv = self.parent_node.sim_slots[slotno]
+
+            remote_sim_dir = self.get_remote_sim_dir_for_slot(slotno)
+            options = "-zxf"
```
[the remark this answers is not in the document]
timsnyder: I see that you documented a restriction to `.tar.gz`.  So feel free to ignore my comments here.  However, I don't think I saw an assertion about the value ending in `.tar.gz` if it is set.

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
timsnyder: I would argue that the purpose of this is what David said.  The CI implications are nice but I've wanted this for a long time because it is both onerous and fraught to have to rebuild the driver at runtime.

timsnyder (approved):
I think most of my feedback falls into the "have you thought about this?" category.  I didn't see anything in the "whoa, this is totally broken." area.  As such, I'm fine with this being merged as is or if you think any of my feedback is useful, continue to refine and I can re-review.

timsnyder:
Oh, and I'm very excited to see this feature coming.  I've wanted it since I started using FireSim several years ago.

deploy/runtools/runtime_config.py
```diff
@@ -257,14 +292,52 @@ def build_sim_driver(self) -> None:
...
+
+        with InfoStreamLogger('stdout'), prefix(f'cd {builddir}'):
+            findcmd = 'find . -mindepth 1 -maxdepth 1 -printf "%P\n"'
+            taroptions = '-czvf'
+
+            # Running through find and xargs is the most simple way I've found to meet these requirements:
+            #   * create the tar with no leading ./ or foldername
+            #   * capture all types of hidden files (.a ..a .aa)
+            #   * avoid capturing the parent folder (..) with globs looking for hidden files
+            cmd = f"{findcmd} | xargs tar {taroptions} ../{tarball_name}"
```
abejgonzalez, line 335: Same comment as above on inlining `cmd` into `run()`

deploy/runtools/runtime_config.py
```diff
@@ -257,14 +292,52 @@ def build_sim_driver(self) -> None:
...
+                self.handle_failure(results, 'local rsync', get_deploy_dir(), cmd)
+
+        with InfoStreamLogger('stdout'), prefix(f'cd {builddir}'):
+            findcmd = 'find . -mindepth 1 -maxdepth 1 -printf "%P\n"'
+            taroptions = '-czvf'
+
+            # Running through find and xargs is the most simple way I've found to meet these requirements:
+            #   * create the tar with no leading ./ or foldername
+            #   * capture all types of hidden files (.a ..a .aa)
+            #   * avoid capturing the parent folder (..) with globs looking for hidden files
```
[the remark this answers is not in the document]
abejgonzalez: If moving into a new dir. within the `sim_slot` I would also verify that the `LD_LIBRARY_PATH` (or whatever points to the copied libraries) is appropriately set.

abejgonzalez (commented):
No major additions outside of what Tim + David said. SGTM!

deploy/runtools/runtime_config.py
```diff
@@ -257,14 +292,52 @@ def build_sim_driver(self) -> None:
...
+            for local_path, remote_path in paths:
+                # The `rsync_project()` function does not allow
+                # copying between two local directories.
+                # This uses the same option flags but operates rsync in local->local mode
+                options = '-pthrvz -L'
+                local_dir=local_path
+                local_remote_dir=pjoin(builddir, remote_path)
+                cmd = f"rsync {options} {local_dir} {local_remote_dir}"
+
+                results = run(cmd)
```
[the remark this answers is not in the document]
sifive-benjamin-morse: sadly this breaks the clean error reporting provided by handle_failure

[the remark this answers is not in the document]
sifive-benjamin-morse: sadly this breaks the clean error reporting provided by `handle_failure`

deploy/runtools/runtime_config.py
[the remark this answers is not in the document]
sifive-benjamin-morse: The purpose of this was to replace original behavior with disturbing anything. Following this principle, I prefer to leave the tarball extraction as it is, which is as close as possible to the behavior before

davidbiancolin (approved):
LGTM. Nice work @sifive-benjamin-morse

PR #1302, opened 2022-11-15
Fix C++ compiler warnings

nandor:
#### Related PRs / Issues

N/A

#### UI / API Impact

N/A

N/A

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/bridges/synthesized_assertions.cc
+++ b/sim/midas/src/main/cc/bridges/synthesized_assertions.cc
@@ -8,10 +8,8 @@ synthesized_assertions_t::synthesized_assertions_t(
     simif_t *sim,
     std::vector<std::string> &args,
     ASSERTBRIDGEMODULE_struct *mmio_addrs,
-    unsigned int num_asserts,
     const char *const *msgs)
-    : bridge_driver_t(sim), mmio_addrs(mmio_addrs), num_asserts(num_asserts),
-      msgs(msgs) {
+    : bridge_driver_t(sim), mmio_addrs(mmio_addrs), msgs(msgs) {
   for (auto &arg : args) {
     if (arg.find("+disable-asserts") == 0)
       enable = false;
--- a/sim/midas/src/main/cc/simif_emul.cc
+++ b/sim/midas/src/main/cc/simif_emul.cc
@@ -30,6 +30,14 @@ double sc_time_stamp() { return (double)main_time; }
 extern void tick();
 #endif // VCS
 
+mmio_t *simif_emul_t::master = new mmio_t(CTRL_BEAT_BYTES);
+mmio_t *simif_emul_t::cpu_managed_axi4 =
+    new mmio_t(CPU_MANAGED_AXI4_BEAT_BYTES);
+
+mm_t *simif_emul_t::slave[MEM_NUM_CHANNELS] = {nullptr};
+
+mm_t *simif_emul_t::cpu_mem = new mm_magic_t;
+
 void finish() {
 #ifdef VCS
   vcs_fin = true;
@@ -239,8 +247,13 @@ simif_emul_t::cpu_managed_axi4_write(size_t addr, char *data, size_t size) {
   size_t strb[len + 1];
   size_t *strb_ptr = &strb[0];
 
-  for (int i = 0; i < len; i++)
+  for (int i = 0; i < len; i++) {
+#if CPU_MANAGED_AXI4_BEAT_BYTES == 64
+    strb[i] = -1;
+#else
     strb[i] = (1LL << CPU_MANAGED_AXI4_BEAT_BYTES) - 1;
+#endif
+  }
 
   if (remaining == CPU_MANAGED_AXI4_BEAT_BYTES)
     strb[len] = strb[0];
--- a/sim/midas/src/main/cc/simif_emul.h
+++ b/sim/midas/src/main/cc/simif_emul.h
@@ -36,9 +36,8 @@ public:
    *
    * These have external linkage to enable VCS to easily access them.
    */
-  inline static mmio_t *master = new mmio_t(CTRL_BEAT_BYTES);
-  inline static mmio_t *cpu_managed_axi4 =
-      new mmio_t(CPU_MANAGED_AXI4_BEAT_BYTES);
+  static mmio_t *master;
+  static mmio_t *cpu_managed_axi4;
   /**
    * @brief Host DRAM models shared across the RTL simulator and driver
    * contexts.
@@ -48,7 +47,7 @@ public:
    * writes directly into the backing memory (0 cycles). See
    * simif_emul_t::load_mems.
    */
-  inline static mm_t *slave[MEM_NUM_CHANNELS] = {nullptr};
+  static mm_t *slave[MEM_NUM_CHANNELS];
   /**
    * @brief A model of FPGA-addressable CPU-host memory.
    *
@@ -56,7 +55,7 @@ public:
    * AXI4 memory subordinate as a proxy for writing into actual host-CPU DRAM.
    * The driver-side of FPGAManagedStreams inspect circular buffers hosted here.
    */
-  inline static mm_t *cpu_mem = new mm_magic_t;
+  static mm_t *cpu_mem;
 
 private:
   // The maximum number of cycles the RTL simulator can advance before
--- a/sim/firesim-lib/src/main/cc/bridges/blockdev.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/blockdev.cc
@@ -268,7 +268,6 @@ void blockdev_t::handle_data(struct blkdev_data &data) {
   }
 
   struct blkdev_write_tracker &tracker = write_trackers[data.tag];
-  struct blkdev_data resp;
 
   /* Copy data into the write tracker */
   tracker.data[tracker.count] = data.data;
--- a/sim/firesim-lib/src/main/cc/bridges/blockdev.h
+++ b/sim/firesim-lib/src/main/cc/bridges/blockdev.h
@@ -56,18 +56,11 @@ public:
 
 private:
   BLOCKDEVBRIDGEMODULE_struct *mmio_addrs;
-  bool a_req_valid;
-  bool a_req_ready;
-  bool a_data_valid;
-  bool a_data_ready;
-  bool a_resp_valid;
-  bool a_resp_ready;
 
   // Set if, on the previous tick, we couldn't write back all of our response
   // data
   bool resp_data_pending = false;
 
-  simif_t *sim;
   uint32_t _ntags;
   uint32_t _nsectors;
   FILE *_file, *logfile;
--- a/sim/firesim-lib/src/main/cc/bridges/simplenic.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/simplenic.cc
@@ -57,9 +57,7 @@ simplenic_t::simplenic_t(simif_t *sim,
                          const int stream_from_cpu_idx,
                          const int stream_from_cpu_depth)
     : bridge_driver_t(sim), stream_to_cpu_idx(stream_to_cpu_idx),
-      stream_to_cpu_depth(stream_to_cpu_depth),
-      stream_from_cpu_idx(stream_from_cpu_idx),
-      stream_from_cpu_depth(stream_from_cpu_depth) {
+      stream_from_cpu_idx(stream_from_cpu_idx) {
   this->mmio_addrs = mmio_addrs;
 
   const char *niclogfile = NULL;
@@ -91,11 +89,10 @@ simplenic_t::simplenic_t(simif_t *sim,
       this->loopback = true;
     }
     if (arg.find(macaddr_arg) == 0) {
-      uint8_t mac_bytes[6];
       int mac_octets[6];
       char *macstring = NULL;
       macstring = const_cast<char *>(arg.c_str()) + macaddr_arg.length();
-      char *trailingjunk;
+      char trailingjunk;
 
       // convert mac address from string to 48 bit int
       if (6 == sscanf(macstring,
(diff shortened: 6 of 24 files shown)


sim/midas/src/main/cc/simif_emul.cc
```diff
@@ -30,6 +30,14 @@ double sc_time_stamp() { return (double)main_time; }
 extern void tick();
 #endif // VCS
 
+mmio_t *simif_emul_t::master = new mmio_t(CTRL_BEAT_BYTES);
```
davidbiancolin, line 33: So did you have to define these because verilator is using a C++ std older than c++17?  The next verilator version we're pulling in should fix that.

davidbiancolin (commented):
Thanks Nandor, just the one comment.

sim/midas/src/main/cc/simif_emul.h
```diff
@@ -36,9 +36,8 @@ class simif_emul_t : public virtual simif_t {
    *
    * These have external linkage to enable VCS to easily access them.
    */
-  inline static mmio_t *master = new mmio_t(CTRL_BEAT_BYTES);
-  inline static mmio_t *cpu_managed_axi4 =
-      new mmio_t(CPU_MANAGED_AXI4_BEAT_BYTES);
+  static mmio_t *master;
+  static mmio_t *cpu_managed_axi4;
```
timsnyder-siv, line 40: Please explain why you want to remove `inline` from these static members.

timsnyder-siv (commented):
If this fixes C++ compiler warnings, can we turn on `-Werror` now?  I think this PR needs some more explanation in places.

sim/midas/src/main/cc/simif_emul.h
davidbiancolin: This is related to the C++17 comment i made.

davidbiancolin: I think verilator is defaulting to C++14, which doesn't permit using this construct, but things still work fine anyways. I'll let @nandor confirm/deny.

nandor: While this is a C++17 feature, I never saw it use. Inline is pretty bad since it relies on weak symbols for its implementation, which aren't standardized and are full of surprises. In general, in all such situations its better to define variables in a single compilation unit, outside the class declaration.

sim/midas/src/main/cc/bridges/synthesized_assertions.cc
```diff
@@ -10,8 +10,7 @@ synthesized_assertions_t::synthesized_assertions_t(
     ASSERTBRIDGEMODULE_struct *mmio_addrs,
     unsigned int num_asserts,
```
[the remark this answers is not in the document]
nandor: It might be used in the constructor itself.

sim/midas/src/main/cc/simif_emul.cc
nandor: I consider this to be a C++ anti-pattern: if you define static vars in a header, you get weak copies of it in all compilation units including the header and the linker has to de-duplicate. This leads to all sort of fun problems if you accidentally compile different compilation units with different flags.

sim/midas/src/main/cc/simif_emul.h
timsnyder-siv: @davidbiancolin I thought we looked at this and explicitly decided to use `inline static` and switch to C++17 because we thought it was the way to have a singleton defined in one place that would be initialized in a thread safe way.

If Verilator is able to build this code with C++14, I'm fairly confident that we don't want to be using that because it isn't guaranteed to have the C++17 standard behavior.

I also think that if you go back to older-style static members (which Nandor has done) that you should somehow guarantee that the initialization is thread safe.

From what I've read, `static inline` doesn't rely on weak symbols.  It's different than using only `inline`.

sim/midas/src/main/cc/bridges/synthesized_assertions.cc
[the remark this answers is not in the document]
timsnyder-siv: I don't see a use after your deletion of the member.  I think this should be cleaned up.

sim/midas/src/main/cc/simif_emul.h
nandor: Initializers for static data members are called before the `main` of the program and the compiler emits guards to ensure only one initializer is executed at any time. Thread safety guarantees remain unchanged.

This code:
```
class test {
  static inline b *field = new b();
};
```

Lowers to:
```
	.type	_ZN4test5fieldE,@object         # @_ZN4test5fieldE
	.section	.bss._ZN4test5fieldE,"aGw",@nobits,_ZN4test5fieldE,comdat
	.weak	_ZN4test5fieldE
	.p2align	3
_ZN4test5fieldE:
	.quad	0
	.size	_ZN4test5fieldE, 8

	.type	_ZGVN4test5fieldE,@object       # @_ZGVN4test5fieldE
	.section	.bss._ZGVN4test5fieldE,"aGw",@nobits,_ZN4test5fieldE,comdat
	.weak	_ZGVN4test5fieldE
	.p2align	3
_ZGVN4test5fieldE:
	.quad	0                               # 0x0
	.size	_ZGVN4test5fieldE, 8
```

Weak symbols are required for the de-duplication of these data items.  

Why do we need initialization guards for those fields? In my opinion, those should not be static data members in the first place. I think that `static` already enforces the initialization of the data member from a constructor that is invoked before the application can create threads (unless the driver does some other weird thing I do not know about).

sim/midas/src/main/cc/bridges/synthesized_assertions.cc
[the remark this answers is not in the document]
nandor: Fixed.

sim/midas/src/main/cc/simif_emul.h
davidbiancolin: For my own elucidation can you share how you got that snippet or a godbolt link? I'd be curious to see how the emission changes for c++14 vs c++17.

nandor: https://godbolt.org/z/7dfch31W5 Absolutely no difference, C++17 behaviour was backported to C++14 with a warning.

davidbiancolin:
Ok, this LGTM. But please add @sifive-benjamin-morse as a reviewer on driver changes moving forward.

davidbiancolin (approved):

sim/midas/src/main/cc/simif_emul.h
timsnyder-siv: I see now how clang does use weak symbols and gcc uses a GNU extension for 'global unique symbols' to implement `inline static`.  For both compilers, the symbol table is the same whether you use c++14 or c++17.  It does seem to be toolchain dependent and probably something to avoid since Nandor thinks weak symbols are fraught.


<details> 
  <summary> What I did to convince myself... </summary>
I tweaked test.cpp from Nandor's godbolt link so that it has an implemented constructor for `b`:

```c++
class b {};

class test {
  inline static b *field = new b();
};
```

Compiling it with clang++ 15 shows that the `inline static b * field` is a weak symbol as indicated by the `V` type in the middle column:

```
(clang) bash-4.2$ clang++ -v
clang version 15.0.5 (https://github.com/conda-forge/clangdev-feedstock 624157057efdcbe4d8985eaf3e472522f67c9d19)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/centos/miniforge3/envs/clang/bin
Found candidate GCC installation: /home/centos/miniforge3/envs/clang/bin/../lib/gcc/x86_64-conda-linux-gnu/12.2.0
Selected GCC installation: /home/centos/miniforge3/envs/clang/bin/../lib/gcc/x86_64-conda-linux-gnu/12.2.0
Candidate multilib: .;@m64
Selected multilib: .;@m64
(clang) bash-4.2$ clang++ -c test.cpp
test.cpp:4:3: warning: inline variables are a C++17 extension [-Wc++17-extensions]
  inline static b *field = new b();
  ^
1 warning generated.
(clang) bash-4.2$ nm -C test.o
                 U __cxa_guard_abort
                 U __cxa_guard_acquire
                 U __cxa_guard_release
0000000000000000 t __cxx_global_var_init
0000000000000000 V DW.ref.__gxx_personality_v0
0000000000000000 r GCC_except_table0
                 U __gxx_personality_v0
                 U _Unwind_Resume
0000000000000000 V guard variable for test::field
0000000000000000 V test::field
                 U operator new(unsigned long)
(clang) bash-4.2$ clang++ --std=c++17 -c test.cpp
(clang) bash-4.2$ nm -C test.o
                 U __cxa_guard_abort
                 U __cxa_guard_acquire
                 U __cxa_guard_release
0000000000000000 t __cxx_global_var_init
0000000000000000 V DW.ref.__gxx_personality_v0
0000000000000000 r GCC_except_table0
                 U __gxx_personality_v0
                 U _Unwind_Resume
0000000000000000 V guard variable for test::field
0000000000000000 V test::field
                 U operator new(unsigned long)
```

Building it with g++ 12, nm says `test::field` is marked `u` :

```
bash-4.2$ g++ -v
Reading specs from /home/centos/src/project_data/federation_221014/firesim/.conda-env/bin/../lib/gcc/x86_64-conda-linux-gnu/12.1.0/specs
Reading specs from /home/centos/src/project_data/federation_221014/firesim/.conda-env/bin/../lib/gcc/x86_64-conda-linux-gnu/12.1.0/conda.specs
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/home/centos/src/project_data/federation_221014/firesim/.conda-env/bin/../libexec/gcc/x86_64-conda-linux-gnu/12.1.0/lto-wrapper
Target: x86_64-conda-linux-gnu
Configured with: ../configure --prefix=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho --with-slibdir=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib --libdir=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib --mandir=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/man --build=x86_64-conda-linux-gnu --host=x86_64-conda-linux-gnu --target=x86_64-conda-linux-gnu --enable-default-pie --enable-languages=c,c++,fortran,objc,obj-c++ --enable-__cxa_atexit --disable-libmudflap --enable-libgomp --disable-libssp --enable-libquadmath --enable-libquadmath-support --enable-libsanitizer --enable-lto --enable-threads=posix --enable-target-optspace --enable-plugin --enable-gold --disable-nls --disable-bootstrap --disable-multilib --enable-long-long --with-sysroot=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/x86_64-conda-linux-gnu/sysroot --with-build-sysroot=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_build_env/x86_64-conda-linux-gnu/sysroot --with-gxx-include-dir=/home/conda/feedstock_root/build_artifacts/gcc_compilers_1652324151713/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/x86_64-conda-linux-gnu/include/c++/12.1.0 --with-pkgversion='conda-forge gcc 12.1.0-16' --with-bugurl=https://github.com/conda-forge/ctng-compilers-feedstock/issues/new/choose
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 12.1.0 (conda-forge gcc 12.1.0-16) 
bash-4.2$ g++ -c test.cpp
bash-4.2$ nm -C test.o
                 U _GLOBAL_OFFSET_TABLE_
0000000000000049 t _GLOBAL__sub_I_test.cpp
0000000000000000 t __static_initialization_and_destruction_0(int, int)
0000000000000000 u guard variable for test::field
0000000000000000 u test::field
                 U operator new(unsigned long)
bash-4.2$ g++ -c --std=c++17 test.cpp
bash-4.2$ nm -C test.o
                 U _GLOBAL_OFFSET_TABLE_
0000000000000049 t _GLOBAL__sub_I_test.cpp
0000000000000000 t __static_initialization_and_destruction_0(int, int)
0000000000000000 u guard variable for test::field
0000000000000000 u test::field
                 U operator new(unsigned long)
bash-4.2$ g++ -c --std=c++14 test.cpp
test.cpp:4:3: warning: inline variables are only available with '-std=c++17' or '-std=gnu++17' [-Wc++17-extensions]
    4 |   inline static b *field = new b();
      |   ^~~~~~
bash-4.2$ nm -C test.o
                 U _GLOBAL_OFFSET_TABLE_
0000000000000049 t _GLOBAL__sub_I_test.cpp
0000000000000000 t __static_initialization_and_destruction_0(int, int)
0000000000000000 u guard variable for test::field
0000000000000000 u test::field
                 U operator new(unsigned long)
```

The man page for GNU `nm` describes `V` and `u` :

```
           "u" The symbol is a unique global symbol.  This is a GNU extension to the standard set of ELF symbol bindings.  For such a symbol
               the dynamic linker will make sure that in the entire process there is just one symbol with this name and type in use.

           "V"
           "v" The symbol is a weak object.  When a weak defined symbol is linked with a normal defined symbol, the normal defined symbol is
               used with no error.  When a weak undefined symbol is linked and the symbol is not defined, the value of the weak symbol becomes
               zero with no error.  On some systems, uppercase indicates that a default value has been specified.
```

</details>


> thread safety

I misremembered my concerns about static initialization order as thread safety and I apologize for being confusing. 
 With the code before or perhaps at some point during #1177, I was concerned about other globals, maybe in `vcs_harness.cpp` that were being initialized from these globals (now static members) but I see that stuff was cleaned up in #1177 and there's no need for us to use `inline static` here.

PR #1308, opened 2022-11-22
Move XDC circuit paths to config

fabianschuiki:
We would like to be able to tell what the pre- and post-link circuit paths of the design are going to be without having to elaborate the `PlatformShim` to get at the corresponding `XDCPathToCircuitAnnotation`.

This commit adds the two new config options `PreLinkCircuitPath` and `PostLinkCircuitPath`, and moves the platform-specific paths out from `F1Shim` and `VitisShim` and into the corresponding `F1Config` and `VitisConfig`, respectively. This allows other parts of the codebase to determine these paths purely from the config. Since the platform- specific shim code now no longer has to specify these paths directly, the `SpecifyXDCCircuitPaths` is moved up into the parent `PlatformShim` class which annotates the paths provided by the config.

#### Related PRs / Issues

None.

#### UI / API Impact

Unaffected.

Unaffected.

### Contributor Checklist
- [x] ~~Did you add at least one test demonstrating the PR?~~

- [x] ~~If applicable, did you regenerate and publicly share default AGFIs?~~

- [x] ~~If applicable, did you apply the `ci:fpga-deploy` label?~~

- [x] ~~If applicable, did you apply the `Please Backport` label?~~

### DIFF
--- a/sim/midas/src/main/scala/midas/platform/PlatformShim.scala
+++ b/sim/midas/src/main/scala/midas/platform/PlatformShim.scala
@@ -12,6 +12,8 @@ import midas.core._
 import midas.passes.fame.{FAMEChannelConnectionAnnotation}
 import midas.widgets.{CStrLit, UInt32, BridgeIOAnnotation}
 import midas.widgets.CppGenerationUtils._
+import midas.targetutils.xdc.SpecifyXDCCircuitPaths
+import midas.{PreLinkCircuitPath, PostLinkCircuitPath}
 
 /**
   * Generates the platform wrapper (which includes most of the chisel-generated
@@ -43,4 +45,9 @@ abstract class PlatformShim(implicit p: Parameters) extends LazyModule()(p) {
 
     top.module.headerConsts map vMacro foreach sb.append
   }
+
+  // Emit a `XDCPathToCircuitAnnotation` with the pre- and post-link circuit
+  // paths provided in the configuration. These locate the design within the
+  // context of the larger FPGA design.
+  SpecifyXDCCircuitPaths(p(PreLinkCircuitPath), p(PostLinkCircuitPath))
 }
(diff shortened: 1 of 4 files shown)


davidbiancolin (commented):
Can you remove now unnecessary `targetutils.xdc` imports where they exist? Otherwise this looks good to me.

sim/midas/src/main/scala/midas/platform/PlatformShim.scala
```diff
@@ -43,4 +45,9 @@ abstract class PlatformShim(implicit p: Parameters) extends LazyModule()(p) {
 
     top.module.headerConsts map vMacro foreach sb.append
   }
+
+  // Emit a `XDCPathToCircuitAnnotation` with the pre- and post-link circuit
+  // paths provided in the configuration. These locate the design within the
+  // context of the larger FPGA design.
+  SpecifyXDCCircuitPaths(p(PreLinkCircuitPath), p(PostLinkCircuitPath))
```
davidbiancolin, line 52: You might as well remove `SpecifyXDCCircuitPaths` too right? Or better yet, add a deprecation warning to it.

davidbiancolin (approved):
Just the one comment. We should probably wait until CI is back up.

sim/midas/src/main/scala/midas/platform/PlatformShim.scala
fabianschuiki: The `WriteXDCFile` pass still relies on that annotation to determine the paths. I can switch that over to just use the parameters directly if we want to entirely deprecate that `XDCPathToCircuitAnnotation`!

davidbiancolin: Ah yeah, right. Nvm me. Lets leave it for now.

fabianschuiki:
Alright, should I merge since the CI has run and passed on this, or wait until it's back up such that we can have some post-merge CI results in due time?

davidbiancolin:
I can think you can ship this since CI was clean. We don't run the full suite of tests post-merge currently.

fabianschuiki:
Sounds good, thanks 👍

PR #1322, opened 2022-12-07
Move termination to Run Farm `Inst` + `monitor_jobs` small rework

### LINKED ISSUE #1289: `terminate_on_completion: yes` is broken
### Background Work

- [X] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/firesim)
- [X] Yes, I searched [prior issues](https://github.com/firesim/firesim/issues)
- [X] Yes, I searched the [documentation](https://docs.fires.im/en/stable/)

### FireSim Version and Hash

1.8.1

### OS Setup

N/A

### Other Setup

_No response_

### Current Behavior

Upon termination (if the `terminate_on_completion:` flag is `yes`) you see an error like the following:

```
[192.168.4.252] Slot 0 completed! copying results.
!!! Parallel execution exception under host '192.168.4.252':
Process 192.168.4.252:
Traceback (most recent call last):
  File "/home/centos/firesim/.conda-env/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/centos/firesim/.conda-env/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/centos/firesim/.conda-env/lib/python3.9/site-packages/fabric/tasks.py", line 217, in _parallel_wrap
    queue.put({'name': name, 'result': task.run(*args, **kwargs)})
  File "/home/centos/firesim/.conda-env/lib/python3.9/site-packages/fabric/tasks.py", line 168, in run
    return self.wrapped(*args, **kwargs)
  File "/home/centos/firesim/.conda-env/lib/python3.9/site-packages/fabric/decorators.py", line 176, in inner
    return func(*args, **kwargs)
  File "/home/centos/firesim/deploy/runtools/firesim_topology_with_passes.py", line 529, in monitor_jobs_wrapper
    return my_node.instance_deploy_manager.monitor_jobs_instance(completed_jobs, teardown, terminateoncompletion, job_results_dir)
  File "/home/centos/firesim/deploy/runtools/run_farm_deploy_managers.py", line 416, in monitor_jobs_instance
    self.terminate_instance()
  File "/home/centos/firesim/deploy/runtools/run_farm_deploy_managers.py", line 621, in terminate_instance
    assert isinstance(self.boto3_instance_object, EC2InstanceResource)
AssertionError
Fatal error: One or more hosts failed while executing task 'monitor_jobs_wrapper'
Underlying exception:
```

### Expected Behavior

There should be no error. It should instead exit cleanly and terminate the instance running the simulation. 

Current workaround is to manually run `firesim terminaterunfarm` afterwards and disable the flag.

### Other Information

More explanation of the cause of the issue + the way to fix it:

Right now the instance deploy manager calls out to terminate_instance (which the deploy manager defines itself)(https://cs.github.com/firesim/firesim/blob/d84db3a39ba3ff22879192d68d8b57fb2ca30e99/deploy/runtools/run_farm_deploy_managers.py#L416). This function is implemented by accessing the boto object associated with the deploy manager. However, this object is always set to None and is never initialized. The bind function (https://cs.github.com/firesim/firesim/blob/d84db3a39ba3ff22879192d68d8b57fb2ca30e99/deploy/runtools/run_farm.py#L368) should set the boto object in the instance deploy manager (since right now the Inst object has no notion of the boto object)

abejgonzalez:
This PR addresses #1289.

It contains a few parts:
1. Changes the `linux-poweroff-*` workloads to auto-terminate with `terminate_on_completion`. This allows this code path to be tested in CI and further avoids FPGA runtime in CI.
2. The run farm deploy manager code for terminating an instance has been moved to the run farm `Inst` object's new `terminate_self` function. This keeps the boto3 object tracking strictly in the run farm instead of passing references to it throughout the codebase.
3. The `monitor_jobs_instance` is now changed to avoid the following scenario/bug: an instance would immediate terminate (due to `terminate_on_completion`), then the monitor jobs loop would be called again which would attempt to copy results off, however since the instance was terminated previously there would be a stall. 
  * The function signature is roughly the same with small changes + 1 addition: `completed_jobs -> prior_completed_jobs`, `teardown -> is_networked`, and a new `is_final_loop`. This `is_final_loop` is used in combination w/ the `is_networked` flag to potentially terminate the instance on the final loop of a networked simulation (instead of earlier) to avoid the issue mentioned.
  * The `instance_logger` now has a `debug` flag to dump to the debug logger
  * Variable renaming, statement re-organization. Happy to revert if need be. I found this easier to read... but since this is subjective I can revert to make the diff easier to read.

#### Related PRs / Issues

Fixes #1289

#### UI / API Impact

N/A

N/A

### Contributor Checklist

### DIFF
--- a/deploy/runtools/firesim_topology_with_passes.py
+++ b/deploy/runtools/firesim_topology_with_passes.py
@@ -521,13 +521,18 @@ class FireSimTopologyWithPasses:
         self.boot_simulation_passes(False, skip_instance_binding=True)
 
         @parallel
-        def monitor_jobs_wrapper(run_farm, completed_jobs: List[str], teardown: bool, terminateoncompletion: bool, job_results_dir: str) -> Dict[str, Dict[str, bool]]:
+        def monitor_jobs_wrapper(
+                run_farm: RunFarm,
+                prior_completed_jobs: List[str],
+                is_final_loop: bool,
+                is_networked: bool,
+                terminateoncompletion: bool,
+                job_results_dir: str) -> Dict[str, Dict[str, bool]]:
             """ on each instance, check over its switches and simulations
             to copy results off. """
             my_node = run_farm.lookup_by_host(env.host_string)
             assert my_node.instance_deploy_manager is not None
-            return my_node.instance_deploy_manager.monitor_jobs_instance(completed_jobs, teardown, terminateoncompletion, job_results_dir)
-
+            return my_node.instance_deploy_manager.monitor_jobs_instance(prior_completed_jobs, is_final_loop, is_networked, terminateoncompletion, job_results_dir)
 
         def loop_logger(instancestates: Dict[str, Any], terminateoncompletion: bool) -> None:
             """ Print the simulation status nicely. """
@@ -601,8 +606,8 @@ class FireSimTopologyWithPasses:
             rootLogger.info("""{}/{} simulations are still running.""".format(runningsims, totalsims))
             rootLogger.info("-"*80)
 
-        # teardown is required if roots are switches
-        teardown_required = isinstance(self.firesimtopol.roots[0], FireSimSwitchNode)
+        # is networked if a switch node is the root
+        is_networked = isinstance(self.firesimtopol.roots[0], FireSimSwitchNode)
 
         # run polling loop
         while True:
@@ -612,24 +617,21 @@ class FireSimTopologyWithPasses:
             def get_jobs_completed_local_info():
                 # this is a list of jobs completed, since any completed job will have
                 # a directory within this directory.
-                jobscompleted = os.listdir(self.workload.job_monitoring_dir)
-                rootLogger.debug("Monitoring dir jobs completed: " + str(jobscompleted))
-                return jobscompleted
+                monitored_jobs_completed = os.listdir(self.workload.job_monitoring_dir)
+                rootLogger.debug(f"Monitoring dir jobs completed: {monitored_jobs_completed}")
+                return monitored_jobs_completed
 
-            jobscompleted = get_jobs_completed_local_info()
-
-
-            # this job on the instance should return all the state about the instance
-            # e.g.:
-            # if an instance has been terminated (really - is termination
-            # requested and no jobs are left, then we will have implicitly
-            # terminated
-            teardown = False
-            instancestates = execute(monitor_jobs_wrapper, self.run_farm,
-                                    jobscompleted, teardown,
-                                    self.terminateoncompletion,
-                                    self.workload.job_results_dir,
-                                    hosts=all_run_farm_ips)
+            # return all the state about the instance (potentially copy back results and/or terminate)
+            is_final_run = False
+            monitored_jobs_completed = get_jobs_completed_local_info()
+            instancestates = execute(monitor_jobs_wrapper,
+                                     self.run_farm,
+                                     monitored_jobs_completed,
+                                     is_final_run,
+                                     is_networked,
+                                     self.terminateoncompletion,
+                                     self.workload.job_results_dir,
+                                     hosts=all_run_farm_ips)
 
             # log sim state, raw
             rootLogger.debug(pprint.pformat(instancestates))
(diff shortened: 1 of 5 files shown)


deploy/runtools/run_farm.py
```diff
@@ -39,6 +40,8 @@ class Inst(metaclass=abc.ABCMeta):
         metasimulation_enabled: true if this instance will be running metasimulations
     """
 
+    run_farm: RunFarm
```
davidbiancolin, line 43: I don't hink you should provide a naked reference to the RunFarm in `Inst`. That seem overly expansive. Since the termination logic can't be put in `Inst` itself, It would be better to make this reference private and add a `terminate_self` method that uses the private runfarm member to call terminate on itself.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -521,13 +521,18 @@ def run_workload_passes(self, use_mock_instances_for_testing: bool) -> None:
         self.boot_simulation_passes(False, skip_instance_binding=True)
 
         @parallel
-        def monitor_jobs_wrapper(run_farm, completed_jobs: List[str], teardown: bool, terminateoncompletion: bool, job_results_dir: str) -> Dict[str, Dict[str, bool]]:
+        def monitor_jobs_wrapper(
```
abejgonzalez, line 524: I can move this back to the old version... wasn't sure if we care about line length limits.

deploy/runtools/firesim_topology_with_passes.py
```diff
@@ -612,56 +617,59 @@ def loop_logger(instancestates: Dict[str, Any], terminateoncompletion: bool) ->
...
-                                    jobscompleted, teardown,
-                                    self.terminateoncompletion,
-                                    self.workload.job_results_dir,
-                                    hosts=all_run_farm_ips)
+                monitored_jobs_completed = os.listdir(self.workload.job_monitoring_dir)
+                rootLogger.debug(f"Monitoring dir jobs completed: {monitored_jobs_completed}")
+                return monitored_jobs_completed
+
+            # return all the state about the instance (potentially copy back results and/or terminate)
+            is_final_run = False
```
abejgonzalez, line 625: The core difference here is that there is now a variable to track whether this execution of `monitor_jobs_wrapper` is the final run or not.

deploy/runtools/run_farm.py
abejgonzalez: Ok I think I addressed this like you recommended.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -305,120 +306,123 @@ def running_simulations(self) -> Dict[str, List[str]]:
...
+        self.instance_logger(f"Prior completed jobs: {prior_completed_jobs}", debug=True)
+
+        def do_terminate():
+            if (not is_networked) or (is_networked and is_final_loop):
+                if terminateoncompletion:
+                    self.terminate_instance()
 
-        rootLogger.debug("completed jobs " + str(completed_jobs))
 
         if not self.instance_assigned_simulations() and self.instance_assigned_switches():
```
abejgonzalez, line 325: This section of the code is the same except for the `do_terminate` section. However, when the `do_terminate` is called here `is_final_loop == True` so the `if terminateoncompletion:` is always reached which makes the statement the same as before.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -305,120 +306,123 @@ def running_simulations(self) -> Dict[str, List[str]]:
...
-                if terminateoncompletion:
-                    # terminate the instance since teardown is called and instance
-                    # termination is enabled
-                    self.terminate_instance()
+                do_terminate()
 
-                # don't really care about the return val in the teardown case
-                return {'switches': dict(), 'sims': dict()}
+                return {'switches': {}, 'sims': {}}
+            else:
```
abejgonzalez, line 341: Shifted this into an `else:` block. The same otherwise.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -305,120 +306,123 @@ def running_simulations(self) -> Dict[str, List[str]]:
...
                 # about switch status, so don't bother to populate it
                 jobnames_to_completed = {jname: True for jname in jobnames}
-                return {'sims': jobnames_to_completed, 'switches': dict()}
+                return {'sims': jobnames_to_completed, 'switches': {}}
 
             # at this point, all jobs are NOT completed. so, see how they're doing now:
             instance_screen_status = self.running_simulations()
+
             switchescompleteddict = {k: False for k in instance_screen_status['switches']}
+            slotsrunning = [x for x in instance_screen_status['simdrivers']]
```
abejgonzalez, line 376: L376 -> 385 is the same except for some variable name changes. Happy to revert.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -305,120 +306,123 @@ def running_simulations(self) -> Dict[str, List[str]]:
...
-
-            slotsrunning = [x for x in instance_screen_status['simdrivers']]
+                    sw_name = switchsim.switch_builder.switch_binary_name()
+                    if sw_name not in switchescompleteddict.keys():
+                        switchescompleteddict[sw_name] = True
 
-            rootLogger.debug("slots running")
-            rootLogger.debug(slotsrunning)
+            # fill in whether sims have terminated
+            completed_jobs = prior_completed_jobs.copy() # create local copy to append to
```
abejgonzalez, line 388: The rest of this diff is all the same except for the `do_terminate` at the end (and some variable renames - again happy to revert).

abejgonzalez:
@sagark @t14916 Ping for review

sagark (approved):
LGTM

t14916 (approved):
lgtm as well

davidbiancolin (approved):

PR #1316, opened 2022-12-02
CI Rework: More Aggressive Culling Of FPGA Resources, Slack/PR Notifications

abejgonzalez:
This PR attempts to fix/alleviate some CI issues that have caused resource leakages (of both manager and FPGA instances).

Existing issues found:
- The workflow monitor did not start properly (a combination of needing all env. variables, and other small oversights). However, this wasn't apparent since there was no logging/notification of something going wrong.
- FPGA instances were left alive for too long (workflow monitor was failing is one reason for this).

It does the following:

- `workflow_monitor.py` and `setup_workflow_monitor.py`
  - The monitor needs access to all environment variables used in `ci_variables.py`. Thus, instead of setting an `env.sh` file that holds needed env. variables, a new `ci_env` dict is created (see the `ci_variables.py` changes) and set for the corresponding `run()` calls.
  - Now there are 2 verification steps to ensure that the workflow monitor at least started properly (the workflow monitor has it's own check to ensure it is properly running)
  - Since the workflow monitor has access to all env. variables needed, it doesn't have the args anymore.
  - The workflow monitor now sends PR comments when starting + failing
  - The workflow monitor now has a `check_and_terminate_select_instances` that has a 45m timeout on FPGA instances (see `platform_lib.py` for more info).
  - The workflow monitor now has a `try/catch` to check if failures happen (at least it should hopefully send the PR comment if something fails).
- `ci_variables.py` - A new `ci_env` dict is populated with all env. variables used throughout CI. Instead of using python variable names for environment variables (which don't have the same naming convention) use the env. variable name for the key in the dict. All code locations are changed to account for this. This dict is type checked so that all keys are considered present. Future CI type checking will happen in a followup PR.
- `firesim-cleanup.yml` - This workflow now sends a Slack message if something goes wrong. This is tied to the `cull-instances` script now returning an error if it culls something it expected not to cull.
- `cull-old-ci-insts.py` - This now explicitly checks for FPGA instances that are live for over 1hr. Also now exits non-zero if something was terminated unexpectedly (triggers a slack message).
- `{github_}common.py` - Moved GitHub specific items to a new file `github_common.py` to avoid a circular dependency with the `platform_lib` file. A new issues URL and issue function is added for PR comment posts.
- `platform_lib.py`:
  - New `find_select_ci_instances` func. for each platform to find "select" CI instances (i.e. FPGA instances) and return them.
[...]

PR #1325, opened 2022-12-09
Add CI typechecking

abejgonzalez:
Fixes CI type checking errors and adds a CI type checking step to CI. Builds off #1316 

### Contributor Checklist

### DIFF
--- a/.github/scripts/build-default-workloads.py
+++ b/.github/scripts/build-default-workloads.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from fabric.api import *
+from fabric.api import prefix, settings, run, execute # type: ignore
 
 from common import manager_fsim_dir, set_fabric_firesim_pem
 
--- a/.github/scripts/github_common.py
+++ b/.github/scripts/github_common.py
@@ -1,8 +1,8 @@
 import math
-from fabric.api import *
 import requests
-from ci_variables import ci_env
 import json
+
+from ci_variables import ci_env
 
 from typing import Dict, List, Any
 
@@ -17,14 +17,14 @@ gha_workflow_api_url = f"{gha_runs_api_url}/{ci_env['GITHUB_RUN_ID']}"
 def get_header(gh_token: str) -> Dict[str, str]:
     return {"Authorization": f"token {gh_token.strip()}", "Accept": "application/vnd.github+json"}
 
-def get_runners(gh_token: str) -> List:
+def get_runners(gh_token: str) -> List[Dict[str, Any]]:
     r = requests.get(gha_runners_api_url, headers=get_header(gh_token))
     if r.status_code != 200:
         raise Exception(f"Unable to retrieve count of GitHub Actions Runners\nFull Response Below:\n{r}")
     res_dict = r.json()
     runner_count = res_dict["total_count"]
 
-    runners = []
+    runners: List[Dict[str, Any]] = []
     for page_idx in range(math.ceil(runner_count / 30)):
         r = requests.get(gha_runners_api_url, params={"per_page" : 30, "page" : page_idx + 1}, headers=get_header(gh_token))
         if r.status_code != 200:
--- a/.github/scripts/platform_lib.py
+++ b/.github/scripts/platform_lib.py
@@ -3,22 +3,20 @@ import abc
 import sys
 import boto3
 import os
-from enum import Enum
-from fabric.api import *
 import pytz
 import datetime
-import requests
-from xmlrpc.client import DateTime
+from enum import Enum
+
+from azure.mgmt.resource import ResourceManagementClient # type: ignore
+from azure.identity import DefaultAzureCredential # type: ignore
+from azure.mgmt.compute import ComputeManagementClient # type: ignore
+import azure.mgmt.resourcegraph as arg # type: ignore
 
 from ci_variables import ci_env
 from github_common import issue_post, get_issue_number
 
-from azure.mgmt.resource import ResourceManagementClient
-from azure.identity import DefaultAzureCredential
-from azure.mgmt.compute import ComputeManagementClient
-import azure.mgmt.resourcegraph as arg
-
-from typing import Any, Callable, Dict, List, Iterable, Tuple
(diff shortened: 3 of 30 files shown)


.github/workflows/firesim-run-tests.yml
```diff
@@ -164,16 +164,27 @@ jobs:
...
-        run: .github/scripts/run-mypy-typechecker.py
+      - name: Run manager Python type checker
+        run: .github/scripts/run-manager-python-typechecking.py
+
+  run-ci-python-typecheck:
+      name: run-ci-python-typecheck
+      needs: [setup-manager]
+      runs-on: aws-${{ github.run_id }}
+      env:
+      TERM: xterm-256-color
```
abejgonzalez, line 183:
```suggestion
        TERM: xterm-256-color
```

.github/scripts/build-default-workloads.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from fabric.api import *
+from fabric.api import prefix, settings, run, execute # type: ignore
```
t14916, line 3: So this ignores all the typing information coming out of functions in fabric.api right? I've looked into this in the past,  and don't remember seeing any way to get the types out of fabric.api, but maybe it's worth looking into again?

.github/scripts/platform_lib.py
```diff
@@ -244,10 +249,17 @@ def change_workflow_instance_states(self, gh_token: str, workflow_tag: str, stat
         elif state_change == 'terminate':
             print(f"Terminating instances: {', '.join(instance_ids)}")
             self.deregister_runners(gh_token, f"aws-{workflow_tag}")
-            self.client.terminate_instances(InstanceIds=instance_ids, DryRun=dryrun)
+            self.platform_terminate_instances(instance_ids, dryrun)
         else:
             raise ValueError(f"Unrecognized transition type: {state_change}")
 
+    def platform_terminate_instances(self, platform_list: List[Any], dryrun: bool = False) -> None:
```
t14916, line 256: So the reason this is done is to avoid some of the typing issues with the boto3 client?

.github/scripts/github_common.py
```diff
@@ -17,14 +17,14 @@
 def get_header(gh_token: str) -> Dict[str, str]:
     return {"Authorization": f"token {gh_token.strip()}", "Accept": "application/vnd.github+json"}
 
-def get_runners(gh_token: str) -> List:
+def get_runners(gh_token: str) -> List[Dict[str, Any]]:
```
t14916, line 20: Is it possible to be more specific than Any? since we're pulling down a json from http, is it possible to get something that isn't either `int` or `string`? This is safer for sure.

.github/workflows/firesim-run-tests.yml
```diff
@@ -164,16 +164,27 @@ jobs:
...
     env:
       TERM: xterm-256-color
     steps:
       - uses: actions/checkout@v3
-      - name: Run mypy type checker
-        run: .github/scripts/run-mypy-typechecker.py
+      - name: Run manager Python type checker
+        run: .github/scripts/run-manager-python-typechecking.py
+
+  run-ci-python-typecheck:
```
t14916, line 178: Just a thought, but what if we made the ci typecheck the first script run in the CI pipeline? This might be a bit too strict, but the idea would be to check earlier on and maybe make this a dependency that forces us to fix typing before testing CI. Just a suggestion, this might be to anal.

.github/workflows/firesim-run-tests.yml
```diff
@@ -164,16 +164,27 @@ jobs:
...
+      name: run-ci-python-typecheck
+      needs: [setup-manager]
+      runs-on: aws-${{ github.run_id }}
+      env:
+        TERM: xterm-256-color
+      steps:
+      - uses: actions/checkout@v3
+      - name: Run CI Python type checker
+          run: .github/scripts/run-ci-python-typechecking.py
```
abejgonzalez, line 187:
```suggestion
        run: .github/scripts/run-ci-python-typechecking.py
```

t14916:
just for personal reference, did you make these changes by hand or did you use some tool to fix some of the mypy errors (expanding the wildcards, adding the type ignore comments,etc.)

.github/workflows/firesim-run-tests.yml
abejgonzalez: I agree this should be done. However, I think the best way to do this is to setup a way to initialize the FireSim repo w/ GH-A managed runners and move this job there so that even spawning the manager is blocked by a typing check. I don't have time to do this, so I'll leave this to you or just future work.

t14916: Yeah that's what I was suggesting, I'll put it in my backlog

.github/scripts/platform_lib.py
abejgonzalez: In the `cull_instances` file you explicitly call the `terminate_azure_vms` which is a hidden method in the platform library. So I added this new method so that we avoid piping platform-specific methods out from the platform lib.

.github/scripts/build-default-workloads.py
abejgonzalez: Correct. Fabric doesn't have any types. I also since we use a hacked version of Fabric (fab-classic) I don't expect it to add types anytime soon. We use this "hack" across the codebase in the interim.

.github/scripts/github_common.py
abejgonzalez: Technically the dict can have values of lists, strs, dicts, bools, floats, etc so its a bit more complicated. Looking around quickly there is a way to specify a JSON type manually (https://github.com/python/typing/issues/182) but I'm not sure where to put the alias definition (I think it would be good to avoid duplicating this in places). I figured this was sufficient since we do this in other places but we can make this an issue for future work.

abejgonzalez:
> just for personal reference, did you make these changes by hand or did you use some tool to fix some of the mypy errors (expanding the wildcards, adding the type ignore comments,etc.)

I did this by hand.

t14916 (approved):
lgtm

PR #1326, opened 2022-12-09
Added UART bridge test

nandor:
Added a UART test in a new group of tests, 'bridges'. This test group includes the bridges from firesim-lib, which rely on more external dependencies. The previous `midasexamples` test suite is limited to simple tests with no dependencies.
The `PointerChaser` test will have to be moved to the `fasedtests` group later on.

### Contributor Checklist

### DIFF
--- a/.github/workflows/firesim-run-tests.yml
+++ b/.github/workflows/firesim-run-tests.yml
@@ -267,9 +267,22 @@ jobs:
         with:
           test-name: "CIGroupB"
 
+  run-test-firesim-lib:
+    name: run-test-firesim-lib
+    needs: [run-test-groupB]
+    runs-on: aws-${{ github.run_id }}
+    env:
+      TERM: xterm-256-color
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run firesim-lib Scala tests
+        uses: ./.github/actions/run-scala-test
+        with:
+          test-name: "BridgeTests"
+
   run-chipyard-tests:
     name: run-chipyard-tests
-    needs: [run-test-groupB]
+    needs: [run-test-firesim-lib]
     runs-on: aws-${{ github.run_id }}
     env:
       TERM: xterm-256-color
--- a/sim/firesim-lib/src/main/cc/bridges/uart.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/uart.cc
@@ -42,10 +42,7 @@ void sighand(int s) {
 }
 #endif
 
-uart_t::uart_t(simif_t *sim,
-               const UARTBRIDGEMODULE_struct &mmio_addrs,
-               int uartno)
-    : bridge_driver_t(sim), mmio_addrs(mmio_addrs) {
+uart_fd_handler::uart_fd_handler(int uartno) {
   this->loggingfd = 0; // unused
 
   if (uartno == 0) {
@@ -92,7 +89,42 @@ uart_t::uart_t(simif_t *sim,
   fcntl(inputfd, F_SETFL, fcntl(inputfd, F_GETFL) | O_NONBLOCK);
 }
 
-uart_t::~uart_t() { close(this->loggingfd); }
+uart_fd_handler::~uart_fd_handler() { close(this->loggingfd); }
+
+std::optional<char> uart_fd_handler::get() {
+  char inp;
+  int readamt;
+  if (specialchar) {
+    // send special character (e.g. ctrl-c)
+    // for stdin handling
+    //
+    // PTY should never trigger this
+    inp = specialchar;
+    specialchar = 0;
+    readamt = 1;
+  } else {
+    // else check if we have input
+    readamt = ::read(inputfd, &inp, 1);
+  }
+
+  if (readamt <= 0)
+    return std::nullopt;
+  return inp;
+}
+
+void uart_fd_handler::put(char data) {
+  ::write(outputfd, &data, 1);
+  if (loggingfd) {
+    ::write(loggingfd, &data, 1);
+  }
+}
+
+uart_t::uart_t(simif_t *sim,
+               const UARTBRIDGEMODULE_struct &mmio_addrs,
+               int uartno)
+    : uart_t(sim, mmio_addrs, std::make_unique<uart_fd_handler>(uartno)) {}
+
+uart_t::~uart_t() {}
 
 void uart_t::send() {
   if (data.in.fire()) {
@@ -119,32 +151,14 @@ void uart_t::tick() {
     this->recv();
 
     if (data.in.ready) {
-      char inp;
-      int readamt;
-      if (specialchar) {
-        // send special character (e.g. ctrl-c)
-        // for stdin handling
-        //
-        // PTY should never trigger this
-        inp = specialchar;
-        specialchar = 0;
-        readamt = 1;
-      } else {
-        // else check if we have input
-        readamt = ::read(inputfd, &inp, 1);
-      }
-
-      if (readamt > 0) {
-        data.in.bits = inp;
+      if (auto bits = handler->get()) {
+        data.in.bits = *bits;
         data.in.valid = true;
       }
     }
 
     if (data.out.fire()) {
-      ::write(outputfd, &data.out.bits, 1);
-      if (loggingfd) {
-        ::write(loggingfd, &data.out.bits, 1);
-      }
+      handler->put(data.out.bits);
     }
 
     this->send();
--- a/sim/firesim-lib/src/main/cc/bridges/uart.h
+++ b/sim/firesim-lib/src/main/cc/bridges/uart.h
@@ -2,8 +2,10 @@
 #ifndef __UART_H
 #define __UART_H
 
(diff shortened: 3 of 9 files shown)


davidbiancolin:
FWIW, i've been meaning to remove the pointer chaser. It's not a good enough test to warrant maintain the various oddities it has.

.github/workflows/firesim-run-tests.yml
```diff
@@ -256,9 +256,22 @@ jobs:
         with:
           test-name: "CIGroupB"
 
+  run-test-bridges:
```
davidbiancolin, line 259: For now refer to these as `firesim-lib` bridges. We started talking about renaming these to berkeley-bridges or something more informative, but i need to pull all ofthe non-bridge stuff out of firesim-lib first.

.github/workflows/firesim-run-tests.yml
```diff
@@ -256,9 +256,22 @@ jobs:
...
+    name: run-test-bridges
+    needs: [run-test-groupB]
+    runs-on: aws-${{ github.run_id }}
+    env:
+      TERM: xterm-256-color
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run CIGroupB Scala tests
```
davidbiancolin, line 267: Update this

sim/firesim-lib/src/main/cc/bridges/uart.cc
```diff
@@ -92,7 +89,42 @@ uart_t::uart_t(simif_t *sim,
   fcntl(inputfd, F_SETFL, fcntl(inputfd, F_GETFL) | O_NONBLOCK);
 }
 
-uart_t::~uart_t() { close(this->loggingfd); }
+uart_fd_handler::~uart_fd_handler() { close(this->loggingfd); }
```
davidbiancolin, line 92: please add `ci:fpga-deploy" just to sanity check things continue to work correctly.

sim/firesim-lib/src/main/cc/bridges/uart.h
```diff
@@ -15,28 +17,68 @@
 // all-caps, suffixed with "_struct" and "_struct_guard" respectively.
 
 #ifdef UARTBRIDGEMODULE_struct_guard
-class uart_t : public bridge_driver_t {
+
+/**
+ * Base class for callbacks handling data coming in our out a UART stream.
```
davidbiancolin, line 22:
```suggestion
 * Base class for callbacks handling data coming in and out of our UART stream.
```
This is what you meant right?

sim/src/main/cc/bridges/UARTModule.cc
```diff
@@ -0,0 +1,75 @@
...
+
+  void run() {
+    // Reset the device.
+    poke(reset, 1);
+    step(1);
+    poke(reset, 0);
+    step(1);
+
+    // Tick for a while.
+    for (unsigned i = 0; i < 30; ++i) {
```
davidbiancolin, line 57: How long does this test take to run? For the string you're writing you could make this test run much more quickly by calling tick more frequently relative to step. For testing bridges like this, you can favor the non-blocking variant of "step" and instead write:

```
step(<some_reasonable_upper_bound>, false)
while(!done() <(&& possibly other conditions)>) {
  uart->tick();
}
```

davidbiancolin: Basically the UART is sitting idle for the bulk of each of the 10K cycle steps.

sim/src/main/cc/midasexamples/Driver.cc
```diff
@@ -97,6 +97,8 @@
 #include "TerminationModule.h"
 #elif defined DESIGNNAME_CustomConstraints
 #include "CustomConstraints.h"
+#elif defined DESIGNNAME_UARTModule
```
davidbiancolin, line 100: This isn't required, but vestigial from trying to put the test in midasexamples first?

davidbiancolin (commented):
First half

sim/src/main/cc/bridges/UARTModule.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "simif_peek_poke.h"
+
+class UARTModuleTest final : public simif_peek_poke_t {
+public:
+  class Handler final : public uart_handler {
+  public:
+    ~Handler() {
+      // Check that the input and output buffers are equal.
+      if (in_buffer != out_buffer) {
+        fprintf(stderr,
```
davidbiancolin, line 21: I'm curious, what's your opinion on streams vs printf for output to stderr?

sim/src/main/cc/bridges/UARTModule.cc
```diff
@@ -0,0 +1,75 @@
...
+      uart->tick();
+      step(10000);
+    }
+  }
+
+private:
+  std::unique_ptr<uart_t> uart;
+};
+
+int main(int argc, char **argv) {
```
davidbiancolin, line 67: Should future tests in this project also define there own mains like this? I need to add some TracerV features and wanted to write tests rebased against this branch

davidbiancolin (commented):
All in all looks good to me. Thanks for taking the initiative here.  A few things pertaining to docs:

Can you skim over the UART bridge walkthrough and make sure it's still pointing at sensible points of the code. Those are hosted here: https://docs.fires.im/en/main/Golden-Gate/Bridge-Walkthrough.html

Can you also add a section for this new target project, following the FASEDtests example here: https://docs.fires.im/en/main/Advanced-Usage/Generating-Different-Targets.html#fased-tests-fasedtests-project

Maybe consider updating this: https://docs.fires.im/en/main/Developer-Docs/GoldenGate-and-Driver-Development.html if necessary 

The rst source for all of this lives under `docs/` in case you weren't aware.

sim/src/main/cc/bridges/UARTModule.cc
nandor: For now, let's define the same method everywhere. The entry points will change in subsequent PRs involving the DPI changes and we can introduce a uniform representation there.

sim/src/main/cc/bridges/UARTModule.cc
nandor: Thanks for clarifying this!

sim/src/main/cc/bridges/UARTModule.cc
nandor: I've used printf since it was used in other parts as well. But overall I would prefer streams.

Moving forward, I would like to move the test checking functions out of `simif_peek_poke_t` into `TestHarness`, adding googletest-like assertions there as well for verification, so there is no need to print anything in tests.

davidbiancolin: Yeah, that sounds ideal. Thanks for the context.

sim/src/main/cc/bridges/UARTModule.cc
```diff
@@ -0,0 +1,84 @@
...
+
+    // Reset the device.
+    poke(reset, 1);
+    step(1);
+    poke(reset, 0);
+    step(1);
+
+    // Tick until the out buffer is filled in.
+    step(300000, false);
+    while (!simif->done()) {
```
davidbiancolin, line 64: Better, now you should add an iteration bound in case something deadlocks in the target. YOu could instead check that the DUT is making forward progress by calling simif_t::actual_tcycle, but that's a little trickier to tune -- i'll leave it to your discretion. 

This sort of thing can be caught by a host-time max cycle bound in the metasim harness, but you can provide a better error here.

davidbiancolin (commented):
One follow up, and then this will be ready to ship.

sim/src/test/scala/bridges/BridgeSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+    s"PLATFORM_CONFIG=${platformConfigs}",
+  )
+
+  def run(
+    backend:  String,
+    debug:    Boolean      = false,
+    logFile:  Option[File] = None,
+    waveform: Option[File] = None,
+    args:     Seq[String]  = Nil,
+  ) = {
```
davidbiancolin, line 38: I won't let this block the PR, but you should always label return types moving forward.

sim/src/test/scala/bridges/BridgeSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+    * @param b
+    *   Backend simulator: "verilator" or "vcs"
+    * @param debug
+    *   When true, captures waves from the simulation
+    * @param args
+    *   A seq of PlusArgs to pass to the simulator.
+    * @param shouldPass
+    *   When false, asserts the test returns a non-zero code
+    */
+  def runTest(b: String, debug: Boolean = false, args: Seq[String] = simulationArgs, shouldPass: Boolean = true) {
```
davidbiancolin, line 61: Similarly, you want ` : Unit = ` here. 

I'll catch all of this when i get around to cleaning up some these tests.

davidbiancolin (approved):

PR #1328, opened 2022-12-12
Introduced an interface to capture the user-defined logic of a simuation

nandor:
Instead of chains of derivation, the user-defined part of a simulation now must implement the `simulation_t` interface. The `simif_t` implementation takes a reference to this interface and runs it.

These changes are useful as they define a uniform interface across tests and the actual FireSim driver.

This PR also nukes the ucontext-based implementation from `context.h` and renames the class into `pcontext` to avoid linking conflicts with the same class from libfesvr.a.

The main method of a simulation is no longer defined alongside the simulation. Instead, the simulation defined a `create_simulation` function which builds a `simulation_t` object. Mains are moved alongside the main classes of specific platforms and simulators, which use the create hook to fetch an instance of the simulation to run.

### Contributor Checklist


fabianschuiki (approved):
This looks like a very reasonable approach to me. I like reducing the reliance on pure inheritance and using more of a mixin/trait kind of approach. Seems to declutter things quite a bit!

sim/midas/src/main/cc/simif.h
```diff
@@ -21,6 +25,41 @@ double diff_secs(midas_time_t end, midas_time_t start);
...
+
+  /**
+   * Simulation main loop.
+   *
+   * @return Exit code to return to the system.
+   */
+  virtual int simulation_run() = 0;
+
+  /**
+   * Simulation initialization. Should set up bridges.
```
davidbiancolin, line 50: Be precise about what "setting up bridges" means vis-a-vis mmio -> maybe note it's calling bridge_driver_t::init(), and so there can and will be SW <-> "HW" interactions. 

Speaking of which, we probably need better names than "driver" and "hw" to dilineate between the CPU-hosted and what will be the FPGA-hosted bit of a simulator. Just something to think about.

sim/midas/src/main/cc/simif.h
```diff
@@ -21,6 +25,41 @@ double diff_secs(midas_time_t end, midas_time_t start);
...
+   */
+  virtual void simulation_init() {}
+
+  /**
+   * Simulation finalization. Should tear down bridges.
```
davidbiancolin, line 55: Again, might be worth referring to the bridge_driver_t method, or being more vague and echoing the doc there.

sim/midas/src/main/cc/simif.h
```diff
@@ -155,26 +172,72 @@ class simif_t {
...
 
-  uint64_t get_seed() { return seed; };
+  /// Returns the next available random number, modulo limit.
+  uint64_t rand_next(uint64_t limit) { return gen() % limit; }
 
-  // Returns the current target cycle of the fastest clock in the simulated
-  // system, based on the number of clock tokens enqueued (will report a larger
-  // number)
+  /**
+   * Returns the target cycle count.
```
davidbiancolin, line 182:
```suggestion
```
Lets not use the definite article here considering there are many target clocks at play.

sim/midas/src/main/cc/simif_emul.h
```diff
@@ -16,28 +16,30 @@ class simif_emul_t : public simif_t {
...
+              size_t threshold_bytes) override;
+  size_t push(unsigned int stream_idx,
+              void *src,
+              size_t num_bytes,
+              size_t threshold_bytes) override;
   /**
    * @brief Pointers to inter-context (i.e., between VCS/verilator and driver)
    * AXI4 transaction channels
    *
    * These have external linkage to enable VCS to easily access them.
```
davidbiancolin, line 39: This needs to be updated now, no?

sim/midas/src/main/cc/simif_emul.h
```diff
@@ -47,23 +49,31 @@ class simif_emul_t : public simif_t {
...
   // switching back to the driver process. +fuzz-host-timings sets this to a
   // value > 1, introducing random delays in axi4 tranactions that MMIO and
   // bridge streams.
   int maximum_host_delay = 1;
-  void advance_target();
+
+  std::string waveform = "dump.vcd";
+  uint64_t memsize = 1L << MEM_ADDR_BITS;
+
+  virtual void advance_target() = 0;
```
davidbiancolin, line 76: It be good to explain what this does now that it's going to be implemented by different subclasses.

sim/midas/src/main/cc/simif_emul_vcs.h
```diff
@@ -0,0 +1,31 @@
+// See LICENSE for license details.
+
+#ifndef __SIMIF_EMUL_VCS_H
+#define __SIMIF_EMUL_VCS_H
+
+#include "emul/context.h"
+#include "simif_emul.h"
+
+/**
+ * VCS-specific software emulator implementation.
```
davidbiancolin, line 10:
```suggestion
 * VCS-specific metasimulator implementation.
```

sim/midas/src/main/cc/simif.h
```diff
@@ -21,6 +25,41 @@ double diff_secs(midas_time_t end, midas_time_t start);
...
 
+/**
+ * Interface for a simulation implementation.
+ *
+ * `simif_t` interacts with the simulation, initialising it, running and and
+ * running finalisation logic at the end.  All simulation implementations
+ * (midasexamples, firesim, bridges, fasedtests) derive this interface,
+ * implementing the driver-specific logic.
+ */
+class simulation_t {
```
davidbiancolin, line 36: So, on some level this is a very appropriate name. However, it's also highly overloaded in our context. You say
`driver-specific logic`, so maybe "Driver" or "driver_t"? or "user_driver"? I don't know.

sim/midas/src/main/cc/simif.h
```diff
@@ -155,26 +172,72 @@ class simif_t {
 
   // End host-platform interface.
 
-  // LOADMEM functions
-  void read_mem(size_t addr, mpz_t &value);
-  void write_mem(size_t addr, mpz_t &value);
-  void write_mem_chunk(size_t addr, mpz_t &value, size_t bytes);
-  void zero_out_dram();
+  /// Returns the seed of the random-number generator.
```
davidbiancolin, line 175: I'd prefer if we'd rally about javadoc/scaladoc variants of doxygen comments if possible.

sim/midas/src/main/cc/simif.h
```diff
@@ -155,26 +172,72 @@ class simif_t {
...
+  /// Returns the last completed cycle number.
+  uint64_t get_end_tcycle() { return end_tcycle; }
+
+  // LOADMEM functions.
+  void read_mem(size_t addr, mpz_t &value);
+  void write_mem(size_t addr, mpz_t &value);
+  void write_mem_chunk(size_t addr, mpz_t &value, size_t bytes);
+
+private:
+  /// Waits for the target to be initialised.
```
davidbiancolin, line 202: I don't like "target_init" it's more like simulator init -- "host_init" captured that better.

davidbiancolin (commented):
This probably should've been three PRs. 
1) Removing ucontext
2) Refactoring associated with simulation_t
3) Building a class hierarchy around simif_emul_t. 

I have to review a lot of PRs these days and would appreciate it if you made it easier on me. 

Regarding the changes themselves. I think they look fantastic for the most part. I mostly have nits about documentation + naming. We will come to regret using `simulation_t` i'm sure so in our weekly we should brainstorm a better name, i won't let it block this PR. 

Since you're already in the business adding new classes, can you remove references to "emul" in favor of metasimulation or metasim? The historical use of "emul" in this code base is ripe for confusion.

sim/midas/src/main/cc/simif.h
nandor: I've adjusted the comments to describe what the tests do now in them. In the future, we should further split these.

sim/midas/src/main/cc/simif.h
nandor: Yes, I think `driver_t` might be more appropriate. But I'll leave it for a later PR.

sim/midas/src/main/cc/simif.h
nandor: I don't like this either, especially since it involves waiting for `INIT_DONE` and loading some memory contents. We should address it in the future, but for now I think `target_init` still captures the meaning since we are doing MMIO to the target and waiting for its response.

nandor:
Sorry about the `pthread` stuff - this already comes from a big chain of PRs and I didn't split it further.

I will remove references to `emul` at a later point, as those PRs would be a pain to adjust right now.

I've adjusted some names and comments, but in the future we'll definitely want to refine the `simulation_t` interface.

nandor:
`ci:fpga-deploy` was timing out after rebasing onto main, probably due to the latest changes. Since the FPGA runs were succeeding previously, I've removed the deploy flag.

davidbiancolin:
A few final questions i forgot to ask. 

1) You've tested this against VCS on sifive machines right? 
2) By any chance, did you measure any performance different moving from ucontext to the pthreads implementation? Any extra latency will probably only matter for small designs so it's likely nbd.

nandor:
I have tested with VCS. I haven't noticed any differences in simulation times (I've mostly been running `BaselineMultichannelTest`).

PR #1336, opened 2022-12-17
Introduced a uniform harness over bridge tests

nandor:
This PR rewrites the UART bridge to introduce a handler that can read and write to regular files.
No all testing logic and assertions are handled on the Scala side.

### Contributor Checklist

### DIFF
--- a/sim/src/main/cc/bridges/BridgeHarness.cc
+++ b/sim/src/main/cc/bridges/BridgeHarness.cc
@@ -1,0 +1,50 @@
+// See LICENSE for license details.
+
+#include "BridgeHarness.h"
+
+#include "bridges/blockdev.h"
+#include "bridges/bridge_driver.h"
+#include "bridges/uart.h"
+
+BridgeHarness::BridgeHarness(const std::vector<std::string> &args,
+                             simif_t *simif)
+    : simif_peek_poke_t(simif, PEEKPOKEBRIDGEMODULE_0_substruct_create),
+      simulation_t(args) {}
+
+BridgeHarness::~BridgeHarness() {}
+
+void BridgeHarness::add_bridge_driver(bridge_driver_t *bridge) {
+  bridges.emplace_back(bridge);
+}
+
+void BridgeHarness::simulation_init() {
+#include "constructor.h"
+  for (auto &bridge : bridges) {
+    bridge->init();
+  }
+}
+
+int BridgeHarness::simulation_run() {
+  // Reset the device.
+  poke(reset, 1);
+  step(1);
+  poke(reset, 0);
+  step(1);
+
+  // Tick until all requests are serviced.
+  step(get_step_limit(), false);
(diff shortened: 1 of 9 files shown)


sim/src/main/cc/bridges/BridgeHarness.h
```diff
@@ -0,0 +1,43 @@
+#ifndef MIDAEXAMPLES_BRIDGEHARNESS_H
+#define MIDAEXAMPLES_BRIDGEHARNESS_H
+
+#include "simif.h"
+#include "simif_peek_poke.h"
+
+class bridge_driver_t;
```
davidbiancolin, line 7: For my own understanding, why the declaration here instead of importing from bridge_driver?

sim/src/main/cc/bridges/BridgeHarness.cc
```diff
@@ -0,0 +1,50 @@
...
+int BridgeHarness::simulation_run() {
+  // Reset the device.
+  poke(reset, 1);
+  step(1);
+  poke(reset, 0);
+  step(1);
+
+  // Tick until all requests are serviced.
+  step(get_step_limit(), false);
+  for (unsigned i = 0; i < get_tick_limit() && !simif->done(); ++i) {
```
davidbiancolin, line 36: If you're going to provide a useful and generic simulation_run you should probably check the bridges don't signal for terminate, and treat it as an error if so. 

Also, hitting the tick limit likely implies deadlock, and should be treated as an error too, unless the watchdog is configured differently.

davidbiancolin (commented):
Mostly looks great. 

The UART has historically been quite fragile. Please run with `ci:fpga-deploy` so we can check we're getting sane output. 

My only material feedback is about the `simulation_run` implementation for the bridge_harness. Some of the comments need to be updated from being copied from midasexamples but that's secondary.

sim/src/main/cc/bridges/BridgeHarness.h
nandor: In general, it is a good idea to use forward declarations in headers to avoid unnecessary dependencies. If some C++ file including `BridgeHarness.h` does not actually need to know about the implementation of the bridge driver, it can be compiled without unnecessarily including the header.

sim/src/main/cc/bridges/BridgeHarness.cc
nandor: This details are not particularly clear to me - we should discuss what a proper harness looks like offline and evolve it from this loop. Currently, if the tick limit is hit, then bridges will not have produced the correct output and the test will fail.

nandor: In particular, in #1342 I am trying to unify all harnesses. We can get these tests to obey the timeout flag parsed by that class afterwards and display a consistent error message.

davidbiancolin:
I'm just testing i `stdin` is working fine then i'll approve this.

davidbiancolin (approved):
Send it

PR #1341, opened 2022-12-21
Separated testing from peek-poke logic

nandor:
This patch moves the testing-related bits of `simif_peek_poke` to `TestHarness`. `peek_poke_t` is not a bridge-like class that interacts through MMIO with the target. Moved the peek-poke bridge to the bridges directory as it is shared across many tests and deserves first-class treament.

### Contributor Checklist
- [x ] Did you delete any extraneous prints/debugging code?


sim/midas/src/main/cc/bridges/peek_poke.h
```diff
@@ -1,18 +1,18 @@
...
+#include "simif.h"
 
 typedef struct PEEKPOKEBRIDGEMODULE_struct {
   uint64_t PRECISE_PEEKABLE;
   uint64_t READY;
 } PEEKPOKEBRIDGEMODULE_struct;
 
-PEEKPOKEBRIDGEMODULE_checks;
-
 // This derivation of simif_t is used to implement small integration
```
davidbiancolin, line 16: This comment needs to be updated

sim/midas/src/main/cc/bridges/peek_poke.h
```diff
@@ -29,43 +29,54 @@ PEEKPOKEBRIDGEMODULE_checks;
...
+  ~peek_poke_t() {}
+
+  void poke(std::string_view id, uint32_t value, bool blocking);
+  void poke(std::string_view id, mpz_t &value);
+
+  uint32_t peek(std::string_view id, bool blocking);
+  void peek(std::string_view id, mpz_t &value);
+  uint32_t sample_value(std::string_view id) { return peek(id, false); }
+
+  /// Returns true if the last request times out.
```
davidbiancolin, line 60: javadoc

davidbiancolin (commented):
So, peek-poke can't be optional until it's no longer responsible for managing time-control in simif.  take_steps writes to the master widget, which is for vestigial reasons connected to peek poke. So we always need it. 

I think a very reasonable way to break this linkage and address a long standing source of technical debt is to have `take_steps` give credits to the clock bridge. Since simif has a reference to the clock bridge it's pretty easy to do this, we'd just need to add some new MMIO registers to the clock bridge.

That aside the change to make peek_poke treated more conventionally is fine (the bulk of this PR can remain) but you need to ensure it's always instantiated once. Once it's been decoupled from simulation time control we should probably ensure it's only used in testing and not in firesim_top.

sifive-benjamin-morse (approved):
Looks Good. Do you know how much of a performance hit is added by changing `peek` and `poke` to use a strings and a std::map?

PR #1374, opened 2023-01-07
Separate XSIM from f1 into simif_xsim

nandor:
The XSIM implementation is largely disjoint from the F1 one. Separate it. This greatly simplifies the critical F1 implementation, allowing XSIM to be build alongside it.

### Contributor Checklist

### DIFF
--- a/sim/Makefile
+++ b/sim/Makefile
@@ -43,10 +43,11 @@ include $(TARGET_PROJECT_MAKEFRAG)
 include make/config.mk
 include make/library.mk
 include make/goldengate.mk
-include make/fpga.mk
 include make/verilator.mk
 include make/vcs.mk
 include make/driver.mk
+include make/fpga.mk
+include make/xsim.mk
 include make/unittest.mk
 include make/scala.mk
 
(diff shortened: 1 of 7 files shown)


davidbiancolin (changes requested):
So "simif_xsim" is very closely tied to F1. It would be preferable to have the xsim implementation be built alongside the deployment driver, and selected at runtime if possible, instead of separating them into separate targets in the Makefile. This is more consistent with how vitis works, and minimizes the user-facing + build system changes until we can figure out how to more intelligently support FPGA-level metasimulation from a build-system perspective.

nandor:
In its present form, `simif_f1` is bad engineering. Based on the platform-specific flag, both the method implementations and the class members are different and almost fully disjoint. The executables produced are still the same as the change is hidden behind the Makefile in the driver directory. The PLATFORM flag is the same, we simply choose a different simif file. We do not need any conditional compilation in driver code and splitting stuff into two classes fully solves that. In a subsequent change, the main method will be de-duplicated as well and the scope of `simif` will be almost narrowed to handling the interface to hardware/simulation, as it should be.

nandor:
Thanks for specifying the new name. I am sorry for causing stress due to this process, however bugs are being fixed and technical debt is being burnt off.

nandor:
I can't get xsc/xvlog to work on internal machines, but the flow works up until that point and the binary is build and reaches the point of opening the file handles. No documentation changes are needed.

davidbiancolin (approved):

PR #1380, opened 2023-01-11
Add a test for existing TracerV bridge

sifive-benjamin-morse:
This is a draft.

# Status
* This fuzzes valid on the TracerV with a known instruction.
* This adds a callback to tracerv.cc, which will handle the traces
* The test uses the callback to operate

# Problems
* I need help to understand where I should put the tracefile output?
* I need help to understand the logic in process_tokens around fireperf (From @davidbiancolin )
* This needs to be rebased to main 
* This uses the simif seeded random however I don't know how to run more seeds
* This doesn't test the `TriggerSink` I need help from David understanding what this test would look like

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
@@ -230,7 +230,6 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
   // check that a tracefile exists (one is enough) since the manager
   // does not create a tracefile when trace_enable is disabled, but the
   // TracerV bridge still exists, and no tracefile is created by default.
-  // std::cout << "this->tracefile " << this->tracefile << "\n";
   if (this->tracefile) {
     if (this->human_readable || this->test_output) {
       for (int i = 0; i < (bytes_received / sizeof(uint64_t)); i += 8) {
@@ -254,6 +253,13 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
                       OUTBUF[i + q + 1] & (~valid_mask));
             } else {
               break;
+            }
+          }
+        }
+        if(callback) {
+          for (int q = 0; q < max_core_ipc; q++) {
+            if (OUTBUF[i + q + 1] & valid_mask) {
+              callback(OUTBUF[i + 0], OUTBUF[i + q + 1] & (~valid_mask));
             }
           }
         }
@@ -304,4 +310,13 @@ void tracerv_t::flush() {
   pull_flush(stream_idx);
   while (this->trace_enabled && (process_tokens(this->stream_depth, 0) > 0))
     ;
+}
+
+/**
+ * Set a callback that will fire with each traced instructions
+ * The first argument is the cycle, the second is the PC. Multiple calls with the same value for cycle
+ * indicates that multiple instructions landed on that cycle
+ */
+void tracerv_t::set_callback(std::function<void(uint64_t, uint64_t)> cb) {
+  callback = cb;
 }
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
@@ -8,10 +8,10 @@
 #include <string.h>
 
 #include <fcntl.h>
+#include <iostream>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-#include <iostream>
 
 #include <sys/mman.h>
 
@@ -256,7 +256,7 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
             }
           }
         }
-        if(callback) {
+        if (callback) {
           for (int q = 0; q < max_core_ipc; q++) {
             if (OUTBUF[i + q + 1] & valid_mask) {
               callback(OUTBUF[i + 0], OUTBUF[i + q + 1] & (~valid_mask));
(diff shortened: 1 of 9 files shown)


sim/firesim-lib/src/main/cc/bridges/tracerv.cc
```diff
@@ -285,6 +296,10 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
 }
 
 void tracerv_t::tick() {
+  if (!init_ran) {
```
nandor, line 299: This is an invariant which should be respected by all drivers and the message is not too relevant at runtime. It should be an assertion.

sim/firesim-lib/src/main/cc/bridges/tracerv.h
```diff
@@ -98,6 +104,8 @@ class tracerv_t : public bridge_driver_t
   std::string tracefilename;
   std::string dwarf_file_name;
   bool fireperf = false;
+  bool init_ran = false;
+  std::function<void(uint64_t, uint64_t)> callback = NULL;
```
nandor, line 108: Can we name this `on_instruction_received`? Indicate what the callback is providing in the name?

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,265 @@
...
+    // modified as we go
+    uint64_t e_cycle = cycle_offset;
+
+    const unsigned tracerv_width = tracerv->max_core_ipc;
+
+    // load MMIO and capture expected outputs
+    auto load = [&](std::vector<uint64_t> iad, std::vector<bool> bit) {
+      assert(iad.size() == bit.size());
+      const auto sz = iad.size();
+      for (unsigned i = 0; i < sz; i++) {
```
nandor, line 187: nit: In LLVM, we usually write: `for (unsigned i = 0, sz = iad.size(); i < sz; ++i)`

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,265 @@
...
+      const uint64_t pull = simif->rand_next(tracerv_width);
+
+      auto pull_iaddr = get_iaddrs(test_step, tracerv_width);
+      auto pull_bits = get_contiguous(pull, tracerv_width);
+
+      load(pull_iaddr, pull_bits);
+      steps(1);
+    }
+
+    auto [final_iaddr, final_valid] = get_final_values(tracerv_width);
```
nandor, line 215: If the return type is a non-primitive, it's best to capture it as a const-ref: `const auto &[final_iaddr, final_valid]`

nandor (commented):
In terms of the architecture of the test: we are adding extra functionality to the bridge just to test it. Instead, I see that we are piping a stream of data through the peek-poke bridge, then validating it afterwards. Can we generate & verify the stimulus on the Scala side? By building a file from which instructions are loaded and checking the tracefile against the data.

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
sifive-benjamin-morse: I'm confused? It's not an invariant because the call of `init()` happens outside the constructor. The way firesim is setup, it's required that `init()` is called before any ticks. I added this because I personally ran into an issue where I forgot to call init and got weird behavior. I forgot because I'm still relatively a new user to firesim, and this is the first bridge test I've written after all of your changes. I look at it like guard rails for developers. I agree it should be moved higher, maybe into `bridge_driver_t` but I'm not sure how.

sifive-benjamin-morse:
> architecture of the test: we are adding extra functionality to the bridge just to test it. Instead, I see that we are piping a stream of data through the peek-poke bridge, then validating it afterwards. Can we generate & verify the stimulus on the Scala side? By



> In terms of the architecture of the test: we are adding extra functionality to the bridge just to test it. Instead, I see that we are piping a stream of data through the peek-poke bridge, then validating it afterwards. Can we generate & verify the stimulus on the Scala side? By building a file from which instructions are loaded and checking the tracefile against the data.

I don't see a problem with adding functionality to TracerV? Before my change, there was no way to get the trace data from inside C++. This seems like a major oversight in the design of TracerV. Regarding the tests from Scala, @davidbiancolin asked me to leave it in C++. I'll leave it up to him if he still feels this way after reviewing the code

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
davidbiancolin: Nandor's refactoring should make it very difficult to forget to call Init because the control flow is more prescribed. 

Perhaps a better way to do this would be to move it into a parent class (i.e, bridge_driver_t), so that this invariant is checked across all bridges. Maybe pull this change out of this PR and do a follow up? @nandor thoughts?

davidbiancolin:
There are a bunch of out-of-date files that may explain the misunderstanding vis-a-vis the init() application?

sifive-benjamin-morse:
@nandor After talking with David, we came up with this plan and wanted your comments:
* remove the callback
* replace fprintf(this->tracefile) with an std::ostringstream
* add a function so that my test can overwrite the std::ostringstream (and capture it's output)
* now the C++ test can test all 3 modes of writing files internally

Any thoughts?

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
nandor: I think we can do away with the check here completely. If `init` allocated some data or created some config vars I'd add an assertion, but there's really no point in adding a flag just for this.

nandor:
If you use a `std::ostream`, you need to either pass it in the constructor or allow it to be configured.

The first approach is problematic as bride construction requires information from the header and a mechanism to let users instantiate bridges. This is not something I would expose to users without carefully thinking of a specific interface.
I'm not to keen to allow the stream to be overriden by a method, as that's yet another bit of functionality we need to support. I'd rather not do this in bridge tests now.

The file format is pretty much defined by a single `fprintf`. A way that doesn't require changes to the bridge or the definition of new mechanisms is to let the C++ harness print the expected output based on the input it passes to the bridge and diff the files in the scala harness (as it is done with synthesized prints and auto counters). The scala harness would pass a path for the test to write to. Could also add a:

```
static void print_traced_instruction(std::ostream &os, uint64_t cycle, uint64_t op);
``

to de-duplicate the printing logic.

PR #1383, opened 2023-01-14
Expand TracerV to support more than 7 IPC

sifive-benjamin-morse:
# TracerV can support more than 7 IPC
Multiple beats are used to send instruction trace information.

#### UI / API Impact
The UI impact should be invisible. However chips with more than 7 IPC are now supported.

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
@@ -271,8 +271,6 @@ void tracerv_t::serialize(
                     OUTBUF[i + 0],
                     q,
                     OUTBUF[i + q + 1] & (~valid_mask));
-          } else {
-            break;
           }
         }
       }
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.cc
@@ -35,9 +35,6 @@ tracerv_t::tracerv_t(simif_t &sim,
     : streaming_bridge_driver_t(sim, stream, &KIND), mmio_addrs(mmio_addrs),
       stream_idx(stream_idx), stream_depth(stream_depth),
       max_core_ipc(max_core_ipc), clock_info(clock_info) {
-  // Biancolin: move into elaboration
-  assert(this->max_core_ipc <= 7 &&
-         "TracerV only supports cores with a maximum IPC <= 7");
   const char *tracefilename = nullptr;
   const char *dwarf_file_name = nullptr;
   this->tracefile = nullptr;
--- a/sim/firesim-lib/src/main/cc/bridges/tracerv.h
+++ b/sim/firesim-lib/src/main/cc/bridges/tracerv.h
@@ -5,7 +5,6 @@
 #include "core/bridge_driver.h"
 #include "core/clock_info.h"
 #include <functional>
-#include <iostream>
 #include <vector>
 
 class TraceTracker;
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -247,11 +247,11 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
 
     streamEnq.bits := streamMux
 
-    val maybeFire = !allValidMux
+    val maybeFire = !allValidMux || (counter === (armCount-1).U)
     val maybeEnq  = anyValidMux
 
     val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, anyValidMux, traceEnable)
-    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire, traceEnable)
+    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire)
 
     // Note, if we dequeue a token that wins out over the increment below
     when(do_fire_helper.fire()) {
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -210,7 +210,7 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
 
     hPort.fromHost.hValid := tFireHelper.fire(hPort.fromHost.hReady)
 
-    // the maximum widht of a single arm, this is determined by the 512 bit width of a single beat
+    // the maximum width of a single arm, this is determined by the 512 bit width of a single beat
     val armWidth = 7
     
     // divide with a ceiling round, to get the total number of arms
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -250,8 +250,8 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
     val maybeFire = !allValidMux
     val maybeEnq  = anyValidMux
 
-    val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, anyValidMux)
-    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire)
+    val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, anyValidMux, traceEnable)
+    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire, traceEnable)
 
     // Note, if we dequeue a token that wins out over the increment below
     when(do_fire_helper.fire()) {
@@ -260,7 +260,7 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
       counter := counter + 1.U
     }
 
-    streamEnq.valid := do_enq_helper.fire(streamEnq.ready)
+    streamEnq.valid := do_enq_helper.fire(streamEnq.ready, trigger)
     hPort.toHost.hReady := do_fire_helper.fire(hPort.toHost.hValid)
 
     when (hPort.toHost.fire) {
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -215,10 +215,6 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
     
     // divide with a ceiling round, to get the total number of arms
     val armCount = (traces.length + armWidth - 1) / armWidth
-    val armCountU = (armCount).U
-
-    // A sequence with the number of traces in each arm
-    val armWidths = Seq.tabulate(armCount)(x => math.min(traces.length - (x * armWidth), armWidth))
 
     // A Seq of Seq which represents each arm of the mux
     val allTraceArms = traces.grouped(armWidth).toSeq
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -13,34 +13,33 @@ import midas.widgets._
 
 case class TracerVKey(
   insnWidths: TracedInstructionWidths, // Widths of variable length fields in each TI
-  vecSize: Int // The number of insns in the traced insn vec (= max insns retired at that core) 
+  vecSize:    Int,                     // The number of insns in the traced insn vec (= max insns retired at that core)
 )
 
-
 class TracerVTargetIO(insnWidths: TracedInstructionWidths, numInsns: Int) extends Bundle {
-  val trace = Input(new TileTraceIO(insnWidths, numInsns))
+  val trace         = Input(new TileTraceIO(insnWidths, numInsns))
   val triggerCredit = Output(Bool())
-  val triggerDebit = Output(Bool())
+  val triggerDebit  = Output(Bool())
 }
-/**
-  * Target-side module for the TracerV Bridge.
+
+/** Target-side module for the TracerV Bridge.
   *
-  * @param insnWidths A case class containing the widths of configurable-length
-  * fields in the trace interface.
+  * @param insnWidths
+  *   A case class containing the widths of configurable-length fields in the trace interface.
   *
-  * @param numInsns The number of instructions captured in a single a cycle
-  * (generally, the commit width of the pipeline)
+  * @param numInsns
+  *   The number of instructions captured in a single a cycle (generally, the commit width of the pipeline)
   *
-  * Warning: If you're not going to use the companion object to instantiate
-  * this bridge you must call [[TracerVBridge.generateTriggerAnnotations] _in
-  * the parent module_.
+  * Warning: If you're not going to use the companion object to instantiate this bridge you must call
+  * [[TracerVBridge.generateTriggerAnnotations] _in the parent module_.
   */
-class TracerVBridge(insnWidths: TracedInstructionWidths, numInsns: Int) extends BlackBox
+class TracerVBridge(insnWidths: TracedInstructionWidths, numInsns: Int)
+    extends BlackBox
     with Bridge[HostPortIO[TracerVTargetIO], TracerVBridgeModule] {
   require(numInsns > 0, "TracerVBridge: number of instructions must be larger than 0")
-  val io = IO(new TracerVTargetIO(insnWidths, numInsns))
-  val bridgeIO = HostPort(io)
-  val constructorArg = Some(TracerVKey(insnWidths, numInsns))
+  val io                                 = IO(new TracerVTargetIO(insnWidths, numInsns))
+  val bridgeIO                           = HostPort(io)
+  val constructorArg                     = Some(TracerVKey(insnWidths, numInsns))
   generateAnnotations()
   // Use in parent module: annotates the bridge instance's ports to indicate its trigger sources
   // def generateTriggerAnnotations(): Unit = TriggerSource(io.triggerCredit, io.triggerDebit)
@@ -71,7 +70,7 @@ object TracerVBridge {
     tracerv
   }
 
-  def apply(tracedInsns: TileTraceIO)(implicit p:Parameters): TracerVBridge = {
+  def apply(tracedInsns: TileTraceIO)(implicit p: Parameters): TracerVBridge = {
     val tracerv = withClockAndReset(tracedInsns.clock, tracedInsns.reset) {
       TracerVBridge(tracedInsns.insnWidths, tracedInsns.numInsns)
     }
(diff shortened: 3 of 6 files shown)


davidbiancolin (commented):
Were sitting in draft...

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -198,21 +208,70 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val armCountU = (armCount).U
+
+    // A sequence with the number of traces in each arm
+    val armWidths = Seq.tabulate(armCount)(x => math.min(traces.length - (x * armWidth), armWidth))
+
+    // A Seq of Seq which represents each arm of the mux
+    val allTraceArms = Seq.tabulate(armCount)(x => {Seq.tabulate(armWidths(x))(y => {traces(y + (armWidth*x))})})
+
+    // an intermediate value used to build allStreamBits
+    val allUintTraces = allTraceArms.map(arm=>arm.map((trace => Cat(trace.valid, trace.iaddr).pad(64))).reverse)
```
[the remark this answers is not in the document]
sifive-benjamin-morse: I don't understand? `allUintTraces` is the bits for a single arm. if I were to calculate this before the arms it doesn't make sense to me? what am I missing?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -198,21 +208,70 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val counterBits = log2Ceil(armCount+1)
+
+    // This counter acts to select the mux arm
+    val counter = RegInit(0.U(counterBits.W))
+    
+    // The main mux where the input arms are different possible valid traces, and the output goes to streamEnq
+    val streamMux = MuxLookup(counter, allStreamBits(0), Seq.tabulate(armCount)(x=>x.U->allStreamBits(x)))
+    
+    // a parallel set of arms to a parallel mux, true if any instructions in the arm are valid (OR reduction)
+    val anyValid = allTraceArms.map(arm=>arm.map(trace => trace.valid).reduce((a,b) => (a|b)))
```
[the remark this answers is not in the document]
sifive-benjamin-morse: This works, however it doesn't provide a default value (at least when I tried it). The code you gave me, requires that a default value of 0 occurs when the counter counts past the highest arm

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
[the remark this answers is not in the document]
davidbiancolin: What i was suggesting was doing the Cat + pad before you call group so you don't have to map twice, which would make it slightly more readable. Your call. I should've explained why i was suggesting what i was suggesting.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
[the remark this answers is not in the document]
davidbiancolin: True

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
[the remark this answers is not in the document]
sifive-benjamin-morse: I think I prefer to leave it as it is, it's closer to the original design before I touched it

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
```diff
@@ -232,64 +228,100 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
...
   }
-  return bytes_received;
+}
+
+void tracerv_t::write_header(FILE *file) {
+  fputs(this->clock_info.file_header().c_str(), file);
 }
 
 void tracerv_t::tick() {
+  if (!init_ran) {
```
nandor, line 321: This can be removed - the surrounding system should ensure that init is called before tick.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -198,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+      counter := 0.U 
+    }.elsewhen(do_enq_helper.fire()) {
+      counter := counter + 1.U
+    }
+
+    streamEnq.valid := do_enq_helper.fire(streamEnq.ready)
+    hPort.toHost.hReady := do_fire_helper.fire(hPort.toHost.hValid)
 
-    when (tFireHelper.fire()) {
+    when (hPort.toHost.hValid && hPort.toHost.hReady) {
```
nandor, line 266: Isn't this equivalent to `hPort.toHost.fire`?

sim/src/main/makefrag/midasexamples/driver.mk
```diff
@@ -13,6 +13,7 @@ DRIVER_CC := \
 TARGET_CXX_FLAGS := \
 		-I$(driver_dir) \
 		-I$(driver_dir)/midasexamples \
+		-I$(firesim_lib_dir) \
```
nandor, line 16: We shouldn't have a dependency on the firesim bridges from the base library.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+  return ss.str();
+};
+
+static std::string namev(const unsigned x) {
+  std::stringstream ss;
+  ss << "io_insns_" << x << "_valid";
+  return ss.str();
+};
+
+class TracerVModule : public TestHarness {
```
nandor, line 48: We are not reusing much of the `midasexamples` test harness, but we are adding a cross-project dependency. Please inherit from `simulation_t` directly and duplicate the declaration of the random seed and random number generator, as argument parsing is already done here as well. This test only uses the peek-poke bridge to push in data, we don't care much about inspecting individual pins with the wrappers provided by `TestHarness`.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+
+    return was_done;
+  }
+
+  std::vector<std::pair<uint64_t, std::vector<uint64_t>>> expected_pair;
+
+  void run_test() override {
+
+    // Reset the DUT.
+    poke("reset", 1, /*blocking=*/true);
```
nandor, line 127: ```peek_poke.poke("reset", 1, /*blocking=*/true);```

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+      // because expected_pair doesn't contain the instruction value for non
+      // valid instructions only the human readable output will compare
+      // correctly
+      tracerv_t::serialize(OUTBUF,
+                           sizeof(OUTBUF),
+                           expected,
+                           nullptr,
+                           tracerv.max_core_ipc,
+                           true,
+                           false,
```
nandor, line 219: If the arguments are true/false or integral constants, please add a `/*argname=*/` comment before them.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Returns the next available random number, modulo limit.
+   */
+  uint64_t rand_next(uint64_t limit) { return gen() % limit; }
+
+private:
+  unsigned get_total_trace_tests() const { return 128; }
+
+  // random numbers
+  uint64_t seed = 0;
```
nandor, line 235: This already duplicates the random logic - there's really no reason left to inherit from the other test harness.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+   * Writes the contents of the member variable expected_pair
+   * to the expected file pointer. This is done via the tracerv_t::serialize
+   * function.
+   */
+  void write_expected_file() {
+    uint64_t OUTBUF[8];
+
+    for (const auto &beat : expected_pair) {
+      const auto &[cycle, insns] = beat;
+      memset(OUTBUF, 0, sizeof(OUTBUF));
```
nandor, line 203: There's no reason to 0-initialize here. The scope of OUTBUF could be narrowed into the for loop. It could also be renamed into `buffer` since it is not a constant or special variable.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+      OUTBUF[0] = cycle;
+      assert(insns.size() < 8);
```
nandor, line 205: ```assert(insns.size() < 8 && "give a reason for the failure");```

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,242 @@
...
+      for (int i = 0; i < insns.size(); i++) {
```
nandor, line 206: I'd make the bufffer a `std::vector` and `push_back` elements into it.

sim/firesim-lib/src/main/cc/bridges/tracerv.h
```diff
@@ -51,12 +51,25 @@ class tracerv_t : public streaming_bridge_driver_t {
   virtual bool terminate() { return false; }
   virtual int exit_code() { return 0; }
   virtual void finish() { flush(); };
+  static void serialize(const uint64_t *const OUTBUF,
```
nandor, line 54: Can we coalesce `OUTBUF` and `bytes_received` into a `const std::vector<uint64_t>`?

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
```diff
@@ -232,64 +228,100 @@ size_t tracerv_t::process_tokens(int num_beats, int minimum_batch_beats) {
...
+
+void tracerv_t::serialize(
+    const uint64_t *const OUTBUF,
+    const size_t bytes_received,
+    FILE *tracefile,
+    std::function<void(uint64_t, uint64_t)> addInstruction,
+    const int max_core_ipc,
+    const bool human_readable,
+    const bool test_output,
+    const bool fireperf) {
```
nandor, line 259: Arguments to functions should not be `const` - `clang-tidy` catches this now.

nandor, line 259: Isn't fireperf tied to `addInstruction` somehow? Could we get rid of this boolean flag and rely on `addInstruction` being set or not? Also, would it be possible to pull the `addInstruction` logic completely out? It is not required for the serialization of data coming from the test.

nandor (changes requested):
I'm mostly concerned with the dependency on `midasexamples` and the serialize method at the moment. The dependency should be avoided and we can restructure the stuff within serialize to make it a bit safer.

sim/src/main/cc/bridges/TracerVModule.cc
sifive-benjamin-morse: refactoring the body of serialize into using modern std::vector is outside the scope of this PR

sim/firesim-lib/src/main/cc/bridges/tracerv.h
sifive-benjamin-morse: refactoring the body of serialize into using modern std::vector is outside the scope of this PR

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
sifive-benjamin-morse: refactoring the body of serialize into using modern std::vector is outside the scope of this PR

sim/src/main/makefrag/bridges/driver.mk
```diff
@@ -13,13 +13,16 @@ DRIVER_H = $(shell find $(driver_dir) -name "*.h")
 
 DRIVER_CC := \
 		$(driver_dir)/bridges/BridgeHarness.cc \
+		$(driver_dir)/midasexamples/TestHarness.cc \
```
nandor, line 16: This is not needed here, now that `TracerVModule` inherits from `simulation_t`.

nandor:
Rebased onto main. Managed to run the verilator test locally and it passed. Not sure why CI is failing.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
     hPort.hBits.triggerCredit := trigger && !triggerReg
 
-    val uint_traces = (traces map (trace => Cat(trace.valid, trace.iaddr).pad(64))).reverse
-    streamEnq.bits := Cat(uint_traces :+ trace_cycle_counter.pad(64)).pad(BridgeStreamConstants.streamWidthBits)
-
-    hPort.toHost.hReady := tFireHelper.fire(hPort.toHost.hValid)
     hPort.fromHost.hValid := tFireHelper.fire(hPort.fromHost.hReady)
 
-    streamEnq.valid := tFireHelper.fire(streamEnq.ready, trigger) && traceEnable
+    // the maximum widht of a single arm, this is determined by the 512 bit width of a single beat
```
nandor, line 213: nit: width

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val armWidth = 7
+    
+    // divide with a ceiling round, to get the total number of arms
+    val armCount = (traces.length + armWidth - 1) / armWidth
+    val armCountU = (armCount).U
```
nandor, line 218: Parenthesis not necessary. I'd also prefer to remove this variable.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+
+    // A sequence with the number of traces in each arm
+    val armWidths = Seq.tabulate(armCount)(x => math.min(traces.length - (x * armWidth), armWidth))
+
+    // A Seq of Seq which represents each arm of the mux
+    val allTraceArms = traces.grouped(armWidth).toSeq
+
+    // an intermediate value used to build allStreamBits
+    val allUintTraces = allTraceArms.map(arm=>arm.map((trace => Cat(trace.valid, trace.iaddr).pad(64))).reverse)
```
nandor, line 227: Can you rune scalafmt?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    streamEnq.bits := streamMux
+
+    val maybeFire = !allValidMux
+    val maybeEnq  = anyValidMux
+
+    val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, anyValidMux, traceEnable)
+    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire, traceEnable)
+
+    // Note, if we dequeue a token that wins out over the increment below
+    when(do_fire_helper.fire()) {
```
nandor, line 257: unnecessary `()`

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    
+    val allValid = allTraceArms.map(arm=>arm.map(trace => trace.valid).reduce((a,b) => (a&b)))
+    val allValidMux = MuxLookup(counter, false.B, Seq.tabulate(armCount)(x=>x.U->allValid(x)))
+
+    streamEnq.bits := streamMux
+
+    val maybeFire = !allValidMux
+    val maybeEnq  = anyValidMux
+
+    val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, anyValidMux, traceEnable)
```
nandor, line 253: Instead of defining those two variables which rely on anyValidMux, I think the code would be much clearer if some comment explained when we can enqueue items and when does the queue fire.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire, traceEnable)
+
+    // Note, if we dequeue a token that wins out over the increment below
+    when(do_fire_helper.fire()) {
+      counter := 0.U 
+    }.elsewhen(do_enq_helper.fire()) {
```
nandor, line 259: Here we don't use the `excludes`/`includes` functionality of the `DecoupledHelper`, which means that the wrappers simply and all arguments. Can we have that in the code? I usually find the use of this class confusing.

sifive-benjamin-morse: I'll let @davidbiancolin address the usage of `DecoupledHelper` here.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
sifive-benjamin-morse: Both armCount and armCountU are required for operation. Perhaps you have a better name I can use?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
sifive-benjamin-morse: @davidbiancolin What is he talking about?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
sifive-benjamin-morse: I ran `scalafmtAll`

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: generating the `UInt` literal in situ with `.U` is more idiomatic Chisel, so you should just do that.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: It still skips this file to maintain my sanity doing upstream merges. You should able to run it on lines that change manually though (i don't personally know how, but there's a script in our codebase that does that, message me out-of-band

davidbiancolin: But generally there should be spaces between operators, notably `=>`

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val allStreamBits = allUintTraces.map(uarm=>Cat(uarm :+ trace_cycle_counter.pad(64)).pad(BridgeStreamConstants.streamWidthBits))
+
+    // Number of bits to use for the counter, the +1 is required because the counter will count 1 past the number of arms
+    val counterBits = log2Ceil(armCount+1)
+
+    // This counter acts to select the mux arm
+    val counter = RegInit(0.U(counterBits.W))
+    
+    // The main mux where the input arms are different possible valid traces, and the output goes to streamEnq
+    val streamMux = MuxLookup(counter, allStreamBits(0), Seq.tabulate(armCount)(x=>x.U->allStreamBits(x)))
```
davidbiancolin, line 239:
```suggestion
    val streamMux = MuxLookup(counter, allStreamBits(0), Seq.tabulate(armCount)(x => x.U -> allStreamBits(x)))
```
for example

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -208,15 +208,62 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    streamEnq.bits := streamMux
+
+    val maybeFire = !anyValidRemainMux || (counter === (armCount-1).U)
+    val maybeEnq  = anyValidRemainMux
+
+    val do_enq_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeEnq, traceEnable)
+    val do_fire_helper = DecoupledHelper(hPort.toHost.hValid, streamEnq.ready, maybeFire)
+
+    // Note, if we dequeue a token that wins out over the increment below
+    when(do_fire_helper.fire()) {
```
davidbiancolin, line 257:
```suggestion
    when(do_fire_helper.fire) {
```

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: I believe removing the parens on nullary functions is now standard under 2.13, even for functions with side-effects (i.e., all of chisel...)

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: See above

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: If you like you can assign `fire` to a wire, but I think this is fine. DecoupledHelper is the right tool for the job.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -10,10 +10,10 @@
 #include <iostream>
 #include <string_view>
 
-static std::vector<bool> get_contiguous(unsigned bits, unsigned total) {
+static std::vector<bool> get_valids(unsigned seed, unsigned total) {
   std::vector<bool> ret;
   for (unsigned i = 0; i < total; i++) {
-    const bool value = (i < bits);
+    const bool value = seed & (1 << i);
```
davidbiancolin, line 16: So is it true that this will only produce inputs with a single valid instruction or am i misunderstanding? Could you not just call out to an RNG at this point?

davidbiancolin (commented):
Mostly just style nits. Can you explain to me in a little more detail how you tested non-contiguous inputs?

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
sifive-benjamin-morse: But bloop throws a warning. sorry I didn't include this in my original comment:
```
Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method fire,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function.bloop
```

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
sifive-benjamin-morse: But bloop throws a warning. sorry I didn't include this in my original comment:
```
Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method fire,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function.bloop
```

sim/src/main/cc/bridges/TracerVModule.cc
sifive-benjamin-morse: No, this code sweeps a 1 bit mask through seed. This is the core mechanic to turn a uint into a vector of bools.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: Derp, I got it twisted and the compiler is always right. The infix notation (i.e, no ()) is gone i think for functions with default arguments.

sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
davidbiancolin: Leave this as is per ^

sim/src/main/cc/bridges/TracerVModule.cc
davidbiancolin: Ok right.  It's a shame there's not a convenient one liner for this. Maybe you want an assert to check the requested total is less than the bitwidth of the seed.

nandor (approved):
LGTM

PR #1390, opened 2023-01-18
Cache the classpath between SBT runs

nandor:
### Contributor Checklist

### DIFF
--- a/sim/make/scala-build.mk
+++ b/sim/make/scala-build.mk
@@ -1,0 +1,141 @@
+# See LICENSE for license details.
+
+################################################################################
+#
+# This file defines rules to build the Scala components of FireSim.
+#
+# The file provides 3 targets caching the classpath containing compiled jars
+# to be used by subsequent java invocations, fully bypassing the slow sbt setup
+# on repeated invocations to SBT.
+#
+# The 3 cached classpaths are:
+#
+#  FIRESIM_MAIN_CP: contains the GoldenGate entry point.
+#
+#  FIRESIM_TEST_CP: contains generators and configurations for most tests.
+#
+#  TARGET_CP: contains the entry point of the target Chisel generator.
+#             It is built out of the TARGET_SBT_PROJECT. TARGET_SOURCE_DIRS
+#             must enumerate all the directories containing the scala sources
+#             to compile this target.
+#
+# Any rule containing a Java invocation must add as a dependency one of these
+# classpaths. Corresponding SBT invocations still work (substitute the classpath
+# with the correct project).
+#
+################################################################################
+
+################################################################################
+# Helper to run SBT in the project.
+################################################################################
+
+SBT_COMMAND ?= shell
+SBT_NON_THIN ?= $(subst $(SBT_CLIENT_FLAG),,$(SBT))
+.PHONY: sbt
+sbt:
+	cd $(base_dir) && $(SBT_NON_THIN) ";project $(firesim_sbt_project); $(SBT_COMMAND)"
+
+################################################################################
+# Target Configuration
+################################################################################
+
+# Target project containing the generator of the design.
+TARGET_SBT_PROJECT ?= $(firesim_sbt_project)
+
+# If the target project is not the implicit FireSim one, this definition must
+# enumerate all directories containing the sources of the generator.
+TARGET_SOURCE_DIRS ?=
+
+################################################################################
+# Build jars using SBT assembly and cache them.
+################################################################################
+
+# Helper to build a classpath for a project. This first builds the project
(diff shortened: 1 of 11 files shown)


davidbiancolin (commented):
Nice changes. So really the performance improvement comes from eliding the extra SBT invocations right?

sim/make/build.mk
```diff
@@ -0,0 +1,87 @@
+# See LICENSE for license details.
+
+################################################################################
+# Helper to run SBT in the project.
+################################################################################
+
+SBT_COMMAND ?= shell
+SBT_NON_THIN ?= $(subst $(SBT_CLIENT_FLAG),,$(SBT))
```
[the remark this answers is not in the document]
abejgonzalez: You can just run with `ENABLE_SBT_THIN_CLIENT=1` (you can export this as well)

sim/make/build.mk
```diff
@@ -0,0 +1,87 @@
...
+			midas \
+			midas/targetutils \
+			firesim-lib\
+	)
+firesim_main_srcs = $(foreach submodule, $(firesim_source_dirs),\
+	$(shell find $(submodule)/ -iname "[!.]*.scala" -print 2> /dev/null | grep -E 'src/main/scala'))
+firesim_test_srcs = $(foreach submodule, $(firesim_source_dirs),\
+	$(shell find $(submodule)/ -iname "[!.]*.scala" -print 2> /dev/null | grep -E 'src/test/scala'))
+
+TARGET_SBT_PROJECT ?= $(firesim_sbt_project)
```
abejgonzalez, line 30: Isn't this set in the `TARGET_PROJECT_MAKEFRAG` (e.g. `src/main/makefrag/firesim/config.mk`)? Why does this need to be potentially set again here?

sim/make/build.mk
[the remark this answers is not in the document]
abejgonzalez: Although, last time I tried running the SBT thin client with FireSim, I noticed that make rebuilt things unnecessarily.

sim/make/build.mk
```diff
@@ -0,0 +1,87 @@
...
+# Build jars using SBT assembly and cache them.
+################################################################################
+
+define build_classpath
+	$(SBT_NON_THIN) \
+		"set showSuccess := false; project $(1); compile; package"
+	$(SBT_NON_THIN) \
+		--error \
+		"set showSuccess := false; project $(1); compile; package; export $(2):fullClasspath" \
+		| head -n 1 | tr -d '\n' > $@
```
abejgonzalez, line 48: Does this have to be done without the thin client?

abejgonzalez (commented):
Can you describe a bit more about what this PR is and how it interacts with the current build flow in the PR description? Can you describe the difference between this and just constructing a fat jar (I'm not sure what happened to that work).

Additionally, double-checking that the thin client mode working with this would be good.

sim/make/build.mk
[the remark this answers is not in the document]
nandor: Last time I tried running that things failed pretty badly (output got mangled from multiple threads and all the ascii colouring escape strings showed up in an unreadable way, the build also got stuck).

sim/make/build.mk
nandor: For projects that build tests, this defaults to the value I set here.

sim/make/build.mk
nandor: I don't know. But this is what's been used before, so it's not changing here.

nandor:
I could never get the thin client to work in the first place. Nor could I get assembled jars to work. These require additional changes to maintain, while the classpath-based solution works as long as the sbt-based rules work. This PR does not change the build flow or the build commands to be executed.

sim/make/scala-build.mk
```diff
@@ -0,0 +1,141 @@
...
+# The 3 cached classpaths are:
+#
+#  FIRESIM_MAIN_CP: contains the GoldenGate entry point.
+#
+#  FIRESIM_TEST_CP: contains generators and configurations for most tests.
+#
+#  TARGET_CP: contains the entry point of the target Chisel generator.
+#             It is built out of the TARGET_SBT_PROJECT. TARGET_SOURCE_DIRS
+# 					  must enumerate all the directories containing the scala sources
+# 					  to compile this target.
```
abejgonzalez, line 20: Nit: Spacing is off

abejgonzalez (approved):
LGTM. I think it would be worth adding a Make error if the `ENABLE_SBT_THIN_CLIENT` flag is broken in FireSim. I know many people who auto-enable this flag in their shell config so having a nice error would be useful.

Issue #901, opened 2021-12-15
Metasimulation broken under Modern VCS Releases


davidbiancolin:
It appears that a some version between after L-2016.06-SP2-4_Full64 and M-2017.03-SP2_Full64, VCS started clobbering older allocations to the heap made by the driver (we enter through the driver before switching into vcs_main). This manifests first in segmentation faults when referencing mmio_t and mm_t datastructures used in the DPI call to switch back to the driver context.

I've reproduced this on millennium (Ubuntu) and sifive (RHEL) machines for a variety of different VCS versions.  I'm going to try reworking VCS-based metasimulation setup to use VCS's main as the entrypoint for the program instead of the driver, but i'm worried that later heap allocations in either context could create problems. 

h/t @aswaterman

timsnyder-siv:
@davidbiancolin you said that you're no longer seeing this, right?

davidbiancolin:
i have vcs metasimulators running with some change to how the mm_t and mmio_t pointers are handled, but i'm not confident that this problem is resolved yet.

PR #1396, opened 2023-01-20
Add VCS metasimulation to CI

abejgonzalez:
Adds parallel VCS simulations on the FPGA CI machine. This should help test that VCS works in addition to Verilator metasims to avoid issues like https://github.com/firesim/firesim/issues/901

### Contributor Checklist


abejgonzalez:
@mergifyio rebase main

.github/scripts/run-parallel-vcs-metasims.py
```diff
@@ -0,0 +1,75 @@
...
+                    rc = run("./marshal -v build test/bare.yaml &> bare.full.log").return_code
+                    if rc != 0:
+                        run("cat bare.full.log")
+                        raise Exception("Building test/bare.yaml failed to run")
+
+                run("./marshal -v install test/bare.yaml")
+
+            run("firesim managerinit --platform vitis")
+
+            def run_w_timeout(workload: str, timeout: str):
```
t14916, line 37: This is largely the same as the `run_w_timeout` in the fpga-sim CI scripts right? I have a PR in our fork right now that helps pull some of this out, but would be helpful trying to understand any differences between them.

deploy/workloads/hello-world-localhost-vcs-metasim.yaml
```diff
@@ -0,0 +1,48 @@
+run_farm:
```
t14916, line 1: I feel like there are enough of these to warrant a subdirectory in workloads or perhaps a different directory to put CI tests in, as I feel like it's getting a bit cluttered with a lot of similar config files.

.github/scripts/run-parallel-vcs-metasims.py
abejgonzalez: For the most part yes. The main difference is that the actual "run workload" line (https://github.com/firesim/firesim/blob/ae9ea901880fcf7762c69da32c2efae40eff0730/.github/scripts/run-linux-poweroff-vitis.py#L44) in the FPGA sim case is setup to point to a different `run-workload.sh` script with a different set of files (see the function args) compared to this function which mainly wraps around the existing `deploy/workloads/run-workload.sh` script.

deploy/workloads/hello-world-localhost-vcs-metasim.yaml
abejgonzalez: I agree. I wasn't sure if we wanted to use this PR to reorganize things. Happy to make a new directory and add things there (including the vitis CI configs). I was thinking it would be something like `deploy/workloads/ci/<subfolders here>`

abejgonzalez:
Happy to do re-organization here if need be @t14916

t14916:
If you're fine with doing it yeah that might be helpful (I'm not as familiar with the fpga-sim side of CI so I'm not sure which configs are used where). If you think it's better off in another PR though I'm fine with approving this

t14916 (approved):
lgtm

PR #1401, opened 2023-01-22
Fix loadmem bug in firesim_tsi.cc

jerryz123:
The default chunk_align (32b) was less than the default host-DRAM interface (64b), which caused drops on some loadmem_writes. Fix this for now by setting chunk_align to 8.

This error would manifest only when the address written had `(addr + len) % 8 == {5|6|7}`. 
For example, writing 0x6 bytes to 0x100 would result in a sequence of writes:
1. write 2 bytes to 0x104
2. write 4 bytes to 0x100

The second write would generate a 8-byte write to 0x100, clobbering the first write.

#### UI / API Impact

None

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
+++ b/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
@@ -34,6 +34,10 @@ protected:
   void load_mem_write(addr_t addr, size_t nbytes, const void *src) override;
   void load_mem_read(addr_t addr, size_t nbytes, void *dst) override;
 
+  // This must be exactly the same as hKey.dataBits
+  // TODO: Avoid hardcoding this value here
+  size_t chunk_align() override { return 8; }
+
   std::deque<firesim_loadmem_t> loadmem_write_reqs;
   std::deque<firesim_loadmem_t> loadmem_read_reqs;
   std::deque<char> loadmem_write_data;
--- a/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
+++ b/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
@@ -34,7 +34,8 @@ protected:
   void load_mem_write(addr_t addr, size_t nbytes, const void *src) override;
   void load_mem_read(addr_t addr, size_t nbytes, void *dst) override;
 
-  // This must be exactly the same as hKey.dataBits
+  // This must be exactly the same as hKey.dataBits in the LoadMem widget
+  // See discussion here: https://github.com/firesim/firesim/pull/1401
   // TODO: Avoid hardcoding this value here
   size_t chunk_align() override { return 8; }
 


abejgonzalez (commented):
Does this have an impact on performance?

jerryz123:
No. This only affects initial binary loading, and only for the tail case - the last 8 bytes in a section in a ELF.

davidbiancolin:
We've had other issues with handling the last bytes of loadmem in the past. In lieu of actually testing the infrastructure better, could we check in a few binaries that have elf sizes that trigger this problem?

jerryz123:
I think the right way to do this is to add some self-test in `firesim_tsi_t`. Checking some sequence of `write`/`read` commands can validate that the host-DRAM interface is sane before the target payload is loaded.

One of the test vectors should check this case. It is sufficient to check that:
```
write(0x8000_0000, 6, &data);
read(0x8000_0000, 6, &data);
```
behaves correctly.

davidbiancolin:
> I think the right way to do this is to add some self-test in `firesim_tsi_t`. Checking some sequence of `write`/`read` commands can validate that the host-DRAM interface is sane before the target payload is loaded.
> 
> One of the test vectors should check this case. It is sufficient to check that:
> 
> ```
> write(0x8000_0000, 6, &data);
> read(0x8000_0000, 6, &data);
> ```
> 
> behaves correctly.

That's a great idea so long as you pull in the address somewhere sane. Could you do that here?

jerryz123:
> That's a great idea so long as you pull in the address somewhere sane. Could you do that here?

It turns out this is actually not so easy to do safely.

The problem is:
 - SerialBridge, which instantiates fesvr, doesn't know anything about the underlying implementation of target memory (it might not even be FPGA-DRAM)
 - LoadMem doesn't know the target address map, LoadMem talks directly to DRAM through the host address map
 
So the only way to add a FESVR-side self-test of the loadmem interface is to add assumptions on the relationship between the FESVR-accessible memory map and the LoadMem widget. I don't think this should be done.

I think there are more pressing issues with the current setup. I believe an ELF file with sections mapped to invalid memory regions can cause a bus hang, or even worse, generate writes to host DRAM regions allocated to unrelated bridges without any indication. This ought to be fixed by moving the loadmem widget behind the AXI4 Xbar in FPGATop (instead of in front of it).

In any case, I'd like this PR to get merged soon, as its blocking a chisel/rc/scala bump. I've improved this fix to pull hKey.dataBits from the generated header, instead of hardcoding it. Fixing the loadmem setup to be more robust should be a longer-term goal.

jerryz123:
Nvm, my better fix is no longer applicable now that the driver has been split apart more. Hardcoding chunk_align to 8 is the best temporary solution, but the longer term goal IMO should be to more tightly couple the loadmem path to fesvr.

abejgonzalez (approved):
LGTM. You should also make an issue so we can track this there as well.

sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
```diff
@@ -34,6 +34,10 @@ class firesim_tsi_t final : public testchip_tsi_t {
   void load_mem_write(addr_t addr, size_t nbytes, const void *src) override;
   void load_mem_read(addr_t addr, size_t nbytes, void *dst) override;
 
+  // This must be exactly the same as hKey.dataBits
```
davidbiancolin, line 37:
```suggestion
  // This must be exactly the same as hKey.dataBits in the LoadMem widget
  // See discussion here: https://github.com/firesim/firesim/pull/1401
```

davidbiancolin (approved):
I'd like to see the small comment change so this is more trackable, otherwise ship it

PR #1404, opened 2023-01-25
Allowed the driver to control clock token delivery

nandor:
The `ClockBridge` now exposes two registers, `CREDIT` and `CREDIT_COUNT` which can be used to set and query the number of clock tokens which the bridge is allowed to deliver to the target. Tests which rely on `PeekPokeBridge` credit the system with a maximal number of tokens. `firesim_top` also infinitely credits the system alongside a dummy `PeekPokeBridge`, but in future PRs the sole control mechanism should be that through the clock bridge.

AGFI must be regenerated since the register mapping of the clock bridge changed.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/bridges/clock.cc
+++ b/sim/midas/src/main/cc/bridges/clock.cc
@@ -26,3 +26,11 @@ uint64_t clockmodule_t::hcycle() {
   uint32_t cycle_h = simif.read(mmio_addrs.hCycle_1);
   return (((uint64_t)cycle_h) << 32) | cycle_l;
 }
+
+void clockmodule_t::credit(uint32_t credits) {
+  simif.write(mmio_addrs.CREDIT, credits);
+}
+
+bool clockmodule_t::has_credits() {
+  return simif.read(mmio_addrs.HAS_CREDIT) != 0;
+}
--- a/sim/midas/src/main/cc/bridges/clock.h
+++ b/sim/midas/src/main/cc/bridges/clock.h
@@ -12,6 +12,8 @@
 class simif_t;
 
 struct CLOCKBRIDGEMODULE_struct {
+  uint64_t HAS_CREDIT;
+  uint64_t CREDIT;
   uint64_t hCycle_0;
   uint64_t hCycle_1;
   uint64_t hCycle_latch;
@@ -42,6 +44,22 @@ public:
    * Returns the current host cycle as measured by a hardware counter
    */
   uint64_t hcycle();
+
+  /**
+   * Credit the bridge to deliver <credit> more base clock cycles.
+   *
+   * Note, if a total of N credits have been granted, no intermediate positive
+   * edges in other, out-of-phase clock domains will be simulated.  In other
+   * words, the simulation will halt after the Nth positive edge in the base
+   * clock domain + a delta in which combinational logic driven by that
+   * launching edge has been resolved.
+   */
+  void credit(uint32_t credits);
+
+  /**
+   * Check whether there are any credits remaining.
+   */
+  bool has_credits();
 
 private:
   const CLOCKBRIDGEMODULE_struct mmio_addrs;
--- a/sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
+++ b/sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
@@ -147,13 +147,33 @@ class ClockTokenVector(numClocks: Int) extends Bundle with HasChannels with Cloc
 class ClockBridgeModule(params: ClockParameters)(implicit p: Parameters)
     extends BridgeModule[ClockTokenVector] {
   val clockInfo = params.clocks
+  val phaseRelationships = clockInfo map { cInfo => (cInfo.multiplier, cInfo.divisor) }
+
   lazy val module = new BridgeModuleImp(this) {
+
   val io = IO(new WidgetIO())
   val hPort = IO(new ClockTokenVector(clockInfo.size))
-  val phaseRelationships = clockInfo map { cInfo => (cInfo.multiplier, cInfo.divisor) }
-  val clockTokenGen = Module(new RationalClockTokenGenerator(phaseRelationships))
-  hPort.clocks <> clockTokenGen.io
+  val clocks = hPort.clocks
 
+  // Register which reports the number of remaining tokens to the driver.
+  val credit = RegInit(0.U(64.W))
+  attach(credit =/= 0.U, "HAS_CREDIT", ReadOnly)
+
+  // Register used to set the credit. Increments to the credit count are always accepted.
+  val creditIn = Wire(Decoupled(UInt(ctrlWidth.W)))
+  genAndAttachQueue(creditIn, "CREDIT")
+  creditIn.ready := true.B
+
+  // Generate tokens when credit is non-zero.
+  val clockTokenGen = Module(new RationalClockTokenGenerator(phaseRelationships))
+  clockTokenGen.io.ready := credit =/= 0.U && clocks.ready
+  clocks.bits := clockTokenGen.io.bits
+  clocks.valid := clockTokenGen.io.valid
+
+  // When credit is provided, increment the counter. Decrement whenever the base clock is ticked.
+  credit := credit - Mux(clocks.fire && clocks.bits(0), 1.U, 0.U) + Mux(creditIn.fire, creditIn.bits, 0.U)
+
+  // Expose a host cycle counter to the driver.
   val hCycleName = "hCycle"
   val hCycle = genWideRORegInit(0.U(64.W), hCycleName)
   hCycle := hCycle + 1.U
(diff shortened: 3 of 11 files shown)


davidbiancolin (commented):
This is a great step in the right direction. I'm  somewhat worried about the soundness of these changes, since there's a deficiency of testing around scheduled events + stepping. Hopefully i've outlined actionable next steps. if they are not clear we should chat / drop a meeting in my calendar.

sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
```diff
@@ -143,13 +143,38 @@ class ClockTokenVector(numClocks: Int) extends Bundle with HasChannels with Cloc
...
+  val credit = RegInit(0.U(32.W))
+  attach(credit, "CREDIT_COUNT", ReadOnly)
+
+  // Reset the credit count if the register is written. Otherwise, whenever a
+  // clock token is delivered, decrement the count. Stop when it reaches 0.
+  when (creditIn.fire) {
+    credit := creditIn.bits
+  }.elsewhen (clocks.fire) {
+    credit := credit - 1.U
+  }
```
fabianschuiki, line 169: The problem her right now is that from the software side you can do `credit(100); credit(100);` and you'd expect there to be 200 credits in total arriving at the hardware. In practice there's likely going to be a data race here where the first call sets the credit to 100, which starts ticking down, but before it's at 0 the second call will reset the credit counter to 100, overwriting the non-zero count in there.

I see two approaches to fix this: One would be to add the new credits to the counter instead of resetting the counter, and also lowering the ready wire if the new credit would overflow the counter. That way there's no race  condition and always the exact number of credits coming from the host will be applied.
```
creditIn.ready := false.B
val changedCredit = Wire(credit)
when (creditIn.valid) {
  val addedCredit = newCredit +& creditIn.bits
  when (addedCredit >> 32 == 0.U) {
    changedCredit := addedCredit
    creditIn.ready := true.B
  }
}
when (clocks.fire) {
  changedCredit := changedCredit - 1.U
}
credit := changedCredit
```
An alternative approach would be to just refuse to accept new credit until the old has been drained:
```
creditIn.ready := false.B
val changedCredit = Wire(credit)
when (creditIn.valid && changedCredit == 0) {
  changedCredit := creditIn.bits
  creditIn.ready = true.B
}
when (clocks.fire) {
  changedCredit := changedCredit - 1.U
}
credit := changedCredit
```

In both cases I think you'll want to have the effects of `creditIn` and `clocks` on the credit count stack up, such that credit doesn't get lost. With the `elsewhen` approach you wouldn't subtract 1 from the credit on a clock token if there's a creditIn at the same time.

nandor: I agree, incrementing the credit count makes sense. Updated the implementation.

sim/midas/src/main/cc/bridges/clock.cc
```diff
@@ -23,3 +23,11 @@ uint64_t clockmodule_t::hcycle() {
   uint32_t cycle_h = simif.read(mmio_addrs.hCycle_1);
   return (((uint64_t)cycle_h) << 32) | cycle_l;
 }
+
+void clockmodule_t::credit(uint32_t credits) {
+  simif.write(mmio_addrs.CREDIT, credits);
```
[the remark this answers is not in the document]
nandor: Updated the implementation to increment credit. That should allow writes at arbitrary points.

sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
```diff
@@ -154,6 +154,26 @@ class ClockBridgeModule(params: ClockParameters)(implicit p: Parameters)
   val clockTokenGen = Module(new RationalClockTokenGenerator(phaseRelationships))
   hPort.clocks <> clockTokenGen.io
 
+  // Register used to set the credit.
+  val creditIn = Wire(Decoupled(UInt(32.W)))
```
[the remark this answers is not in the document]
nandor: I would like to use the decoupled aspect here. It is quite convenient to be able to read `.fire` to notice when the driver wrote something. Updated the logic to increment the credit whenever that happens.

sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
```diff
@@ -154,6 +154,26 @@ class ClockBridgeModule(params: ClockParameters)(implicit p: Parameters)
...
+
+  // Register which reports the number of remaining tokens to the driver.
+  val credit = RegInit(0.U(32.W))
+  attach(credit, "CREDIT_COUNT", ReadOnly)
+
+  // Reset the credit count if the register is written. Otherwise, whenever a
+  // clock token is delivered, decrement the count. Stop when it reaches 0.
+  when (creditIn.fire) {
+    credit := creditIn.bits
+  }.otherwise {
```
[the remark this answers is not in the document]
nandor: I would like to map credits to the delivery of a piece of information (a token vector) to the target. Actual clock frequencies should be correlated with that.

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4182550374. Verify CI state before submitting PR.

sim/src/main/cc/firesim/firesim_top.cc
```diff
@@ -85,16 +85,28 @@ void firesim_top_t::simulation_init() {
...
   while (!simulation_complete() && !finished_scheduled_tasks()) {
     run_scheduled_tasks();
-    peek_poke.step(get_largest_stepsize(), false);
-    while (!peek_poke.is_done() && !simulation_complete()) {
-      for (auto &e : sim.get_registry().get_all_bridges())
+
+    if (!clock.has_credits()) {
+      clock.credit(step_size);
+    }
+    if (peek_poke) {
```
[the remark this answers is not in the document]
nandor: I rewrote this such that the clock bridge takes full control of counting cycles, while the peek-poke bridge is fed a new batch of steps whenever it runs out.

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4280629779. Verify CI state before submitting PR.

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4284787468. Verify CI state before submitting PR.

sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
```diff
@@ -147,13 +147,33 @@ class ClockTokenVector(numClocks: Int) extends Bundle with HasChannels with Cloc
...
+
+  // Generate tokens when credit is non-zero.
   val clockTokenGen = Module(new RationalClockTokenGenerator(phaseRelationships))
-  hPort.clocks <> clockTokenGen.io
+  clockTokenGen.io.ready := credit =/= 0.U && clocks.ready
+  clocks.bits := clockTokenGen.io.bits
+  clocks.valid := clockTokenGen.io.valid
 
+  // When credit is provided, increment the counter. Decrement whenever the base clock is ticked.
+  credit := credit - Mux(clocks.fire && clocks.bits(0), 1.U, 0.U) + Mux(creditIn.fire, creditIn.bits, 0.U(32.W))
```
davidbiancolin, line 174: nit: width inference is your friend here. 
```suggestion
  credit := credit - Mux(clocks.fire && clocks.bits(0), 1.U, 0.U) + Mux(creditIn.fire, creditIn.bits, 0.U)
```

sim/midas/src/main/scala/midas/widgets/ClockBridge.scala
```diff
@@ -147,13 +147,33 @@ class ClockTokenVector(numClocks: Int) extends Bundle with HasChannels with Cloc
...
   val hPort = IO(new ClockTokenVector(clockInfo.size))
-  val phaseRelationships = clockInfo map { cInfo => (cInfo.multiplier, cInfo.divisor) }
+  val clocks = hPort.clocks
+
+  // Register which reports the number of remaining tokens to the driver.
+  val credit = RegInit(0.U(64.W))
+  attach(credit =/= 0.U, "HAS_CREDIT", ReadOnly)
+
+  // Register used to set the credit. Increments to the credit count are always accepted.
+  val creditIn = Wire(Decoupled(UInt(32.W)))
```
davidbiancolin, line 163: I know this is totally inconsistent, but there's a member in Widget that is suitable for grabbing the width of the MMIO bus: https://github.com/firesim/firesim/blob/main/sim/midas/src/main/scala/midas/widgets/Widget.scala#L58

sim/midas/src/main/cc/bridges/clock.h
```diff
@@ -43,6 +45,20 @@ class clockmodule_t final : public widget_t {
    */
   uint64_t hcycle();
 
+  /**
+   * Credit the system to deliver a given number of clock ticks to the target.
```
davidbiancolin, line 49: This needs to be updated. Here's what i'm thinking. Potentially link out to this: https://docs.fires.im/en/stable/Advanced-Usage/Generating-Different-Targets.html?highlight=base%20clock#the-base-clock

```suggestion
   * Credits the bridge, allowing the simulation to advance up to <credits> more base clock cycles.
   * 
   * Note, if a total of N credits have been granted, no intermediate positive edges in other, out-of-phase clock domains will be simulated.  In other words. the simulation will halt after the Nth positive edge in the base clock domain + a delta in which combinational logic driven by that launching edge has been resolved. 
   */
```

davidbiancolin: Will need to be formatted.

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4290712433. Verify CI state before submitting PR.

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4292323143. Verify CI state before submitting PR.

PR #1369, opened 2023-01-06
Added a bridge registry to own all bridge instances

nandor:
This PR introduces the `widget_registry` class which centralizes the ownership of all bridges and widgets of a design.
To distinguish bridges, a simple RTTI mechanism is added to widgets to enable an LLVM-style `dyn_cast` method and categorization of widgets. The registry can then be queried to fetch widgets of lists of bridges of a particular kind.
`simif_t` is also changed to add method retrieving the functors for stream engines, allowing the registry to access them while constructing the stream engine widgets.

#### Related PRs / Issues

https://docs.google.com/document/d/1ilBNn-Uy6P76GUQWGiALBvJIfKyQpjzzsaOu2RGl9u8#bookmark=id.utpeyue9vfk5

### Contributor Checklist
- [x] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [x] Did you add Scaladoc/docstring/doxygen to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous prints/debugging code?
- [x] Did you state the UI / API impact?
- [x] Did you specify the Verilog / AGFI compatibility impact?

- [ ] If applicable, did you regenerate and publicly share default AGFIs?

- [ ] If applicable, did you apply the `ci:fpga-deploy` label?

- [ ] If applicable, did you apply the `Please Backport` label?

### Reviewer Checklist (only modified by reviewer)
Note: to run CI on PRs from forks, comment `@Mergifyio copy main` and manage the change from the new PR.
- [ ] Is the title suitable for inclusion in the changelog and does the PR have a `changelog:<topic>` label?
- [ ] Did you mark the proper release milestone?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?

PR #1409, opened 2023-01-27
Fix Vitis driver compilation + Fix CY-as-top driver builds

abejgonzalez:
Fixes #1390 issues listed in #1403 

- Capitalizes all `firesim_sbt_project` to be consistent
- `firesim.build` recipe now directly uses the `FIRESIM_SBT_PROJECT` instead of `$(1)`
- `AXIStreamIO` in `simif_vitis.cc` was undefined. I think its just safe to remove this overridden function and default to the exit behavior given in `simif.cc`
- Adds a `cd $(base_dir)` when calling the Chipyard generator. Needed for https://github.com/ucb-bar/chipyard/blob/95f30d0411645b12dfc5c0b14f9af0a5a678a991/generators/firechip/src/main/scala/TargetConfigs.scala#L29-L30 to properly find the right BootRom path (if this isn't done it will try to read the BootRom from the `target-rtl` location which is an uninitialized submodule instead of the `../../../` path).
- Modifies the `TARGET_SBT_PROJECT` since this SBT invocation is done in the `firesim/sim` directory instead of `chipyard` main directory.

### Contributor Checklist

### DIFF
--- a/sim/make/goldengate.mk
+++ b/sim/make/goldengate.mk
@@ -50,7 +50,7 @@ $(simulator_verilog).intermediate: $(FIRRTL_FILE) $(ANNO_FILE) $(FIRESIM_MAIN_CP
 .PHONY: conf
 conf: $(fame_annos)
 	mkdir -p $(GENERATED_DIR)
-	cd $(base_dir) && $(SBT) "project $(firesim_sbt_project)" "runMain midas.stage.RuntimeConfigGeneratorMain \
+	cd $(base_dir) && $(SBT) "project $(FIRESIM_SBT_PROJECT)" "runMain midas.stage.RuntimeConfigGeneratorMain \
 		-td $(GENERATED_DIR) \
 		-faf $(fame_annos) \
 		-ggcp $(PLATFORM_CONFIG_PACKAGE) \
--- a/sim/src/main/makefrag/firesim/build.mk
+++ b/sim/src/main/makefrag/firesim/build.mk
@@ -2,7 +2,7 @@
 
 $(FIRRTL_FILE) $(ANNO_FILE): $(TARGET_CP)
 	@mkdir -p $(@D)
-	java -cp $$(cat $(TARGET_CP)) chipyard.Generator \
+	cd $(base_dir) && java -cp $$(cat $(TARGET_CP)) chipyard.Generator \
 		--target-dir $(GENERATED_DIR) \
 		--name $(long_name) \
 		--top-module $(DESIGN_PACKAGE).$(DESIGN) \
--- a/sim/src/main/makefrag/firesim/config.mk
+++ b/sim/src/main/makefrag/firesim/config.mk
@@ -19,11 +19,7 @@ PLATFORM_CONFIG ?= BaseF1Config
 TARGET_SRC_DIRS ?= $(chipyard_dir)/generators/
 
 # Project for the target.
-ifdef FIRESIM_STANDALONE
-	TARGET_SBT_PROJECT := {file:${chipyard_dir}}firechip
-else
-	TARGET_SBT_PROJECT := firechip
-endif
+TARGET_SBT_PROJECT := {file:${chipyard_dir}}firechip
 
 # Directory where sources are located.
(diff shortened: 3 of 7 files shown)


sim/src/main/makefrag/firesim/config.mk
```diff
@@ -22,7 +22,7 @@ TARGET_SRC_DIRS ?= $(chipyard_dir)/generators/
 ifdef FIRESIM_STANDALONE
 	TARGET_SBT_PROJECT := {file:${chipyard_dir}}firechip
 else
-	TARGET_SBT_PROJECT := firechip
+	TARGET_SBT_PROJECT := {file:${chipyard_dir}}firechip
```
davidbiancolin, line 25: Is this not the same as above?

abejgonzalez: Yup. Updated it.

sim/make/goldengate.mk
```diff
@@ -50,7 +50,7 @@ $(simulator_verilog).intermediate: $(FIRRTL_FILE) $(ANNO_FILE) $(FIRESIM_MAIN_CP
 .PHONY: conf
 conf: $(fame_annos)
 	mkdir -p $(GENERATED_DIR)
-	cd $(base_dir) && $(SBT) "project $(firesim_sbt_project)" "runMain midas.stage.RuntimeConfigGeneratorMain \
+	cd $(base_dir) && $(SBT) "project $(FIRESIM_SBT_PROJECT)" "runMain midas.stage.RuntimeConfigGeneratorMain \
```
nandor, line 53: Thanks for taking care of this!

sim/src/main/makefrag/firesim/build.mk
```diff
@@ -2,7 +2,7 @@
 
 $(FIRRTL_FILE) $(ANNO_FILE): $(TARGET_CP)
 	@mkdir -p $(@D)
-	java -cp $$(cat $(TARGET_CP)) chipyard.Generator \
+	cd $(base_dir) && java -cp $$(cat $(TARGET_CP)) chipyard.Generator \
```
nandor, line 5: What exactly breaks if the working directory is not changed?

nandor (approved):
LGTM, thanks for spotting this. I pushed a change to fix the drivers.

My burning question is: why does some verilator metasimulation job on CI detect a driver change and why didn't the linux power-on tests fail since the f1/vitis drivers seem to have been broken for quite a while?

sim/src/main/makefrag/firesim/build.mk
abejgonzalez: https://github.com/ucb-bar/chipyard/blob/95f30d0411645b12dfc5c0b14f9af0a5a678a991/generators/firechip/src/main/scala/TargetConfigs.scala#L29-L30 selects the BootROM based off the working directory. In the FireSim-as-Top case, the working directory is `firesim/sim` and the `target-rtl/chipyard` directory is checked out (submodule is init'ed) allowing the code to find the right bootrom. However in the Chipyard-as-Top case (without the fix) things are a bit different. The working directory is still `firesim/sim`, so the code linked 1st tries to find the BootROM in `<working-dir>/generators/...` (a CY path) and fails to find it there. Next it looks in `<working-dir>/target-rtl/chipyard/...`. However, this also fails since the `target-rtl/chipyard` submodule is uninitialized. Changing the working directory to Chipyard case avoids this.

abejgonzalez:
> LGTM, thanks for spotting this. I pushed a change to fix the drivers.

Thanks!

> My burning question is: why does some verilator metasimulation job on CI detect a driver crash and why didn't the linux power-on tests fail since the f1/vitis drivers seem to have been broken for quite a while?

I looked at #1390 and saw that metasims passed (hence the driver passed) so I'm not sure. It looks like #1369 didn't run Vitis testing at all so this wouldn't been missed. I'll probably add some CI pass to build the Vitis driver even if FPGA simulation is not done.

abejgonzalez:
@mergifyio rebase main

abejgonzalez:
Re-running CI since Linux poweroff was cancelled for some reason.

PR #1419, opened 2023-01-31
Removed the compiler-generated runtime config

nandor:
Since the default values are already used to initialize the registers on the hardware side, the configuration replicating them was eliminated.

To help filter out errors, if no arguments are specified to the bridge, this default configuration is used and no warnings are emitted.
Otherwise, if at least one register is specified, all other registers are expected to be specified. However, if the `+useHardwareDefaultRuntimeSettings`, the bridge emits warnings instead of errors and resorts to the default values.

#### UI / API Impact

The `make` flow no longer accepts runtime config files.

### Contributor Checklist


sim/midas/src/main/cc/bridges/fased_memory_timing_model.cc
```diff
@@ -63,24 +63,32 @@ FASEDMemoryTimingModel::FASEDMemoryTimingModel(
...
+      continue;
+    }
+
+    auto sub_arg = arg.substr(4);
+    size_t delimit_idx = sub_arg.find("=");
+    size_t suffix_idx = sub_arg.find(suffix);
+    std::string key = sub_arg.substr(0, suffix_idx).c_str();
+
+    // This is the only nullary plusarg supported by fased
+    if (key == std::string("useHardwareDefaultRuntimeSettings")) {
```
davidbiancolin, line 77: So if using the HW defaults is going to be the default, shouldn't this be renamed to "imposeCompleteConfiguration" or something similar, and the member name updated?

davidbiancolin (commented):
I'm fine with removing the emission of default +args under GG, but please don't remove the RuntimeConfigGenerator phase/stage in this PR. That requires a little more thought.

sim/midas/src/main/cc/bridges/fased_memory_timing_model.cc
nandor: It would be useful to clean this up, but we should have a discussion first, as I feel this ties into the stuff we want to do with argument parsing.

PR #1426, opened 2023-02-04
Test for existing TracerV bridge including trigger modes

sifive-benjamin-morse:
Adds a unit test for TracerV, including the various `+trace-select` modes.

#### Related PRs / Issues
During the development, I noticed this issue: https://github.com/firesim/firesim/issues/1428  I've this omitted testing for `+trace-select=3` until the issue becomes resolved. The test code DOES exist here, but will fail due to the outstanding issue

#### UI / API Impact
There is no impact, this is just more testing.

no

### Contributor Checklist


sim/src/test/scala/bridges/TracerVSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+    expected.deleteOnExit()
+
+    // Create the output file. tracerv will always append '-C0' to the end of the path provided in the plusarg
+    val output     = File.createTempFile("output", ".txt-C0")
+    output.deleteOnExit()
+    val outputPath = output.getPath.substring(0, output.getPath.length() - 3)
+
+    // group the optional function args together with the correct plusarg string names
+    val optionalArgs = Seq(
+      Seq("+trace-select=", trace),
```
davidbiancolin, line 34:
```suggestion
      ("+trace-select=", trace),
```

sim/src/test/scala/bridges/TracerVSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+      Seq("+trace-start=", start),
```
davidbiancolin, line 35:
```suggestion
      ("+trace-start=", start),
```

sim/src/test/scala/bridges/TracerVSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+    // a seq starting with fixed plusargs, ending with optional plusargs
+    // the optional plusargs are properly constructed or dropped
+    val args = Seq(
+      s"+tracefile=${outputPath}",
+      s"+tracerv-expected-output=${expected.getPath}",
+    ) ++ optionalArgs.collect { case Seq(a, Some(b)) =>
```
davidbiancolin, line 43:
```suggestion
    ) ++ optionalArgs.collect { case (a, Some(b)) =>
```

sim/src/test/scala/bridges/TracerVSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+    assert(runResult == 0)
+
+    val expectedContents = scala.io.Source.fromFile(expected.getPath).mkString
+    val outputContents   = scala.io.Source.fromFile(output.getPath).mkString
+    outputContents should equal(expectedContents)
+  }
+}
+
+class TracerVF1TestCount1 extends TracerVTestBase(BaseConfigs.F1, 1);
+// class TracerVF1TestCount5 extends TracerVTestBase(BaseConfigs.F1, 5, Some(3), Some("FF0000001C")); // in hex
```
davidbiancolin, line 58: Explain why this is commented here / refer to an issue.

sim/src/test/scala/bridges/TracerVSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+) extends BridgeSuite("TracerVModule", s"TracerVModuleTestCount${width}", platformConfig) {
+  override def runTest(backend: String, debug: Boolean) {
+    // Create an expected file.
+    val expected = File.createTempFile("expected", ".txt")
+    expected.deleteOnExit()
+
+    // Create the output file. tracerv will always append '-C0' to the end of the path provided in the plusarg
+    val output     = File.createTempFile("output", ".txt-C0")
+    output.deleteOnExit()
+    val outputPath = output.getPath.substring(0, output.getPath.length() - 3)
```
davidbiancolin, line 30: Hoist the suffix out into it's own `val` and use [stripSuffix](https://www.scala-lang.org/api/2.12.7/scala/collection/immutable/StringLike.html#stripSuffix(suffix:String):String)

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+    // load into MMIO, and tick the system
+    for (unsigned test_step = 0; test_step < get_total_trace_tests();
+         test_step++) {
+      const uint64_t pull = rand_next(tracerv_width + 1);
+
+      auto pull_iaddr = get_iaddrs(test_step, tracerv_width);
+      auto pull_bits = get_contiguous(pull, tracerv_width);
+
+      load(pull_iaddr, pull_bits);
+      steps(1);
```
davidbiancolin, line 212: If i'm understanding this correctly, the DUT is going to stall every single because you're poking new values right. How hard would it be to support runs of time where the bridge is being driven at full throughput? 

Maybe you could capture the values you're enqueuing here in a really larger FIFO in the Chisel DUT, and then let that drain into tracerV (perhaps with some fuzzing of valid) once you say "go". 

Still this is good start.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+                 std::back_inserter(filtered),
+                 fn);
+    expected_pair = std::move(filtered);
+  }
+
+  /**
+   * Depending on the mode (0,1,2,3) remove expected traces
+   * to match the behavior of the trigger in scala
+   */
+  void filter_expected_test() {
```
davidbiancolin, line 250: I'd like to see the end trigger too, but that can be a follow up PR.

davidbiancolin (commented):
@nandor can you review / review the C++ side of this.

sim/src/main/cc/bridges/TracerVModule.cc
sifive-benjamin-morse: I think this would be easy to do with some support from you. The question here, is this in the scope of this PR? Should we add this to the scope of the wide-tracerv? PR? lets talk

sim/make/cpp-lint.mk
```diff
@@ -24,6 +24,7 @@ clang_tidy_files := $(shell \
 clang_tidy_flags :=\
 	-I$(firesim_base_dir)/midas/src/main/cc \
 	-I$(firesim_base_dir)/firesim-lib/src/main/cc \
+	-I$(firesim_base_dir)/src/main/cc/midasexamples \
```
nandor, line 27: To settle some more of the style issues, I would suggest enabling `clang-tidy` on the header. [Here](https://github.com/firesim/firesim/blob/71b8eb9c137deccb104b8e5044221eca36203a7a/sim/make/cpp-lint.mk#L15), we could filter out the `.cpp` and nested stuff while still validating the header. The problem is with the fact that `clang-tidy` needs the full set of dependencies which are slow to install. If we upgraded libelf/libdwarf, we could install them through the system packages and we'd be able to run clang-tidy.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+    ret.emplace_back(value);
+  }
+
+  return ret;
+}
+
+static std::vector<uint64_t> get_iaddrs(const unsigned step,
+                                        const unsigned total) {
+  constexpr uint64_t offset =
+      1024; // should be larger than total but doesn't really matter
```
nandor, line 27: Please put the comment on the line before so we don't get this weird split.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+#include "bridges/tracerv.h"
+#include "core/bridge_driver.h"
+#include "core/simif.h"
+#include "core/simulation.h"
+
+#include <algorithm>
+#include <iostream>
+#include <string_view>
+
+static std::vector<bool> get_contiguous(const unsigned bits,
```
nandor, line 13: While `clang-tidy` doesn't complain about primitive const in function definitions, please don't const-qualify all primitive arguments. It is quite noisy.

sim/firesim-lib/src/main/cc/bridges/tracerv.cc
```diff
@@ -23,15 +23,13 @@ char tracerv_t::KIND;
...
-constexpr uint64_t valid_mask = (1ULL << 40);
-
 tracerv_t::tracerv_t(simif_t &sim,
                      StreamEngine &stream,
                      const TRACERVBRIDGEMODULE_struct &mmio_addrs,
                      int tracerno,
                      const std::vector<std::string> &args,
-                     int stream_idx,
-                     int stream_depth,
+                     const int stream_idx,
```
nandor, line 31: No const here.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+
+static std::string namev(const unsigned x) {
+  std::stringstream ss;
+  ss << "io_insns_" << x << "_valid";
+  return ss.str();
+};
+
+// a simple way to turn the iaddr values we use
+// in this test into a 5 bit value
+static uint32_t fold(const uint64_t x) {
```
nandor, line 57: Could you rename the argument/locals so that the `iaddr` in the comment is somewhat tied to the source?

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+  std::vector<uint64_t> ret;
+
+  for (unsigned i = 0; i < total; i++) {
+    ret.emplace_back(step * offset + i);
+  }
+
+  return ret;
+}
+
+static std::string namei(const unsigned x) {
```
nandor, line 37: Could we replace `x` with `i` or `index`?

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+            std::vector<uint64_t>(valid_i.begin() + i, valid_i.begin() + last);
+        expected_pair.emplace_back(std::make_pair(e_cycle, chunk));
+      }
+
+      e_cycle++;
+    };
+
+    // loop over tests. choose random valids with a simple pattern of iaddr
+    // load into MMIO, and tick the system
+    for (unsigned test_step = 0; test_step < get_total_trace_tests();
```
nandor, line 204: In order to avoid repeated invocations to the bound function, in LLVM it is quite common to capture it in a variable: `for (unsigned step = 0, bound = get_total_trace_tests(); step < bound; ++step)`.

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+    uint64_t e_cycle = cycle_offset;
+
+    const unsigned tracerv_width = tracerv.max_core_ipc;
+
+    // load MMIO and capture expected outputs
+    auto load = [&](std::vector<uint64_t> iad, std::vector<bool> bit) {
+      std::vector<uint64_t> valid_i;
+      assert(iad.size() == bit.size());
+      for (unsigned i = 0; i < iad.size(); i++) {
+        // std::cout << "loading " << i << " with " << iad[i] << "," << bit[i]
```
nandor, line 179: remove comment

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+
+    tracerv.flush();
+
+    steps(100);
+
+    // if testing a trigger mode, the output need to be filtered
+    filter_expected_test();
+
+    // write out a file which contains the expected output
+    write_expected_file();
```
nandor, line 228: If my understanding is correct, I would rather create a copy of the filtered list instead of modifying it in-place. Should help with readability a lot: ```write_expected_file(filter_expected_test())```

sim/src/main/cc/bridges/TracerVModule.cc
```diff
@@ -0,0 +1,367 @@
...
+    for (auto &arg : args) {
+      if (arg.find("+seed=") == 0) {
+        random_seed = strtoll(arg.c_str() + 6, nullptr, 10);
+        fprintf(stderr, "Using custom SEED: %ld\n", random_seed);
+      }
+
+      if (arg.find("+tracerv-expected-output=") == 0) {
+        const std::string fname =
+            arg.c_str() + 25; // 25 is the length of the argument to find
+        expected = fopen(fname.c_str(), "w");
```
nandor, line 86: The scope of open files should be narrowed. The constructor can parse the file name, which should only be opened in the write function below.

nandor (changes requested):
Thanks for building this test - I have left some comments on the C++ bits that should make the tests a bit easier to read for those unfamiliar with TracerV.

nandor (approved):
LGTM

PR #1431, opened 2023-02-11
Do not materialize memory connections if the target does not use them

nandor:
In line with the omission of managed stream engines and their associated connections to the top-level, this PR also eliminates unused memory connections. The top-level ports to the platform hardware are left unchanged. If no memories are used in the target, the `LoadMem` widget is omitted.

#### Related PRs / Issues

#1430 

The port mappings are not affected.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/cc/core/simulation.cc
+++ b/sim/midas/src/main/cc/core/simulation.cc
@@ -125,13 +125,20 @@ void simulation_t::wait_for_init() {
 }
 
 void simulation_t::init_dram() {
-  auto &loadmem = registry.get_widget<loadmem_t>();
-  if (do_zero_out_dram) {
-    fprintf(stderr, "Zeroing out FPGA DRAM. This will take a few seconds...\n");
-    loadmem.zero_out_dram();
-  }
+  if (auto *loadmem = registry.get_widget_opt<loadmem_t>()) {
+    if (do_zero_out_dram) {
+      fprintf(stderr,
+              "Zeroing out FPGA DRAM. This will take a few seconds...\n");
+      loadmem->zero_out_dram();
+    }
 
-  if (!load_mem_path.empty()) {
-    loadmem.load_mem_from_file(load_mem_path);
+    if (!load_mem_path.empty()) {
+      loadmem->load_mem_from_file(load_mem_path);
+    }
+  } else {
+    if (do_zero_out_dram || !load_mem_path.empty()) {
+      fprintf(stderr,
+              "Skipping memory initialization: target does not use DRAM\n");
+    }
   }
 }
--- a/sim/midas/src/main/cc/core/widget_registry.h
+++ b/sim/midas/src/main/cc/core/widget_registry.h
@@ -56,10 +56,23 @@ public:
    */
   template <typename T>
   T &get_widget() {
+    auto *widget = get_widget_opt<T>();
+    assert(widget && "cannot find widget");
+    return *widget;
+  }
+
+  /**
+   * Return a widget of a particular kind, if it has a single instance.
+   *
+   * @tparam T Type of the widget to fetch
+   * @return Pointer to the widget instance.
+   */
+  template <typename T>
+  T *get_widget_opt() {
     auto it = widgets.find(&T::KIND);
-    assert(it != widgets.end() && "no bridges registered");
-    assert(it->second.size() == 1 && "multiple bridges registered");
-    return *static_cast<T *>(it->second[0].get());
+    if (it == widgets.end() || it->second.size() != 1)
+      return nullptr;
+    return static_cast<T *>(it->second[0].get());
   }
 
   /**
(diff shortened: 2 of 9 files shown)


sim/midas/src/main/cc/core/simulation.cc
```diff
@@ -125,13 +125,15 @@ void simulation_t::wait_for_init() {
...
 void simulation_t::init_dram() {
-  auto &loadmem = registry.get_widget<loadmem_t>();
-  if (do_zero_out_dram) {
-    fprintf(stderr, "Zeroing out FPGA DRAM. This will take a few seconds...\n");
-    loadmem.zero_out_dram();
-  }
+  if (auto *loadmem = registry.get_widget_opt<loadmem_t>()) {
+    if (do_zero_out_dram) {
+      fprintf(stderr,
+              "Zeroing out FPGA DRAM. This will take a few seconds...\n");
```
davidbiancolin, line 131: It might be wise to print something to the effect of: "Skipping DRAM zero-initialization since FPGA DRAM is unused" or something to that effect in an else.

sim/midas/src/main/cc/core/widget_registry.h
```diff
@@ -56,10 +56,23 @@ class widget_registry_t final {
...
+  }
+
+  /**
+   * Return a widget of a particular kind, if it has a single instance.
+   *
+   * @tparam T Type of the widget to fetch
+   * @return Pointer to the widget instance.
+   */
+  template <typename T>
+  T *get_widget_opt() {
```
davidbiancolin, line 71: Why use a pointer instead of `std::optional`? (for my own elucidation)

sim/midas/src/main/scala/midas/platform/xilinx/AXI4Crossing.scala
```diff
@@ -170,6 +170,39 @@ class XilinxAXI4Bundle(val bundleParams: AXI4BundleParameters, val isAXI4Lite: B
       }
     }
   }
+
+  def tieoff(axiClock: Clock, axiReset: Bool): Unit = {
```
davidbiancolin, line 174:
```suggestion
  def tieoff(): Unit = {
```

So i know there are examples of this elsewhere, but this should be renamed to somthing like: `tieoffAsManager` or or something else to indicate this bidirectional is interface is being disabled from the manager-side if that makes sense.

sim/midas/src/main/scala/midas/platform/xilinx/AXI4Crossing.scala
```diff
@@ -170,6 +170,39 @@ class XilinxAXI4Bundle(val bundleParams: AXI4BundleParameters, val isAXI4Lite: B
...
+    awid.foreach { _ := DontCare }
+    awaddr  := DontCare
+    awlen.foreach { _ := DontCare }
```
davidbiancolin, line 177: Have your forays into Vivado synthesis issues warrant that we start tying things off to known values instead of `DontCare`, instead of things that may be implementation defined? I'm fine with this as is.

sim/midas/src/main/scala/midas/platform/VitisShim.scala
```diff
@@ -78,18 +78,24 @@ class VitisShim(implicit p: Parameters) extends PlatformShim {
...
+      host_mem_cdc.io.s_axi.araddr  := 0x4000000000L.U(addrWidth) + top.module.mem(0).ar.bits.addr
+      host_mem_cdc.io.s_axi.awaddr  := 0x4000000000L.U(addrWidth) + top.module.mem(0).aw.bits.addr
+      host_mem_cdc.io.s_axi.arcache.foreach { _ := AXI4Parameters.CACHE_MODIFIABLE }
+      host_mem_cdc.io.s_axi.awcache.foreach { _ := AXI4Parameters.CACHE_MODIFIABLE }
+
+      host_mem_0                    <> host_mem_cdc.io.m_axi
+      host_mem_cdc.io.m_axi_aclk    := ap_clk
+      host_mem_cdc.io.m_axi_aresetn := ap_rst_n
+    } else {
+      host_mem_0.tieoff(hostClock, hostSyncReset)
```
davidbiancolin, line 97:
```suggestion
      host_mem_0.tieoff()
```
Once you remove the parameters. You should also rename this.

davidbiancolin (commented):
Just minor changes. All-in all this is a nice change.

sim/midas/src/main/cc/core/widget_registry.h
nandor: We want a non-owning reference here and a pointer is the way to do it (I don't want a shared_pointer since the registry is the owner of the objects). `std::optional` should be used with value types.

sim/midas/src/main/scala/midas/platform/xilinx/AXI4Crossing.scala
nandor: Good to know `0.U` works, I prefer that.

fabianschuiki (approved):

sim/midas/src/main/scala/midas/platform/xilinx/AXI4Crossing.scala
davidbiancolin: So long as valid is tied off you should be able to set the payload values to whatever you want, and then yeah, 0.U will have its width inferred accordingly.

PR #1432, opened 2023-02-15
Add URI support to tarball path and xclbin path

sifive-benjamin-morse:
* This allows `config_hwdb.yaml` to use URI's for the `driver_tar` and `xclbin` keys.
* This also fixes an issue in PR #1299 regarding `server_hardware_config` being a string
* CI runs which previously used wget to download an `.xclbin` (such as `run-linux-poweroff-vitis.py`) can now directly specify the URL as a URI

#### UI / API Impact
Allows much easier collaboration with new developers as pre-compiled resources can be put on the web simplifying first time runs. 

Should be no change.

### Contributor Checklist


deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -172,12 +239,21 @@ def copy_sim_slot_infrastructure(self, slotno: int) -> None:
...
+                # downloads will survive until after the rsync below,
+                # at which point the local copies are not needed and removed.
+                files_to_copy.extend(self.local_download_all_uri(slotno, uridir))
+
+                for local_path, remote_path in files_to_copy:
+                    # -z --inplace
+                    rsync_cap = rsync_project(local_dir=local_path, remote_dir=pjoin(remote_sim_rsync_dir, remote_path),
+                                ssh_opts="-o StrictHostKeyChecking=no", extra_opts="-L", capture=True)
+                    rootLogger.debug(rsync_cap)
+                    rootLogger.debug(rsync_cap.stderr)
```
timsnyder-siv, line 256: I was hoping that when we added support for URIs that we would not download everything to the manager than then copy it to the runtime hosts, we would do the downloading in a distributed fashion on the runtime hosts.  When running large simulations of dozens of nodes, we're already constrained by the bandwidth available on the single manager host.

It also seems like this will change repeated `infrasetup` runs from being a checksum check done by the rsync to re-downloading the local copy and then doing the checksum to see that it doesn't need to be copied remotely.

If downloading locally makes it easier to add the URI support, we can distribute the downloading later to address the bottleneck.

deploy/runtools/runtime_config.py
```diff
@@ -97,6 +106,20 @@ def __init__(self, name: str, hwconfig_dict: Dict[str, Any]) -> None:
...
+        """ Private method to init / validate any URI. """
+
+        uri = hwconfig_dict.get(prop)
+
+        if uri is None:
+            return None
+
+        is_uri = re.match(_RFC_3986_PATTERN, uri)
+        if not is_uri:
+            raise Exception(f"when {prop} is set, it must be a URI. (found '{uri}')")
```
timsnyder-siv, line 120: In https://github.com/firesim/firesim/pull/1140/files#r927042552 @davidbiancolin wanted to support things optionally being a filepath because URIs don't have a means for specifying a relative path.  His thinking may have changed on that though.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -94,6 +142,12 @@ def terminate_instance(self) -> None:
         """
         raise NotImplementedError
 
+    @classmethod
+    @abc.abstractmethod
+    def sim_command_requires_sudo(self) -> bool:
```
timsnyder-siv, line 147: nit: if this needs to be a `@classmethod` it's customary to have the param be called `class` instead of `self` because that's what you'll be getting.

deploy/runtools/firesim_topology_elements.py
```diff
@@ -546,10 +554,16 @@ def get_rootfs_name(self) -> Optional[str]:
...
+    @classmethod
+    def get_tar_name(cls) -> str:
         """ Get the name of the tarball on the run host"""
         return "driver-bundle.tar.gz"
 
+    @classmethod
+    def get_xclbin_name(cls) -> str:
+        """ Get the name of the xclbin on the run host"""
+        return "bitstream.xclbin"
+
```
timsnyder-siv, line 566: Seems strange to me that these methods returning names relative to the `sim_slot_X` directory of a particular sim  are `@classmethod` of the `FireSimServerNode` class.

sifive-benjamin-morse: Ok, I wasn't exactly sure where to put these

conda-reqs.yaml
```diff
@@ -23,11 +23,28 @@ dependencies:
...
+    - libprotobuf==3.21.11
+    - krb5==1.19.3
+    - libllvm15==15.0.6
+    - libtiff==4.4.0
+    - openssl==3.0.7
+    - pkginfo==1.9.2
+    - harfbuzz==5.3.0
+    - aws-sam-translator==1.55.0
+    - fontconfig==2.14.1
+
```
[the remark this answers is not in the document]
sifive-benjamin-morse: conda-lock doesn't support updating one or two dependencies, without updating all others to the latest version. The solution of locking everything down came about after a discussion with @davidbiancolin : He said that the prebuilt drivers will break in future revisions if we have CPP dependencies constantly updating.

The reason I need to touch this at all, is so that I can add: 
    - fsspec, s3fs, aiohttp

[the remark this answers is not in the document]
timsnyder-siv: the CPP dependencies are going to be bundled with the driver at build time in the tarball, aren't they?

deploy/runtools/run_farm_deploy_managers.py
timsnyder-siv: Benjamin pointed out my flawed understanding of `fabric.api.execute()`.  Distributing the downloads is out of scope for this PR.

conda-reqs.yaml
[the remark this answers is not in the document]
davidbiancolin: I was worried about potential discrepancies between the two ways we build the driver. That may be ill-founded. We'll need to take this out-of-band.

.github/scripts/run-linux-poweroff-vitis.py
```diff
@@ -30,10 +30,7 @@ def run_linux_poweroff_vitis():
 
                 run("./marshal -v install test/outputs.yaml")
 
-            # download prebuilt xclbin to /tmp
-            # this CI image is managed by the default CI account
-            with prefix('cd /tmp'):
-                run('wget https://firesim-ci-vitis-xclbins.s3.us-west-2.amazonaws.com/firesim_rocket_singlecore_no_nic_d148b73.xclbin')
+            # No download, edit config_hwdb.yaml directly and specify URI there
```
davidbiancolin, line 33:
```suggestion
```

Nit: It's preferable to make this as a comment in the PR (to help with review) because the comment (without any accompanying code) will lose context over time.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -687,3 +779,16 @@ def infrasetup_instance(self) -> None:
     def terminate_instance(self) -> None:
         """ VitisInstanceDeployManager machines cannot be terminated. """
         return
+
+class AzureInstanceDeployManager(VitisInstanceDeployManager):
```
davidbiancolin, line 783: IIRC, I think you plan to remove the azure support before the final merge right?

deploy/runtools/runtime_config.py
davidbiancolin: I spoke to @sifive-benjamin-morse about this out-of-band (after you made this comment) and i really don't have a strong opinion anymore (so i'm deferring to Ben). I guess there's a strong possibility i'll want the feature when i start using the mangaer.

Whether non-URI, relative paths should be supported is probably a question best posed to the `firesim-dev` chat.

conda-reqs.yaml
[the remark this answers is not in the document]
timsnyder-siv: conda-lock does support 'updating' something that is in the existing lockfile to a new version in a minimal fashion with it's `--update` flag.  However, it seems like you can't add new packages to the requirements spec and also use `--update` for them to minimally install new packages.  I am trying to get a discussion of this usecase going in https://github.com/conda/conda-lock/pull/355#discussion_r1111114862

I think I see how I could patch conda-lock to support using `--update` with new packages.  I'm not sure how long it would take me to work through actually making that enhancement given that I just started looking at the project's code.  However, I'm hoping that this is just a documentation issue with conda-lock and that it's already possible to do this or trivial for the original dev or maintainers.

deploy/runtools/run_farm_deploy_managers.py
sifive-benjamin-morse: These are necessary but not sufficient changes to add Azure support. I don't have access to running an `.xclbin` flow anywhere except for Azure. So I was thinking of leaving these in for now.

sifive-benjamin-morse:
I think that moving URI downloads out to a shared location is a bit complicated. Right now, with N run hosts, N URI downloads will occur. To fix this I would need to:
* edit `infrasetup_passes()` such that a URI download occurs outside of the loop.
* each node node can pick from shared or unique URI

My thoughts are:
* if the nodes are all the same type this is easy
* But there is no guarantee that each node is the same type making it complicated to share URIs.

Is there a simple solution here?


https://github.com/firesim/firesim/blob/6e993f5089a2290fba44ded39aadd972754c1083/deploy/runtools/firesim_topology_with_passes.py#L420-L436

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -646,25 +812,35 @@ def terminate_instance(self) -> None:
...
             with settings(warn_only=True), hide('everything'):
-                temp_file = "/tmp/xbutil-examine-out.json"
-                collect = run(f"xbutil examine --format JSON -o {temp_file}")
-                with open(temp_file, "r") as f:
-                    json_dict = json.loads(f.read())
+                # Examine forcibly puts the JSON in an output file (on the remote); The stdout
+                # it produces is difficult to parse so use process substitution
+                # to pipe JSON to stdout instead.
+                xbutil_examine_json = run("xbutil examine --force --format JSON -o >(cat) > /dev/null")
+                json_dict = json.loads(xbutil_examine_json)
```
timsnyder-siv, line 837: I can see how they were writing to a hardcoded tempfile on a possibly remote host and then opening it without copying it back.  In this new way of doing things, anything `xbutil` emits into `STDERR` could possibly cause difficult to understand and diagnose JSON parsing errors.

It better to pass `combine_stderr=False` (see [fabric docs](https://docs.fabfile.org/en/1.14/usage/interactivity.html#combine-streams))  to this invocation of `run()` and DEBUG log `xbutil_examine_json.stderr` if it isn't empty.

I don't remeember what uncaught `JSONDecodeError` stack traces look like but I vaguely remember them being unhelpful when the JSON being parsed is dynamically generated because they tell you where the error happened but don't print out what was parsed.  You might want to DEBUG log `xbutil_examine_json` or catch `JSONDecodeError` to raise our own meaningful `RuntimeError` message that includes the `JSONDecodeError.doc` in it's message

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+        local_dest_path=test_dest_file_path
+    )
+
+    logger_mock.debug.assert_has_calls([
+        call(f"Downloading '{file_uri}' to '{test_dest_file_path}'"),
+        call(f"Overwriting {test_dest_file_path.resolve()}"),
+        call(f"Downloading '{file_uri}' to '{test_dest_file_path}'")
+    ])
+
+    os.remove(str(test_dest_file_path))
```
timsnyder-siv, line 69: I think you got this from the person who helped me write this test or me but it would be better to:
```suggestion
    test_dest_file_path.unlink()
```

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+    if test_dest_file_path.exists():
+        os.remove(str(test_dest_file_path))
+
+    downloadURI(
+        uri=file_uri,
+        local_dest_path=test_dest_file_path
+    )
+
+    assert os.path.exists(test_dest_file_path), f"{test_dest_file_path} was not created."
```
timsnyder-siv, line 54:
```suggestion
    assert test_dest_file_path.exists(), f"{test_dest_file_path} was not created."
```

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+            mock_s3_client.upload_file(str(test_file_path), test_bucket, test_bucket_key)
+            file_uri = f"s3://{test_bucket}/{test_bucket_key}"
+        except ClientError as e:
+            pytest.fail("Failed to mock an S3 client and upload a file.")
+
+    if protocol_type == 'file':
+        file_uri = f"file://{test_file_path}"
+
+    if test_dest_file_path.exists():
+        os.remove(str(test_dest_file_path))
```
timsnyder-siv, line 47:
```suggestion
    test_dest_file_path.unlink(missing_ok=True)
```

deploy/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
+from pathlib import Path
+from unittest.mock import MagicMock, call
+from botocore.exceptions import ClientError
+import pytest
+import os
```
timsnyder-siv, line 5:
```suggestion
```

I think the only uses of `os` were misguided uses of stuff that are better done with the `Path` methods.

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -522,11 +522,31 @@ key/value in the same recipe).
...
+* `gcfs for Google Cloud Services <https://gcsfs.readthedocs.io/en/latest/#credentials>`_
+* `s3fs for AWS S3 <https://s3fs.readthedocs.io/en/latest/#credentials>`_
+
+For SSH, add any required keys to your ssh-agent.
+
+Please note that while some protocol backendss provide authentication via their own configuration
+files or environment variables (e.g. AWS credentials stored in ``~/.aws``, created by ``aws
+configure``), one can additionally configure ``fsspec`` with additional default keyword arguments
+per backend protocol by using one of the `fsspec configuration
+<https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration>`_ methods.
```
timsnyder-siv, line 549: Now that we have multiple things that can be a URI, it might be better to move this to a section at the top of this page or it's own page and link to it from anything that can use a URI.

scripts/machine-launch-script.sh
```diff
@@ -231,6 +231,8 @@ set -o pipefail
         azure-identity==1.5.0 \
         azure-mgmt-compute \
         azure-mgmt-network \
+        fsspec \
+        s3fs \
```
timsnyder-siv, line 235: The requirements in this file a minimal and only serve to give the user enough of an environment to be able to clone the firesim repo and run `build_setup.sh`.  I see that other stuff has snuck in here since I last looked but are `fsspec` and `s3fs` needed before we have the manager?

timsnyder-siv: I think they may have been here in the original implementation of this PR because we hadn't yet moved the full requirements into conda-reqs.yml.

timsnyder-siv (commented):
Thanks for pushing this toward the finish line.  I only saw a couple of things I had questions or concerns about.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -50,6 +60,113 @@ def get_nbd_for_imagename(self, imagename: str) -> str:
...
+        if uri is None:
+            return None
+
+        # choose a repeatable, path based on the hash of the URI
+        destination = pjoin(local_dir, self.hashed_name(uri))
+        
+        return (uri, destination)
+
+    # download to a hashed destination name
+    # if the file exists this will not overwrite. Use this inside a
```
russell-horvath, line 124: Is this comment duplicated? Its basically the same as the one inside the function

russell-horvath (commented):
small thing in comment above*

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -172,6 +323,10 @@ def copy_sim_slot_infrastructure(self, slotno: int) -> None:
             run(f"mkdir -p {remote_sim_rsync_dir}")
 
             files_to_copy = serv.get_required_files_local_paths()
+
+            # append any URI paths to the end of this list
```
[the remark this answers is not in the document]
sifive-benjamin-morse: No. Each InstanceDeployManager object maintains a member, uri_list, which is a list of URI that need to be copied. This is not shared. Neighbor URI's are not copied over. Is there something I can change to make this more clear?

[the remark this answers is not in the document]
timsnyder-siv: I think I was confused by the singular `get_local_uri_path`.  If it was named `get_local_uri_paths`, it might have helped.  I just wanted to make sure you weren't copying the whole cache by passing rsync a single directory name to copy.

scripts/machine-launch-script.sh
sifive-benjamin-morse: Without these changes, `setup-manager` fails with this import path.
```
/opt/conda/envs/firesim/lib/python3.11/site-packages/paramiko/transport.py:178: CryptographyDeprecationWarning: Blowfish has been deprecated
  'class': algorithms.Blowfish,
Traceback (most recent call last):
  File "/home/centos/actions-runner-0/_work/firesim/firesim/./.github/scripts/install-firesim-pem.py", line 7, in <module>
    from common import manager_home_dir, manager_fsim_pem, set_fabric_firesim_pem
  File "/home/centos/actions-runner-0/_work/firesim/firesim/.github/scripts/common.py", line 6, in <module>
    from platform_lib import PlatformLib, AWSPlatformLib, AzurePlatformLib, Platform
  File "/home/centos/actions-runner-0/_work/firesim/firesim/.github/scripts/platform_lib.py", line 24, in <module>
    from awstools.awstools import get_instances_with_filter
  File "/home/centos/actions-runner-0/_work/firesim/firesim/deploy/awstools/awstools.py", line [33](https://github.com/firesim/firesim/actions/runs/4296629292/jobs/7488725823#step:3:34), in <module>
    from util.io import firesim_input
  File "/home/centos/actions-runner-0/_work/firesim/firesim/deploy/util/io.py", line 3, in <module>
    from fsspec.core import url_to_fs, open_local # type: ignore
    ```

(from https://github.com/firesim/firesim/actions/runs/4296629292/jobs/7488725823)

File: deploy/runtools/run_farm_deploy_managers.py [hunk 17]
```diff
@@ -444,6 +596,15 @@ def do_terminate():
...
 
+    @classmethod
+    def get_tar_name(cls) -> str:
+        """ Get the name of the tarball on the run host"""
+        return "driver-bundle.tar.gz"
+
+    @classmethod
+    def get_xclbin_name(cls) -> str:
+        """ Get the name of the xclbin on the run host"""
+        return "bitstream.xclbin"
```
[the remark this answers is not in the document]
sifive-benjamin-morse: @davidbiancolin do you have a better place to put these?

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4296740217. Verify CI state before submitting PR.

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -50,6 +60,113 @@ def get_nbd_for_imagename(self, imagename: str) -> str:
 
         return self.allocated_dict[imagename]
 
+class URIContainer:
+    """ A class which contains the details for downloading a single URI. InstanceDeployManager
+    contains a list of this class. """
```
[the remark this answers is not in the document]
sifive-benjamin-morse: OK this makes sense. I will try to paint a better picture for reviewers in the PR comments

deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -225,17 +380,14 @@ def start_sim_slot(self, slotno: int) -> None:
...
 
             # make the local job results dir for this sim slot
             server.mkdir_and_prep_local_job_results_dir()
-            sim_start_script_local_path = server.write_sim_start_script(slotno, has_sudo())
-            put(sim_start_script_local_path, remote_sim_dir)
 
             with cd(remote_sim_dir):
-                run("chmod +x sim-run.sh")
-                run("./sim-run.sh")
+                run(server.get_sim_start_command(slotno, (self.sim_command_requires_sudo() and has_sudo())))
```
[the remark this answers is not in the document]
sifive-benjamin-morse: I got this code by grabbing lots of things from https://github.com/firesim/firesim/compare/main...tsondhi/azure_validation_backup

I believe that this change was combined with a change to add a new property `sim_command_requires_sudo()` I

deploy/runtools/run_farm_deploy_managers.py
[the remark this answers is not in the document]
davidbiancolin: As far as location goes, i think it's OK to keep them in the class, either as methods or whatever the pythonic way is of definine class-scoped constants, though the XCLBIN method should only be defined in the VitisDeployManager or w.e if possible. 

I think the naming of the methods is more critical: `get_slot_relative_<name>` might be more along the lines of what @timsnyder-siv is asking for

scripts/machine-launch-script.sh
timsnyder-siv: Thanks.  I think I got Abe's attention to try and dedup the requirements a bit in #1445

timsnyder-siv (approved):
Thanks again for getting this really implemented @sifive-benjamin-morse

deploy/util/io.py
```diff
@@ -46,16 +46,18 @@ def downloadURI(uri: str, local_dest_path: str, retries: int = 4) -> None:
...
     if lpath.exists():
         rootLogger.debug(f"Overwriting {lpath.resolve(strict=False)}")
-    rootLogger.debug(f"Downloading '{uri}' to '{lpath}'")
 
-    for attempt in range(retries):
+    assert tries > 0, "tries argument must be larger than 0"
+    for attempt in range(tries):
+        rootLogger.debug(f"Download attempt {attempt+1} of {tries}: '{uri}' to '{lpath}'")
         try:
             fs, rpath = url_to_fs(uri)
```
timsnyder-siv, line 54: Are you sure you want the `url_to_fs()` call in the try block?  It seems like exceptions arising here may be indicative of a larger issue that shouldn't be retried?

sifive-benjamin-morse: For background, these two runs: 
* https://github.com/firesim/firesim/actions/runs/4301469200/jobs/7498730505
* https://github.com/firesim/firesim/actions/runs/4300595364/jobs/7496935845

fail for two different reasons on the same line:
* `aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host firesim-ci-vitis-xclbins.s3.***.amazonaws.com:443 ssl:default [Name or service not known]`
* `aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host firesim-ci-vitis-xclbins.s3.***.amazonaws.com:443 ssl:default [Temporary failure in name resolution]`

I added url_to_fs() inside the block because I wasn't sure if it did the DNS resolution or not. I will move it out. Thanks for reviewing this

PR #1438, opened 2023-02-23
VCS post-synthesis RTL simulators

nandor:
This PR adds make rules to get the post-synth RTL out of vivado and build a VCS simulator out of it. On VCS, a 150ns startup delay is hard-wired on all simulation modes to yield identical waveforms. The delay is required to initialize vivado gate-level libraries.

Running `make vcs-post-synth` and `make vcs-post-synth-debug` builds the simulators.

### Contributor Checklist

### DIFF
--- a/sim/midas/src/main/verilog/top.sv
+++ b/sim/midas/src/main/verilog/top.sv
@@ -107,6 +107,15 @@ import "DPI-C" function void simulator_tick
   output mem_fwd_t                                    mem_3_fwd
 );
 
+// Vivado gate-level models go through a setup phase through `glbl.v`. This delay 
+// ensures that the libraries initialise themselves beforehand. It effectively 
+// waits for sufficient time to elapse before the Vivado model delivers the reset 
+// signal to the relevant components. It is hard-wired here to ensure that the 
+// waveforms of RTL and gate-level simulations are identical.
+`ifndef VERILATOR
+`define VIVADO_INIT_DELAY 150ns
+`endif
+
 module emul(
 `ifdef VERILATOR
   input bit clock,
@@ -119,16 +128,15 @@ module emul(
   bit clock;
   reg fin = 1'b0;
   initial begin
-    clock = 1'b0;
+    #`VIVADO_INIT_DELAY;
     while (!fin) begin
-      clock = #(`CLOCK_PERIOD / 2.0) ~clock;
+      clock = #(`CLOCK_PERIOD * 1000 / 2.0) ~clock;
     end
   end
 
-  reg reset;
+  reg reset = 1'b1;
   initial begin
-    reset = 1'b1;
-    #(`CLOCK_PERIOD * 9.0) reset = 1'b0;
+    #(`VIVADO_INIT_DELAY + `CLOCK_PERIOD * 1000 * 9.0) reset = 1'b0;
   end
 `endif
 
--- a/sim/scripts/synth_fpga.tcl
+++ b/sim/scripts/synth_fpga.tcl
@@ -1,0 +1,120 @@
+# See LICENSE for license details.
+
+################################################################################
+# This file can be used to synthesize the `FPGATop` module of a design. It
+# emulates the strategy & frequency that the AWS flows might use.
+################################################################################
+
+################################################################################
+# Command-line Arguments
+################################################################################
+
+set strategy                [lindex $argv  0]
+set desired_host_frequency  [lindex $argv  1]
+set output                  [lindex $argv  2]
+set firesim_xdc             [lindex $argv  3]
+set firesim_sv              [lindex $argv  4]
+
+set device_type         "xcvu9p-flgb2104-2-i"
+set uram_option         2
+
+################################################################################
+# Read design files
+################################################################################
+
+if {$uram_option != 2} {
+  set_param synth.elaboration.rodinMoreOptions {rt::set_parameter disableOregPackingUram true}
+  set_param physynth.ultraRAMOptOutput false
+}
+
+# Maintain DONT TOUCH functionality for 2020.2 onwards
+if {[string match *2020.2* [version -short]] || [string match *2021.* [version -short]]} {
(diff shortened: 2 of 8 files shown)


sim/midas/src/main/verilog/top.sv
```diff
@@ -119,16 +119,15 @@ module emul(
   bit clock;
   reg fin = 1'b0;
   initial begin
-    clock = 1'b0;
+    #150ns;
```
davidbiancolin, line 122: Can you add a comment explaining this delay, and DRY it out using a macro.

Could you link to a doc explaining how to handle GSR under post-synthesis simulations as well? I'd prefer strongly that this be done in a standard way.

sim/scripts/synth_fpga.tcl
```diff
@@ -0,0 +1,132 @@
...
+        set uramHeight 4
+    }
+    default {
+        set synth_uram_option "-max_uram_cascade_height 1"
+        set uramHeight 4
+    }
+}
+
+switch $strategy {
+    "BASIC" {
```
davidbiancolin, line 67: Could you somehow pull in these values from the underlying strategies in AWS-FPGA?

davidbiancolin (commented):
I'd strongly prefer it if the generation of post-synth netlists was better coupled to each of the platform build systems, since the output of Vivado is very sensitive to different synthesis directives / options, but this is a valuable addition.

sim/scripts/synth_fpga.tcl
nandor: I'd rather keep these separate.

nandor:
Coupling with shims adds a lot of overhead. It prototyped changes for that as well, but they add a lot of complexity. This simulation is useful on its own, so I would like to see it in action.

sim/scripts/synth_fpga.tcl
nandor: We'll want to tweak these options and it's easier to do if they're in this repository.

nandor: Also, I want this to be a bit target-independent in the hopes of finding vivado messups.

sim/scripts/synth_fpga.tcl
```diff
@@ -0,0 +1,132 @@
...
+# Licensed under the Amazon Software License (the "License"). You may not use
+# this file except in compliance with the License. A copy of the License is
+# located at
+#
+#    http://aws.amazon.com/asl/
+#
+# or in the "license" file accompanying this file. This file is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
```
timsnyder-siv, line 17: Is this file being copied from aws-fpga into firesim?  Why?  It would be helpful to have a file-level comment  explaination of why it is being vendored into firesim and what this copy does that the original didn't do.

sim/midas/src/main/verilog/top.sv
timsnyder-siv: Using a macro was half of what David asked for.  Am I missing the link to the doc about GSR in post-synth simulations because I don't see it.

nandor: I will not be writing a doc here and now.

timsnyder-siv: I don't think he was asking you to write a doc.  Where does the 150ns come from?  There are gate-level libraries coming in from vivado that need this delay, right?  How did you determine that it needed to be 150ns?  Manual inspection of the waves or looking at verilog or is there some Xilinx doc that told you about the 150ns.

Perhaps I'm way off-base and he was asking you to write a doc, in which case, it's fine for that to be a follow-up.

nandor: There is a magic `glbl.v` in Vivado which hard-wires a similar delay. This is slightly higher to give some wiggle-room. I had a comment added before but I lost it when I pushed from a different machine, it should be there now.

sim/scripts/synth_fpga.tcl
nandor: Added another comment explaining the purpose of this file.

sim/midas/src/main/verilog/top.sv
timsnyder-siv: Thanks for adding the comment.  It satisfies my desire to have something more explaining the magic 150ns.

sim/midas/src/main/verilog/top.sv
```diff
@@ -119,16 +128,15 @@ module emul(
...
+      clock = #(`CLOCK_PERIOD * 1000 / 2.0) ~clock;
     end
   end
 
-  reg reset;
+  reg reset = 1'b1;
   initial begin
-    reset = 1'b1;
-    #(`CLOCK_PERIOD * 9.0) reset = 1'b0;
+    #(`VIVADO_INIT_DELAY + `CLOCK_PERIOD * 1000 * 9.0) reset = 1'b0;
```
timsnyder-siv, line 139: Why does `\`CLOCK_PERIOD` need to be multiplied by 1000 as a part of this PR?

timsnyder-siv: Oh, I see it.  You did change the timescale from `1ns/1ps` to `1ps/1ps`.  Why was that necessary?

nandor: Vivado sets that timescale in `glbl` for something. I'd rather use the smallest unit, if possible, than mix timescales.

timsnyder-siv: Makes sense.  We need to also adjust the timescale in Makefrag-verilator as well then.  It would probably be good to mention the timescale change higher in the PR so that it isn't buried in here.

fabianschuiki (approved):
Glad to see post-synth simulation capability!

timsnyder-siv:
@nandor did you adjust the verilator timescale before merging this?

PR #1443, opened 2023-02-27
pickup fab-classic#77 by bumping to 1.19.2

timsnyder-siv:
fab-classic has logic to react to paramiko `SSHException` raised when server load is too high.  In this case, fab-classic should retry the connection until it gets to `env.connection_attempts` (see https://github.com/firesim/firesim/blob/0755382dae22e5be1e933a3e888bcadc1a7354c9/deploy/firesim#L437) and then raise `NetworkError`.

The logic in fab-classic 1.19.1 does not correctly match the paramiko exception message in all the required cases and incorrectly stops retrying the connection before reaching `env.connection_attempts` (it falls down a path where it aborts because the code thinks it should be querying the user for a password).

This fix makes the firesim manager more robust in environments where the server networking performance is struggling to keep up.

We may further want to catch `NetworkError` when in the `runworkload` monitoring loop so that hosts could be marked `???` for as long as we're unable to connect to them, rather than causing the monitoring loop to die so that the user is responsible for manually monitoring their workload and collecting results.  I leave that for a future PR.

#### Related PRs / Issues

fab-classic#77 fixing fab-classic#76

#### UI / API Impact

No impact

No impact.

### Contributor Checklist
- [X] Did you add Scaladoc/docstring/doxygen to every public function/method? **n/a**
- [X] Did you add at least one test demonstrating the PR? **n/a**

### DIFF
--- a/.github/scripts/requirements.txt
+++ b/.github/scripts/requirements.txt
@@ -1,4 +1,4 @@
-fab-classic==1.19.1
+fab-classic==1.19.2
 boto3==1.20.21
 pytz
 pyyaml
--- a/conda-reqs.conda-lock.yml
+++ b/conda-reqs.conda-lock.yml
@@ -5123,14 +5123,14 @@ package:
     paramiko-ng: '*'
     six: '>=1.10.0'
   hash:
-    sha256: 7fe3dfd0d9d5d0dd7e650b42fc7d62ec5d643ac4275a77f483ec2b57f19c3e58
+    sha256: 8edfd97ff58ca616cdd6e77bc42d3f71d27842197f13771c77cf0b553f3b4311
   manager: pip
   name: fab-classic
   optional: false
   platform: linux-64
   source: null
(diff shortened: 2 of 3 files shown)


.github/scripts/requirements.txt
```diff
@@ -1,4 +1,4 @@
-fab-classic==1.19.1
+fab-classic==1.19.2
```
timsnyder-siv, line 1: @abejgonzalez why do we have this separate requirements.txt?  Is this vestigial and can it be removed now?  It would be good for this file to have a comment explaining what requirements are kept here and why.  I suspect this is only stuff needed to bootstrap something running `machine_launch_script.sh` and that it is a very minimal set of things needed by a single script.

@t14916 & @sifive-benjamin-morse 👀

abejgonzalez: Correct. In order to spawn an AWS instance manager, we need to first boot into a GH managed VM, install the Fabric/AWS deps, then launch the manager.

https://github.com/firesim/firesim/blob/38a02b3b9083a6c98d1afef75c7b64d4a7bab22c/.github/workflows/firesim-run-tests.yml#L103

abejgonzalez (approved):
LGTM

PR #1451, opened 2023-03-03
Unpin most conda reqs now that we have a lockfile

abejgonzalez:
We shouldn't need to pin all dependencies in the conda requirements file (only the minimal set of requirements) now that we by default use the conda lock file to "pin" all dependencies to a working version. This helps the solver in the future choose the right set of packages. Additionally, adds a small fix to non-sudo guestmount logic so that corresponding screen sessions are referenced by a unique hash (occurred in this PR's CI).

I was a bit aggressive and unpinned everything assuming things would work. If people need me to pin specific packages we should update this PR.

### Contributor Checklist

### DIFF
--- a/conda-reqs/firesim.yaml
+++ b/conda-reqs/firesim.yaml
@@ -36,7 +36,7 @@ dependencies:
     - qemu # from ucb-bar channel - https://github.com/ucb-bar/qemu-feedstock
     - rsync
     - psutil
-    - doit=0.35.0
+    - doit>=0.35.0
     - gitpython
     - humanfriendly
     - e2fsprogs
@@ -101,23 +101,23 @@ dependencies:
     # conda environment, pip's dependency resolution can conflict with
     # conda and create broken environments.  It's best to use the conda
     # packages so that the environment is consistent
-    - colorama==0.4.3
-    - argcomplete==1.12.3
-    - python-graphviz==0.19
-    - pyparsing==3.0.6
-    - numpy==1.19.5
-    - kiwisolver==1.3.1
-    - matplotlib-base==3.3.4
-    - pandas==1.1.5
-    - awscli==1.22.21
-    - pytest==6.2.5
-    - pytest-dependency==0.5.1
-    - pytest-mock==3.7.0
-    - moto==3.1.0
-    - mypy==0.931
+    - colorama
+    - argcomplete
+    - python-graphviz
+    - pyparsing
+    - numpy
+    - kiwisolver
+    - matplotlib-base
+    - pandas
+    - awscli
+    - pytest
+    - pytest-dependency
+    - pytest-mock
+    - moto
+    - mypy
     - s3fs
     - aiohttp
     - pip
     - pip:
-        - sure==2.0.0
-        - pylddwrap==1.2.1
+        - sure
+        - pylddwrap
--- a/conda-reqs/firesim.yaml
+++ b/conda-reqs/firesim.yaml
@@ -30,7 +30,7 @@ dependencies:
     - binutils
 
     - dromajo # from ucb-bar channel - https://github.com/riscv-boom/dromajo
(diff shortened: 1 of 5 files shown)


abejgonzalez:
This PR is blocking bumping to the most recent RISC-V toolchain so merging this in as soon as possible would be great (assuming it's fine with you all and the CI passes)

t14916:
Metasim is failing? Looks like one of the vcs failed, possibly due to flakiness?

conda-reqs/firesim.yaml
```diff
@@ -30,13 +30,13 @@ dependencies:
...
     - dromajo # from ucb-bar channel - https://github.com/riscv-boom/dromajo
-    - riscv-tools # from ucb-bar channel - https://github.com/ucb-bar/riscv-tools-feedstock
+    - riscv-tools=1.0.1 # from ucb-bar channel - https://github.com/ucb-bar/riscv-tools-feedstock
 
     # firemarshal deps
     - qemu # from ucb-bar channel - https://github.com/ucb-bar/qemu-feedstock
     - rsync
     - psutil
-    - doit=0.35.0
+    - doit>=0.35.0
```
timsnyder-siv, line 39: I thought this This was added with `=`, referencing a firemarsal issue in the commit message.  However, I went back through things and it looks like Abe added it in https://github.com/firesim/firesim/commit/aed821b504607ec8cc33187de1a892f2fee9c4b7 without really pointing to why.  The firemarshal reqs say `>0.34`

timsnyder-siv:
As an aside, I noticed recently that python is effectively being pinned by the ucb-bar/riscv-tools to 3.8. (I don't think anything in riscv-tools links against the python dynamic lib and it shouldn't even been a runtime dep).

 If you are interested in fixing that, let me know.  It will require reworking the riscv-tools recipe a little to split the `build` and `host` requirements.

conda-reqs/firesim.yaml
abejgonzalez: Good catch. This can be `>=0.34` based on this: https://github.com/firesim/FireMarshal/pull/242.

abejgonzalez:
> Metasim is failing? Looks like one of the vcs failed, possibly due to flakiness?

I'll look into it right now.

> As an aside, I noticed recently that python is effectively being pinned by the ucb-bar/riscv-tools to 3.8. (I don't think anything in riscv-tools links against the python dynamic lib and it shouldn't even been a runtime dep).
> 
> If you are interested in fixing that, let me know. It will require reworking the riscv-tools recipe a little to split the `build` and `host` requirements.

Sure. Looking at the recipe it looks like we don't specify a Python version but if something is off, I'd be happy to merge in a change.

abejgonzalez:
I tested FireMarshal on `doit>=0.34` and it seemed to work. The error that the local FPGA setup has is weird, it errors saying an old CMake version is found... but cmake was a higher version. I re-updated the lock file to try again.

timsnyder-siv:
> Sure. Looking at the recipe it looks like we don't specify a Python version but if something is off, I'd be happy to merge in a change.

It's a weird side-effect of not having the `requiements.host` section in your recipe.  Everything in `requirements.build` that has dependencies that get exported to runtime deps ends up in your runtime deps.  I'm going to see what problem you ran into when using the `requirements.host` section and if it is too annoying to fix that way, we can just override the runtime dep that python is injecting.

I opened https://github.com/ucb-bar/riscv-tools-feedstock/pull/5

abejgonzalez:
Perfect. I'll review the PR later today/tomorrow when I get the chance. In the interim, this PR should be good to merge (pending any other changes to the requirements).

abejgonzalez:
Ping @t14916 @timsnyder-siv for re-review

t14916 (approved):
LGTM, idk that there is anything I see here that needs to be pinned.

PR #1460, opened 2023-03-08
Support AL2 manager instances

abejgonzalez:
AL2 based manager instances are nice in that they have support for an in-browser based terminal (useful for future FireSim tutorials). To help support this, I've abstracted out the `centos` keyword from the manager (now it is auto-determined by the `whoami` call throughout the codebase). Additionally, I've moved the machine launch log/status file to the `/tmp` directory since on AL2 there were weird issues with the machine launch script not being able to write to `/`. I also thought this was a generally better location (not sure why things were placed in `/` by default). 

Unfortunately, the one downside of using the AL2-based AMI is that it runs 2021.1 (not 2021.2) Vivado. AL2 is supposed to be supported in the future - https://github.com/aws/aws-fpga/issues/600 - so hopefully when the centos AMI is updated the AL2 one will be updated as well.

Since this not tested in CI, I've avoided adding this to the documentation (experimental hidden feature). Once this is tested in CI we can add some docs on it (i.e. you need to update your `default_{build/sim}_dir` in the `yaml` files)

### Contributor Checklist

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -39,7 +39,15 @@ rootLogger = logging.getLogger()
 # https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#Images:visibility=public-images;search=FPGA%20Developer;sort=name
 # And whenever this changes, you also need to update deploy/tests/test_amis.json
 # by running scripts/update_test_amis.py
-f1_ami_name = "FPGA Developer AMI - 1.12.1-40257ab5-6688-4c95-97d1-e251a40fd1fc"
+def get_f1_ami_name() -> str:
+    cuser = local("whoami", capture=True)
+    if cuser == "centos":
+        return "FPGA Developer AMI - 1.12.1-40257ab5-6688-4c95-97d1-e251a40fd1fc"
+    elif cuser == "amzn":
+        return "FPGA Developer AMI(AL2) - 1.11.3-62ddb7b2-2f1e-4c38-a111-9093dcb1656f"
+    else:
+        assert False, "Unknown user given by 'whoami' (expected centos/amzn). Are you running on AWS EC2?"
+        return ""
 
 class MockBoto3Instance:
     """ This is used for testing without actually launching instances. """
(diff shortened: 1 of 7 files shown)


deploy/runtools/run_farm_deploy_managers.py
```diff
@@ -234,10 +237,10 @@ def sim_node_qcow(self) -> None:
         if self.nbd_tracker is not None and self.parent_node.qcow2_support_required():
             self.instance_logger("""Setting up remote node for qcow2 disk images.""")
             # get qemu-nbd
-            ### XXX Centos Specific
+            ### XXX ec2-user Specific
```
sagark, line 240: bad global replace?

sagark (approved):

deploy/awstools/awstools.py
```diff
@@ -39,7 +39,15 @@
 # https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#Images:visibility=public-images;search=FPGA%20Developer;sort=name
 # And whenever this changes, you also need to update deploy/tests/test_amis.json
 # by running scripts/update_test_amis.py
-f1_ami_name = "FPGA Developer AMI - 1.12.1-40257ab5-6688-4c95-97d1-e251a40fd1fc"
+def get_f1_ami_name() -> str:
+    cuser = local("whoami", capture=True)
```
timsnyder, line 43: Is there an advantage to calling out to a shell for `whoami` potentially many times vs using `env['USER']`?

Also, are you expecting that the AMI being used by the manager and runhosts needs to match?  I know this is silent support for AL2 but you might want to put a comment in this method documenting the assumption for future devs if you aren't documenting it in the actual docs.

abejgonzalez: No advantage. I can change it to looking at `env['USER']`.

Right now that is the current assumption (I'm not sure anyone else has done otherwise). I'll add a comment.

sagark (approved):

abejgonzalez:
Linux boots were fixed in another PR #1466 . I manually checked the CI on this PR and it looks correct. Merging.

PR #1464, opened 2023-03-09
Bump bitstreams and fix AWS Linux Boot

abejgonzalez:
Update the AGFIs/xclbins and fixes AWS Linux boot issue that was exposed in #1442 (the check for `profile_interval == -1` was removed, but the FireSim manager still defaults to `-1` for all workloads in the `config_runtime.yaml` - https://github.com/firesim/firesim/pull/1442/files#diff-09c0be21679335bf6c7efb037debdce62e0139ece96559bf14cd3171cf50d21fL62). 

### Contributor Checklist

### DIFF
--- a/deploy/sample-backup-configs/sample_config_hwdb.yaml
+++ b/deploy/sample-backup-configs/sample_config_hwdb.yaml
@@ -11,23 +11,23 @@
 
 # DOCREF START: Example HWDB Entry
 firesim_boom_singlecore_nic_l2_llc4mb_ddr3:
-    agfi: agfi-0d460a200ae495922
+    agfi: agfi-00668528c06eb1d68
     deploy_triplet_override: null
     custom_runtime_config: null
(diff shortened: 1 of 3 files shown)


abejgonzalez:
Ok I should have fixed the AWS issue. Vitis bitstreams will be fixed in a future PR (it should still fail in this PR).

davidbiancolin:
Can you explain more about the root cause here? Why did the assertion fire? You should be able to specify -1 as a value and it should be treated as a max uint64_t.

That aside, a better fix would be to allow setting `null` in the manager for the profile interval, which would elide passing the +arg.

davidbiancolin:
I read the code. I suspect this code is still broken for any profile interval. The hackiest "fix" is to relax the forward progress assertion to exclude time 0, or to change the check to assert that forward progress is made in at least one of every two consective trips

davidbiancolin:
I think the latter check is more sound. Please do that.

PR #1478, opened 2023-03-22
Fix conda lockfile docs / make lockfile generation easier

t14916:
Creates a callable script that is also used by build-setup.sh to generate the conda lockfile, which makes it easier to make changes to dependencies / move dependencies between ci-shared and general.
#### UI / API Impact

Only UI change is instead of manually running conda-lock, there is now a script that generates the lockfile.
### Contributor Checklist

### DIFF
--- a/build-setup-nolog.sh
+++ b/build-setup-nolog.sh
@@ -136,10 +136,9 @@ if [ "$IS_LIBRARY" = true ]; then
     fi
 else
     # note: lock file must end in .conda-lock.yml - see https://github.com/conda-incubator/conda-lock/issues/154
-    LOCKFILE="$RDIR/conda-reqs/conda-reqs.conda-lock.yml"
     if [ "$USE_PINNED_DEPS" = false ]; then
         # auto-gen the lockfile
-        conda-lock -f "$RDIR/conda-reqs/firesim.yaml" -f "$RDIR/conda-reqs/ci-shared.yaml" -p linux-64 --lockfile "$LOCKFILE"
+	./scripts/generate-conda-lockfile.sh
     fi
     conda-lock install -p $RDIR/.conda-env $LOCKFILE
     source $RDIR/.conda-env/etc/profile.d/conda.sh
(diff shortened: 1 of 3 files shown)


build-setup-nolog.sh
```diff
@@ -136,10 +136,9 @@ if [ "$IS_LIBRARY" = true ]; then
     fi
 else
     # note: lock file must end in .conda-lock.yml - see https://github.com/conda-incubator/conda-lock/issues/154
-    LOCKFILE="$RDIR/conda-reqs/conda-reqs.conda-lock.yml"
```
abejgonzalez, line 139: You still need to point to the lock file to use in the `conda-lock install` step below.

abejgonzalez: Maybe you can make the path it dumps to a argument to the script?

abejgonzalez (changes requested):

build-setup-nolog.sh
t14916: I'd prefer it if there was a single source of truth for the path to the lockfile, and for that to live in the script generating the lockfile (so it can be called without any args). Maybe build-setup can read the script and get the path that way?

abejgonzalez: Hmm. What about this... the script has the hardcoded path (something like `RDIR/conda-reqs/lock/lockfile.conda-lock.yml`. Then in the `build-setup-nolog.sh` script does a `find` to get the name of the file looking in that exact location (`RDIR/conda-reqs/lock`). Or something like that (maybe you don't need the extra directory and instead just search in `conda-reqs` for any file that ends with `conda-lock.yml`).

scripts/generate-conda-lockfile.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+REQS_DIR="$(pwd)/conda-reqs"
```
abejgonzalez, line 3: This should be more robust so that it can be called in multiple places. I would use the following: https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script

abejgonzalez (approved):
LGTM

PR #1492, opened 2023-04-20
Create separate security group for build/run farm instances that is only accessible from within the firesim VPC

sagark:
The new setup is:
* `firesim` security group is the same as before, used for managers
* `for-farms-only-firesim` security group is created in `firesim managerinit` if it does not already exist in the user's account. This is created with ingress allowed only from within the VPC.

This means all users will get naturally updated to the new setup the first time they run firesim managerinit with a version of firesim with this PR merged in. Old firesim installations/managers will continue to work as normal.

The CI setup should also get upgraded naturally via this process.

Added a clarifying note to the manager setup docs that `for-farms-only-firesim` should not be used for managers. First-time users will not have this security group when they create their first manager instance anyway.

#### Related PRs / Issues

N/A

#### UI / API Impact

We always instructed accessing build/run farm instances from the manager, so no change. Users will get an extra message the first time they run firesim managerinit on an aws account stating that this new security group is being created.

N/A

### Contributor Checklist

### DIFF
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -229,7 +229,9 @@ def farm_security_group_setup() -> None:
 
     if len(firesimsecuritygroup) > 1:
         rootLogger.critical(f"Too many security groups named {securitygroupname}. Exiting.")
+        assert False
     elif len(firesimsecuritygroup) == 1:
+        rootLogger.debug(f"Security group {securitygroupname} already exists. Skipping setup.")
         return
 
     # at this point, we do not have the required security group, so create it
--- a/deploy/awstools/awstools.py
+++ b/deploy/awstools/awstools.py
@@ -181,7 +181,7 @@ def aws_resource_names() -> Dict[str, Any]:
         'tutorial_mode'  :   False,
         # regular users are instructed to create these in the setup instructions
         'vpcname':           'firesim',
-        'securitygroupname': 'firesim',
+        'securitygroupname': 'for-farms-only-firesim',
         # regular users are instructed to create a key named `firesim` in the wiki
         'keyname':           'firesim',
         's3bucketname' :     None,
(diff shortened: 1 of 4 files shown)


deploy/awstools/awstools.py
```diff
@@ -209,6 +209,56 @@ def aws_resource_names() -> Dict[str, Any]:
...
+    ec2 = boto3.resource('ec2')
+    client = boto3.client('ec2')
+
+    operation_params = {
+        'Filters': [{'Name':'group-name', 'Values': [securitygroupname]}]
+    }
+    firesimsecuritygroup = depaginated_boto_query(client, 'describe_security_groups', operation_params, 'SecurityGroups')
+
+    if len(firesimsecuritygroup) > 1:
+        rootLogger.critical(f"Too many security groups named {securitygroupname}. Exiting.")
```
abejgonzalez, line 229: Should this error?

deploy/awstools/awstools.py
```diff
@@ -209,6 +209,56 @@ def aws_resource_names() -> Dict[str, Any]:
...
+    # managers and other build/run farm instances
+    allowed_cidr = '192.168.0.0/16'
+
+    sec_group.authorize_ingress(IpPermissions=[
+        {u'PrefixListIds': [], u'FromPort': 60000, u'IpRanges': [{u'Description': 'mosh', u'CidrIp': allowed_cidr}], u'ToPort': 61000, u'IpProtocol': 'udp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
+        {u'PrefixListIds': [], u'FromPort': 22, u'IpRanges': [{u'CidrIp': allowed_cidr}], u'ToPort': 22, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
+        {u'PrefixListIds': [], u'FromPort': 10000, u'IpRanges': [{u'Description': 'firesim network model', u'CidrIp': allowed_cidr}], u'ToPort': 11000, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
+        {u'PrefixListIds': [], u'FromPort': 3389, u'IpRanges': [{u'Description': 'remote desktop', u'CidrIp': allowed_cidr}], u'ToPort': 3389, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
+        {u'PrefixListIds': [], u'FromPort': 8443, u'IpRanges': [{u'Description': 'nice dcv (ipv4)', u'CidrIp': allowed_cidr}], u'ToPort': 8443, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []},
```
abejgonzalez, line 256: How does this work with NICE DCV? I assume you don't need the Remote Desktop and NICE DCV part of this?

deploy/awstools/awstools.py
```diff
@@ -209,6 +209,56 @@ def aws_resource_names() -> Dict[str, Any]:
...
+    client = boto3.client('ec2')
+
+    operation_params = {
+        'Filters': [{'Name':'group-name', 'Values': [securitygroupname]}]
+    }
+    firesimsecuritygroup = depaginated_boto_query(client, 'describe_security_groups', operation_params, 'SecurityGroups')
+
+    if len(firesimsecuritygroup) > 1:
+        rootLogger.critical(f"Too many security groups named {securitygroupname}. Exiting.")
+    elif len(firesimsecuritygroup) == 1:
```
abejgonzalez, line 230: Can you add a print here indicating that the security group was found and thus not created.

abejgonzalez (changes requested):
Can you fix the PR title? Otherwise, seems fine (minus the nits)

deploy/awstools/awstools.py
sagark: Good catch, fixed.

deploy/awstools/awstools.py
sagark: I'll leave it for convenience. Doesn't hurt.

deploy/awstools/awstools.py
sagark: Done

abejgonzalez (approved):

PR #1495, opened 2023-04-27
Manager support for custom TARGET_PROJECT

sagark:
This PR updates the manager to support arbitrary `TARGET_PROJECT`s instead of only `firesim`. In effect, all mentions of `triplet`s (consisting of `(DESIGN, TARGET_CONFIG, PLATFORM_CONFIG)`) become `quadruplet`s (consisting of `(TARGET_PROJECT, DESIGN, TARGET_CONFIG, PLATFORM_CONFIG)`).

This adds support for all manager features except producing the command to run a simulation (this needs some more thinking). But building bitstreams/metasims and doing everything up to and including infrasetup (and then running the sim by hand on the run farm instance) works as expected.

This should be fully backwards compatible with existing AGFIs and config yaml files (see below).

This also adds a fix so that the manager doesn't complain about `deploy_quadruplet_override` (formerly `deploy_triplet_override`) being set for vitis bitstreams. FYI: @abejgonzalez 

Internally, you will notice that some mentions of "triplet" remain. These are because the make-system still relies on triplets for naming several things (e.g. directories in `sim/generated-src/f1/`). Updating this is punted to a future PR since this is only really a problem if you work with multiple different `quadruplet`s that have the same `triplet`, which is rare. This has no effect on functionality besides this case.

#### Related PRs / Issues

N/A

#### UI / API Impact

* Entries in `config_hwdb.yaml` now have a `deploy_quadruplet_override` key instead of `deploy_triplet_override`.
* Entries in `config_build_recipes.yaml` now have a `deploy_quadruplet` key instead of `deploy_triplet`.

For both, backwards compatibility is ensured like so:
1. The manager will print a warning if only `deploy_triplet_override` or `deploy_triplet` are defined, instructing the user to upgrade, but will otherwise work as usual.
2. The manager will error if both `deploy_triplet_override` and `deploy_quadruplet_override` are defined or both `deploy_triplet` and `deploy_quadruplet` are defined.
3. The manager will accept "triplet" values supplied for any of these keys and automatically assume `TARGET_PROJECT=firesim` for any triplet. Otherwise, quadruplets are expected and all defaults in config files and docs have been updated to quadruplets.

The serialized versions of triplets stored in AGFI descriptions (`firesim-buildtriplet` and `firesim-deploytriplet`) continue to be kept so that old managers can continue to use AGFIs generated with the version of the manager in this PR. For AGFIs tagged only with a triplet, the manager assumes `TARGET_PROJECT=firesim`. This version of the manager also adds `firesim-buildquadruplet` and `firesim-deployquadruplet` key/values on AGFIs for the new `TARGET_PROJECT` support. Old managers will ignore these keys silently.

I have tested:
* Existing AGFIs booting Linux
* Non-firesim TARGET_PROJECT AGFI build
* Non-firesim TARGET_PROJECT FPGA-run (infrasetup + run by hand on run farm)
* Non-firesim TARGET_PROJECT metasims (infrasetup + run by hand on run farm)

See note above. No user facing changes. No need for AGFI re-gen.

### Contributor Checklist

### DIFF
--- a/deploy/awstools/afitools.py
+++ b/deploy/awstools/afitools.py
@@ -104,9 +104,10 @@ def share_agfi_in_all_regions(agfi_id, useridlist):
         afi_id = get_afi_for_agfi(agfi_id, region)
         share_afi_with_users(afi_id, region, useridlist)
 
-def firesim_tags_to_description(buildtriplet, deploytriplet, commit):
+def firesim_tags_to_description(build_quadruplet, deploy_quadruplet, build_triplet, deploy_triplet, commit):
     """ Serialize the tags we want to set for storage in the AGFI description """
-    return """firesim-buildtriplet:{},firesim-deploytriplet:{},firesim-commit:{}""".format(buildtriplet,deploytriplet,commit)
+    # note that the serialized rep still includes "triplets" for compat
+    return """firesim-buildquadruplet:{},firesim-deployquadruplet:{},firesim-buildtriplet:{},firesim-deploytriplet:{},firesim-commit:{}""".format(build_quadruplet,deploy_quadruplet,build_triplet,deploy_triplet,commit)
 
 def firesim_description_to_tags(description):
     """ Deserialize the tags we want to read from the AGFI description string.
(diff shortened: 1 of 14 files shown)


abejgonzalez (approved):
SFTM Overall

PR #1497, opened 2023-05-03
Bare Xilinx U250/U280 shell support

abejgonzalez:
This adds support for a Xilinx U250/U280 shell (originally created by the folks over at NTNU - https://github.com/EECS-NTNU/firesim/tree/1.15.1_alveo) to the manager. This shell comes with support for the MMIO, DMA, and a single DRAM channel! Additionally, this PR updates the documentation to mention both flows (the Vitis flow or the bare shell U250/U280 flow). Additionally, this PR adds support for tar'ing the bit file with a metadata file (similar to tags that the AGFI has that mentions the deploy quadrupled).

- [x] Make sure the commit message has proper authorship. @bgottschall @davidmetz let me know if the commit message authorship points to the right email address / has the right name.
- [x] Finalize the name of platform. Right now new platform is called `xilinx_alveo_u2*0` (different than the `au2*0` naming that Xilinx board files and OpenNIC shell uses).
- [x] Consider supporting multiple DRAM channels (maybe push this to a later PR) - *Push to later PR*
- [ ] Verify that programming the FPGA (i.e. removing/rescanning the PCI-E device) doesn't affect other simulations running.
- [ ] Upload a pre-built bit file HWDB entry
- [x] Consider adding CI for this (maybe push this to a later PR) - *Push to later PR*
- [x] Copy U280 shell
- [x] Change documentation to have a "setup" FPGAs section before the tutorial(s)
- [x] Fix output paths across all FireSim (now use quintuplet).

Last I checked (before some refactoring) this shell was able to build arbitrary designs and boot simulations with `sudo` access!

### Contributor Checklist


bgottschall:
Nitpicking... Can you change my first name to 'Björn'?

Also, I think it makes more sense to use info@gottschall.no as my email address since the NTNU mail won't be valid for long anymore.

abejgonzalez:
> Nitpicking... Can you change my first name to 'Björn'?
> 
> Also, I think it makes more sense to use [info@gottschall.no](mailto:info@gottschall.no) as my email address since the NTNU mail won't be valid for long anymore.

Absolutely, I'll change your name and email.

abejgonzalez:
@bgottschall When comparing your https://github.com/EECS-NTNU/fpga-util/blob/main/fpga-util.py to the OpenNIC FPGA programming script (which we adapt)(https://github.com/Xilinx/open-nic-shell/blob/main/script/program_fpga.sh). It looks like you don't remove the bridge BDF before flashing the FPGA. Instead you do some `setpci` calls before then after the flash you do the `remove` + `rescan`. Why is that?

bgottschall:
> @bgottschall When comparing your https://github.com/EECS-NTNU/fpga-util/blob/main/fpga-util.py to the OpenNIC FPGA programming script (which we adapt)(https://github.com/Xilinx/open-nic-shell/blob/main/script/program_fpga.sh). It looks like you don't remove the bridge BDF before flashing the FPGA. Instead you do some `setpci` calls before then after the flash you do the `remove` + `rescan`. Why is that? We generally want to avoid rebooting the machine.

We have adapted this flow for flashing the FPGAs (there is a shell script in one of the answers): https://support.xilinx.com/s/question/0D52E00006hpjPHSAY/dell-r720-poweredge-server-reboots-on-fpga-reprogramming?language=en_US 

It disables error reporting and clears an error bit when a PCIe ID disappears (happens only when the PCIe ID changes e.g. going from a Vitis shell to our shell) and shouldn't be necessary if the FPGA is already flashed with a previous shell that has the same ID. 
Without those setpci commands it crashed sometimes if I remember correctly, which should be tested with the OpenNIC script and changing between shells and PCIe IDs (including soft reboots).

.gitignore
```diff
@@ -10,3 +10,5 @@ build
 scala-doc-env.sh
 *.swp
 /.conda-env
+.metals
+vivado*
```
sagark, line 14: gitignoring everything starting with vivado in the whole repo doesn't seem like a good idea

deploy/awstools/afitools.py
```diff
@@ -104,14 +104,13 @@ def share_agfi_in_all_regions(agfi_id, useridlist):
         afi_id = get_afi_for_agfi(agfi_id, region)
         share_afi_with_users(afi_id, region, useridlist)
 
-def firesim_tags_to_description(build_quadruplet, deploy_quadruplet, build_triplet, deploy_triplet, commit):
+def firesim_tags_to_description(build_quintuplet, deploy_quintuplet, commit):
     """ Serialize the tags we want to set for storage in the AGFI description """
-    # note that the serialized rep still includes "triplets" for compat
-    return """firesim-buildquadruplet:{},firesim-deployquadruplet:{},firesim-buildtriplet:{},firesim-deploytriplet:{},firesim-commit:{}""".format(build_quadruplet,deploy_quadruplet,build_triplet,deploy_triplet,commit)
+    return """firesim-buildquintuplet:{},firesim-deployquintuplet:{},firesim-commit:{}""".format(build_quintuplet,deploy_quintuplet,commit)
```
sagark, line 109: need to keep triplet keys here for backwards compat

deploy/sample-backup-configs/sample_config_hwdb.yaml
```diff
@@ -12,54 +12,64 @@
...
-    deploy_quadruplet_override: firesim-FireSim-FireSimRocketMMIOOnlyConfig-BaseVitisConfig
+    deploy_quintuplet_override: firesim-FireSim-FireSimRocketMMIOOnlyConfig-BaseVitisConfig
     custom_runtime_config: null
 vitis_firesim_gemmini_rocket_singlecore_no_nic:
     xclbin: https://firesim-ci-vitis-xclbins.s3.us-west-2.amazonaws.com/vitis_firesim_gemmini_rocket_singlecore_no_nic_1ea5c4.xclbin
-    deploy_quadruplet_override: firesim-FireSim-FireSimLeanGemminiRocketMMIOOnlyConfig-BaseVitisConfig
+    deploy_quintuplet_override: firesim-FireSim-FireSimLeanGemminiRocketMMIOOnlyConfig-BaseVitisConfig
     custom_runtime_config: null
+# DOCREF START: Xilinx Alveo HWDB Entries
+alveou250_firesim_rocket_singlecore_no_nic:
```
sagark, line 67: can you replace all instances of "alveou250" with "alveo_u250" (same for 280)?

docs/Advanced-Usage/Manager/Manager-Configuration-Files.rst
```diff
@@ -526,11 +533,18 @@ Indicates where the bitstream (FPGA Image) is located, may be one of:
   * A Uniform Resource Identifier (URI), (see :ref:`uri-path-support` for details)
   * A filesystem path available to the manager. Local paths are relative to the `deploy` folder.
 
-``deploy_quadruplet_override``
-""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+``bit``
```
sagark, line 536: can we call this bitstream instead of just bit

abejgonzalez: I'll make this `bitstream_tar`

abejgonzalez:
Uncaught 2 FPGA instance shutdown(s) detected for CI run: 4909899866. Verify CI state before submitting PR.

bgottschall:
Currently verifying this PR on our setup and while this is in progress I'm already having some comments. The PCI-E ids or BDFs should be specified per FPGA in the run-farm-recipes (externally_provisioned.yaml). I'm working on a host with 2 FPGAs, one u250 and one u280. The deploy manager currently just greps the n'th PCI-E id from the lspci command for which both FPGAs look exactly the same. This can be a fallback in case the BDFs are not specified in the run-farm however in my case I must be able to specify exactly which ID I use with which deploy scripts.

The next thing is a nice-to-have thing. I'm currently running with our shells simulations on both the u250 and u280 simultaneously with the same design. One has to handle the two FPGAs differently with each having its own bitstream and flashing script. As far as I can see there is no option right now to run many workloads on different FPGAs simultaneously. 
Alternatively, one could run two run-farms in parallel one for the u250 and one for the u280 this si however currently not possible because the xdma driver is unloaded/loaded with infrasetup. I'm sure there is a reason why this is done. We have never reloaded the xdma driver though and everything works fine on our side.

I will proceed with testing of the FPGAs and going to apply a dirty quick fix that the deploy manager grabs the correct FPGA.

bgottschall:
Would it be a significant design change to run multiple bitstream builds with fewer build hosts? One could build as many bitstreams one after the other on a single machine or even allow the same machine to run many builds in parallel. 
For example, one could specify the same IP multiple times with override_build_dir pointing to different places where the builds do not interfere with each other.

bgottschall:
Building the bitstreams fails for me:
```build-bitstream.sh: line 75: vivado: command not found```

I did include in the env.sh to load vivado into the environment PATH. From the shell that I'm executing `firesim buildbitstream` vivado is in the PATH and I also added vivado into the PATH in the bashrc. Not sure from which environment firesim is executing this script but it does not carry over the environment from loaded modules. I also changed my normal shell zsh to bash but nothing helped. Whatever fabric? from python is doing it is creating shell that does not have vivado in its environment.

**Comment by abejgonzalez** [comment]
> Currently verifying this PR on our setup and while this is in progress I'm already having some comments. The PCI-E ids or BDFs should be specified per FPGA in the run-farm-recipes (externally_provisioned.yaml). I'm working on a host with 2 FPGAs, one u250 and one u280. The deploy manager currently just greps the n'th PCI-E id from the lspci command for which both FPGAs look exactly the same. This can be a fallback in case the BDFs are not specified in the run-farm however in my case I must be able to specify exactly which ID I use with which deploy scripts.

Yes. I just discovered this issue last night. This is very tricky. From what I could tell from your `fpga-util.py` script, you need to know the mapping of FPGA serial number to PCI-E BDF. How did you determine this? Just by flashing each card individually from scratch and looking at what BDF was added? Do you know if there is a way to query the PCI-E interface (either with `lspci` or some XDMA commands) that will say what FPGA serial number it is?

Right now, the FireSim manager doesn't really support having heterogeneous FPGAs on the same system (i.e. each run farm host has only 1 deploy manager). Technically you can have a "heterogeneous" deploy manager that understands that the system can have different FPGAs but we don't really support that right now.

> The next thing is a nice-to-have thing. I'm currently running with our shells simulations on both the u250 and u280 simultaneously with the same design. One has to handle the two FPGAs differently with each having its own bitstream and flashing script. As far as I can see there is no option right now to run many workloads on different FPGAs simultaneously. Alternatively, one could run two run-farms in parallel one for the u250 and one for the u280 this si however currently not possible because the xdma driver is unloaded/loaded with infrasetup. I'm sure there is a reason why this is done. We have never reloaded the xdma driver though and everything works fine on our side.

Like I said from above to get this to work, it would be a bit hacky. Multiple run farms would be a fine short term solution. I unloaded the XDMA driver just to match AWS F1's setup but I'm sure it's fine to keep loaded at all times. I can change the setup for this.

> I will proceed with testing of the FPGAs and going to apply a dirty quick fix that the deploy manager grabs the correct FPGA.

Ack. I think there are two steps to getting this fully working in the manager:

1. Determine the BDF associated with a specific FPGA serial number. Once you have the serial number then it's easy to distinguish between FPGA types. For right now, just enforce that a single run farm only has FPGA's of one type.
2. Refactor the manager code such that the run farm deploy manager (i.e. combined setup for an entire machine with multiple slots) is split into "slot deploy manager" i.e. depending on the slot (which might have a different FPGA than another slot) setup the machine for that specific slot. 

> Would it be a significant design change to run multiple bitstream builds with fewer build hosts? One could build as many bitstreams one after the other on a single machine or even allow the same machine to run many builds in parallel. For example, one could specify the same IP multiple times with override_build_dir pointing to different places where the builds do not interfere with each other.

Unfortunately, yes. Right now we send builds to a single IP address (i.e. `localhost`) so we are bounded to 1 build per machine. Similar to the run farm scenario you can have multiple different build farms pointing to different build dirs to run different builds.

> Building the bitstreams fails for me: `build-bitstream.sh: line 75: vivado: command not found`
> 
> I did include in the env.sh to load vivado into the environment PATH. From the shell that I'm executing `firesim buildbitstream` vivado is in the PATH and I also added vivado into the PATH in the bashrc. Not sure from which environment firesim is executing this script but it does not carry over the environment from loaded modules. I also changed my normal shell zsh to bash but nothing helped. Whatever fabric? from python is doing it is creating shell that does not have vivado in its environment.

FireSim expected Vivado to be sourced in your `.bashrc` or equivalent such that if you do `ssh localhost printenv` it is in your path.

**Comment by bgottschall** [comment]
Yes you are right the mapping in our fpga-util.py was handwritten by determining the BDFs and flashing one after the other or by trial and error flashing of the FPGAs and seeing which ID it was. Since this is a very setup specific part it should be put in the "FPGA and Tool Setup" and then one can specify the correlation between JTAG serials and IDs in the run-farm.

I found the issue with the environment. My environment is only set in a login shell but fabric uses a non-login shell `ssh localhost CMD` . I'll look further when I find the time.

**Comment by abejgonzalez** [comment]
> Yes you are right the mapping in our fpga-util.py was handwritten by determining the BDFs and flashing one after the other or by trial and error flashing of the FPGAs and seeing which ID it was. Since this is a very setup specific part it should be put in the "FPGA and Tool Setup" and then one can specify the correlation between JTAG serials and IDs in the run-farm.

I'm trying to script this up to be automatic right now.

> I found the issue with the environment. My environment is only set in a login shell but fabric uses a non-login shell `ssh localhost CMD` . I'll look further when I find the time.

I know that in bash you have to comment out in the `.bashrc` the following:

```
# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac
```

In the future, we can fix this up a bit more.

abejgonzalez:
Once this CI mainly passes (cpp-lint is erroring for reasons I don't know so I'll ignore it for now) I'll merge this in. We can improve the support in smaller upcoming PRs

sagark (approved):

PR #1506, opened 2023-05-11
Fix trigger system common case performance regression

sagark:
Something in the trigger system is causing performance to be significantly reduced when tracerv is included but disabled at runtime (recent CI logs on F1 show 150 MHz FPGA freq -> 40 MHz effective freq).

This PR fixes this for the common case where the outcome of evaluating the trigger will not change during the course of the simulation. In this case, we will keep driving credit/debit tokens into the sim regardless of the normal trigger evaluation dependencies, since they don't matter.

Without this fix, having tracerv in a build but disabled at runtime is significantly reducing performance:

* F1, tracerv disabled at runtime: ~150 MHz FPGA freq -> 40 MHz effective freq
* VCU118, tracerv disabled at runtime: 50 MHz FPGA freq -> 15 MHz effective freq
* VCU118, tracerv enabled at runtime (logging to disk): 50 MHz FPGA freq -> 15 MHz effective freq
* VCU118, with this fix implemented, trigger always true, tracerv disabled at runtime: 50 MHz FPGA freq -> 45 MHz effective freq
* VCU118, with this fix implemented, trigger always true, tracerv enabled at runtime (logging to disk): 50 MHz FPGA freq -> 15 MHz effective freq

Waiting on buildbitstream CI to build F1 bitstreams to report F1 performance results.

#### Related PRs / Issues

N/A

#### UI / API Impact

No API change, only performance improvement in common cases.

Does change verilog, but no compatibility issues.

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
+++ b/sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
@@ -206,13 +206,24 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
       ),
     )
 
+    // performance optimization:
+    //
+    // if any of these cases is true, just spam out hPort.fromHost.hValid=1,
+    // because triggerDebit will never be set and (if trigger result is always
+    // 1) triggerCredit will be set the first cycle then always zero.
+    //
+    // false: trigger condition is unsatisfiable
+    val trigger_always_false = (triggerSelector === 1.U) && (triggerCycleCountStart > triggerCycleCountEnd)
+    // true: triggering is not enabled (i.e. selector = 0) OR using a cycle count trigger with start=0 and end=(Largest 64 Bit UInt)
+    val trigger_always_true = (triggerSelector === 0.U) || ((triggerSelector === 1.U) && (triggerCycleCountStart === 0.U) && (triggerCycleCountEnd === (BigInt("ffffffffffffffff", 16).U(64.W))))
+
     val tFireHelper = DecoupledHelper(streamEnq.ready, hPort.toHost.hValid, hPort.fromHost.hReady, initDone)
 
     val triggerReg = RegEnable(trigger, false.B, tFireHelper.fire())
     hPort.hBits.triggerDebit  := !trigger && triggerReg
     hPort.hBits.triggerCredit := trigger && !triggerReg
 
-    hPort.fromHost.hValid := tFireHelper.fire(hPort.fromHost.hReady)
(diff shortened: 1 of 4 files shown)


sim/firesim-lib/src/main/scala/bridges/TracerVBridge.scala
```diff
@@ -206,13 +206,24 @@ class TracerVBridgeModule(key: TracerVKey)(implicit p: Parameters)
...
+    val trigger_always_true = (triggerSelector === 0.U) || ((triggerSelector === 1.U) && (triggerCycleCountStart === 0.U) && (triggerCycleCountEnd === (BigInt("ffffffffffffffff", 16).U(64.W))))
+
     val tFireHelper = DecoupledHelper(streamEnq.ready, hPort.toHost.hValid, hPort.fromHost.hReady, initDone)
 
     val triggerReg = RegEnable(trigger, false.B, tFireHelper.fire())
     hPort.hBits.triggerDebit  := !trigger && triggerReg
     hPort.hBits.triggerCredit := trigger && !triggerReg
 
-    hPort.fromHost.hValid := tFireHelper.fire(hPort.fromHost.hReady)
+    hPort.fromHost.hValid := tFireHelper.fire(hPort.fromHost.hReady) || ((trigger_always_false || trigger_always_true) && initDone)
```
davidbiancolin, line 226: I guess what i've been trying to say is that:
`tFireHelper.fire(hPort.fromHost.hReady` is likely introducing a combinational dependency on the value of a token coming in from `toHost`, so the "better" solution would be to put what essentially amounts to the `fromHost` token value of the credit and debit in a queue, and then drive hPort.fromHost.{hValid, hReady} with the head of that queue only. 

Finally you have to ensure that the queue is initialized with a single token (say the credits both set to false) after reset. 

This will be less brittle and ensure good FMR even when the trigger is being used.

abejgonzalez:
Uncaught 1 build instance shutdown(s) detected for CI run: 4950534313. Verify CI state before submitting PR.

abejgonzalez:
Uncaught 4 build instance shutdown(s) detected for CI run: 4950534313. Verify CI state before submitting PR.

abejgonzalez (approved):
Pending CI / AGFI testing... I think this solution is fine in the short term (the long-term solution that David mentioned should at least be marked with a detailed issue so that you can solve it later).

abejgonzalez:
Uncaught 5 build instance shutdown(s) detected for CI run: 4958874655. Verify CI state before submitting PR.

abejgonzalez:
Fixed by #1596

PR #1564, opened 2023-06-15
changed mmap to xdma_user instead of pci resource 

cyyself:
#### Related PRs / Issues

Recently, I tried to run firesim on Xilinx VCU118. And I am using Linux v5.4.246 for the Host and use the latest xdma and xvsec driver, which is v2023.1.3 from [Xilinx/dma_ip_drivers](https://github.com/Xilinx/dma_ip_drivers).

However, an error message shows `assert(bar0_base != MAP_FAILED);` failed in `sim/midas/src/main/cc/simif_xilinx_alveo_u250.cc`. I tried a simple C program to test mmap, then found that the user pcie bar can only be mapped when the xdma driver is not loaded.

After reviewing the xdma driver to see why it happens, I found that the xdma driver provides a way to the mmap user bar by mmap `/dev/xdma_user` char dev. So I change the code, and it works.

#### UI / API Impact

No change.

No change.

### DIFF
--- a/sim/midas/src/main/cc/simif_xilinx_alveo_u250.cc
+++ b/sim/midas/src/main/cc/simif_xilinx_alveo_u250.cc
@@ -246,27 +246,10 @@ void simif_xilinx_alveo_u250_t::fpga_setup(uint16_t domain_id,
   assert(ret >= 0);
   fpga_pci_check_file_id(sysfs_name, pci_device_id);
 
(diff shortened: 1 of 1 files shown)


sagark:
Thanks for this, looks good to me! I will let CI run and then approve/merge.

sagark (approved):
LGTM, thanks!

PR #1704, opened 2024-03-14
Add initial implementation of variable FESVR rate during loading phase of simulation

abejgonzalez:
Generally speaking, there are two phases of simulation: (1) initial loading of payloads/binaries + (2) starting execution of the SoC (i.e. the workload). Right now during (1) we use the LoadMem widget to load the main target binary. However, there are cases where we can load other payloads (i.e. `+payload=...` flag in FESVR) that don't use the LoadMem widget. In these cases, the main target binary will load "fast" (i.e. bypassing the target SoC memory fabric directly to DRAM) but the other payloads will load using the traditional TSI xacts to the target memory fabric (this is slow). To help speed this 2nd use case up, this PR adds a "fast FESVR" mode where the FESVR calls at phase (1) are sped up. This ensures that FESVR pushes as many R/W xacts to the target as fast as possible (as fast as the target can sink these xacts). Then once phase (1) is over, it reverts to the user-specified FESVR step size.

Two new simulator flags:
- `+fesvr-disable-early-fast`. boolean control over this new functionality which is enabled by default
- `+fesvr-wait-ticks`. wait N FESVR ticks before driving FESVR xacts. this fixes a corner-case where FESVR could drive xacts from the host into the bridge HW queue, then the reset from the target would clear the bridge HW queue forming malformed TSI xacts. This only occurs when the FESVR step size is low (I tested 8 only). This is set to a default value of 8 (wait 8 TSI ticks) which seems to work well with a small step size of 8).

Side Note:
The `firesim_tsi_t::reset` is overridden to allow some synchronization between the bridge tick and the host thread. In essence we wait for all FESVR xacts from phase (1) to complete before switching to the faster step size.

### Contributor Checklist

### DIFF
--- a/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.cc
+++ b/sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.cc
@@ -1,12 +1,13 @@
 // See LICENSE for license details
-#include <stdio.h>
-#include <inttypes.h>
 #include "firesim_tsi.h"
+#include <inttypes.h>
+#include <stdio.h>
 
 #define fprintf(stdout, fmt, ...) (0)
 
 firesim_tsi_t::firesim_tsi_t(int argc, char **argv, bool can_have_loadmem)
-    : testchip_tsi_t(argc, argv, can_have_loadmem), is_busy(false), is_loaded(false), is_loaded_in_sw(false) {
+    : testchip_tsi_t(argc, argv, can_have_loadmem), is_busy(false),
+      is_loaded_in_host(false), is_loaded_in_target(false) {
   idle_counts = 10;
   std::vector<std::string> args(argv + 1, argv + argc);
   for (auto &arg : args) {
@@ -30,7 +31,11 @@ void firesim_tsi_t::send_loadmem_word(uint32_t word) {
 void firesim_tsi_t::load_mem_write(addr_t addr,
                                    size_t nbytes,
                                    const void *src) {
-  fprintf(stdout, "firesim_tsi_t::load_mem_write addr: %" PRIx64 " nbytes: %" PRIu64 "\n", addr, nbytes);
+  fprintf(stdout,
+          "firesim_tsi_t::load_mem_write addr: %" PRIx64 " nbytes: %" PRIu64
+          "\n",
+          addr,
+          nbytes);
   fflush(stdout);
 
   loadmem_write_reqs.push_back(firesim_loadmem_t(addr, nbytes));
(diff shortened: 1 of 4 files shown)


sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
```diff
@@ -48,5 +52,7 @@ class firesim_tsi_t final : public testchip_tsi_t {
 private:
   size_t idle_counts;
   bool is_busy;
+  bool is_loaded;
```
jerryz123, line 55: Can you add some comments describing what these new fields indicate?

jerryz123: I think perhaps a better naming is `is_loaded_in_dut`/`is_loaded_in_fesvr`? Not sure

abejgonzalez: Good catch, yes the names should change. I think to something like `fesvr_pre_reset_xacts_sent` and `target_synced_all_xacts`.

abejgonzalez:
I should note that in upstream FireSim this is a NOOP (except for the wait ticks work). All xacts done in the `htif::load_program` phase are sent to the loadmem widget because they assume that all payloads (target binary and extra payloads) are in the memory space that the loadmem widget controls (not true in multi-SoC scenarios). Thus there will be no non-loadmem xacts sent before the fesvr step size is sped up again.

sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.cc
```diff
@@ -1,8 +1,12 @@
 // See LICENSE for license details
+#include <stdio.h>
+#include <inttypes.h>
 #include "firesim_tsi.h"
 
+#define fprintf(stdout, fmt, ...) (0)
```
jerryz123, line 6: Are you planning on leaving this debug feature in?

sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.cc
```diff
@@ -48,6 +58,15 @@ void firesim_tsi_t::load_mem_read(addr_t addr, size_t nbytes, void *dst) {
 
 void firesim_tsi_t::tick() { switch_to_host(); }
 
+void firesim_tsi_t::reset() {
+  is_loaded_in_sw = true;
```
jerryz123, line 62: I think a comment here would be useful. This loop spins in the fesvr thread until the target has indicated that it is loaded.

sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.h
joonho3020: Maybe `is_loaded_in_host` and `is_loaded_in_target`?

joonho3020 (commented):
Nit : I feel a bit uneasy about this being the default mode because of the race condition thing. What about we make this an option and document it somewhere?

sim/firesim-lib/src/main/cc/fesvr/firesim_tsi.cc
abejgonzalez: Yes I think not having the prints print is nice (with users having the option to reenable).

joonho3020 (approved):
LGMT

abejgonzalez:
@joonho3020 I assume the CI is broken so I can ignore the CI failures right now?

joonho3020:
I thought that `run-parallel-vcs-metasims-and-vitis-driver` and `run-vitis-check-docs...` should pass though. Seems like the ssh issue on the CI machines?

abejgonzalez:
Note: This is disabled by default.

PR #1727, opened 2024-04-30
Reduce AXI DRAM channels for Alveo U2*0 cards

abejgonzalez:
Downstream work is planning on reworking the build process a bit. To make this easier, this PR reduces the number of AXI DRAM channels the FireSim sim can use from 4 to 1. To most simulations, this won't affect anything unless the user requests more than 16GB of target memory (which is the default on the CY side)

### Contributor Checklist


joonho3020 (commented):
What if we want to run simulations that requires 64GB of backing DRAM?

abejgonzalez:
> What if we want to run simulations that requires 64GB of backing DRAM?

Then this will error at elaboration time. I think we can re-add this feature later once some other projects are upstreamed first. Most folks I don't think use more that 16GB (they just use the default).

joonho3020 (approved):

PR #1791, opened 2024-08-09
FireSim x Chipyard Decoupled

abejgonzalez:
Sister PR: https://github.com/ucb-bar/chipyard/pull/1984

Originally FireSim and Chipyard were coupled sharing things like (1) Scala sources (and SBT sources) (2) Conda environments (3) Bridges (...) and more. However, this PR decouples FireSim from Chipyard s.t. all "target" (ie. DUT) specific features are located in Chipyard (ie. things like FireSim UART bridges, NIC bridges, etc) and FireSim provides a hook to Chipyard so that I can run the FireSim makefile itself.

This includes the following:

- Target-specific CI is moved to Chipyard
- Target-specific makefrags (the API that is exposed to Chipyard to build things with FireSim) is now moved to Chipyard (this includes the `bridges` and `firesim` makefrags). Additionally, target-specific bridges are moved to Chipyard (including all their collateral).
- FireSim has it's own copies of Scala sources that once existed in Chipyard (RocketChip, Hardfloat, Diplomacy, CDE). This also changes the `build.sbt` to depend on those sources exclusively.
- Things like the DDR configuration configs now only exist on the FireSim-side. This requires build recipes to change s.t. those configs are appended to the `PLATFORM_CONFIG` (i.e. `DDR3FRFCFSLLC4MB` needs to be added to the platform configs). This also implies that the configuration for the memory model applies to all ports coming off the target (in practice this seems to be what people do). 
- Scalafix is now updated to fix more things
- Nasti (the ugly AXI copy) now doesn't rely on `implicit p: Parameters` instead it is explicitly parameterized without implicits. This makes things more verbose but reduces the amount of globals present.
- A few specific bridges are kept in `firesim-lib` s.t. Chipyard can use them.
- `firesim-lib` now provides all APIs that Chipyard needs to build it's own bridges, build testing infra. and more. This should be minimal and be cross compilable with Chisel6 and old Chisel. Eventually this will use firrtl2.
- The `BridgeAnnotation` used to define a bridge now takes a `string` referring to a class in the `midas` SBT project. This allows decoupling of FireSim SBT sources from Chipyard sources when creating a bridge at the cost of less type safety.
- New API is added for make called `symlink_sources_and_clear` allowing downstream projects (i.e. Chipyard) to symlink sources into midas cleanly.

_Bridge APIs/Setup:_
Now that bridges can live in the target, we have APIs exposed to the target to compile a bridge. It is roughly as follows (documentation will come later) and is assuming Chipyard for this explanation.

Bridges are composed of N parts.
  * (1) Items shared between Chipyard and FireSim. This includes all case classes and IOs. These need to be fully isolated from the target (i.e. they can cross compile between Chisel3 and Chisel6, and do not depend on any target-specific classes/IOs/etc).
  * (2) The bridge stub that lives in the target. This references the items in (1) so that it can create and dump a BridgeAnnotation. This depends on `firesim-lib` which provides the necessary FIRRTL classes, scala code, etc. This is only compiled in the target.
  * (3) The bridge implementation that lives in the target *but is compiled in FireSim-only* (the target does not compile this). This is symlinked into the midas compiler so it can use `midas` APIs.
  * (4) The FireSim build system makefrag hooks (this is a set of `build/config/driver/metasim.mk` files) to which the user of FireSim needs to point to with `TARGET_PROJECT_MAKEFRAG` when invoking FireSim's make. These files copy all relevant files into midas, build the target RTL, setup configs, and more.

#### Related PRs / Issues

See above.

#### UI / API Impact

Target configs now do not reference any memory model-related configs. All memory model-related configs are to be appended to the `PLATFORM_CONFIG`.

Should be the same. 

#### Contributor Checklist

#### Reviewer Checklist (only modified by reviewer)

#### CI Help
Add the following labels to modify the CI for a set of features.
Generally, a label added only affect subsequent changes to the PR (i.e. new commits, force pushing, closing/reopening).
See `ci:*` for full list of labels:
- `ci:fpga-deploy` - Run FPGA-based E2E testing
- `ci:local-fpga-buildbitstream-deploy` - Build local FPGA bitstreams for platforms that are released
- `ci:persist-prior-workflows` - Prevent prior CI workflows from automatically cancelling with subsequent changes
- `ci:disable` - Disable CI
- `ci:disable-scala-tests` - Disable Scala tests in CI


abejgonzalez:
This PR does not work.

sim/midas/src/main/scala/junctions/Nasti2AXI4.scala
```diff
@@ -6,31 +6,32 @@ import chisel3._
...
   AXI4NastiAssigner.toAXI4Slave(io.axi4, io.nasti)
 }
 
-class Nasti2AXI4Monitor(params: AXI4BundleParameters)(implicit p: Parameters)
+class Nasti2AXI4Monitor(params: AXI4BundleParameters)
     extends RawModule with InlineInstance {
   val io = IO(new Bundle {
     val axi4 = Output(new AXI4Bundle(params))
-    val nasti = Input(new NastiIO()(p alterPartial { case NastiKey => NastiParameters(params) } ))
+    val nasti = Flipped(new NastiIO(CreateNastiParameters(params)))
```
abejgonzalez, line 34: TODO: Revert.

sim/midas/src/main/scala/midas/models/dram/BankConflictModel.scala
```diff
@@ -67,7 +72,7 @@ object BankConflictConstants {
   val bankIdle :: bankBusy :: bankPrecharge :: Nil = Enum(nBankStates)
 }
 
-
+// TODO: TimingModel/SplitTransactionModel's children should directly use p(NastiKey)
```
abejgonzalez, line 75: TODO: Remove.

sim/midas/src/main/scala/midas/widgets/Widget.scala
```diff
@@ -146,13 +150,11 @@ abstract class WidgetImp(wrapper: Widget) extends LazyModuleImp(wrapper) {
     val reg         = RegInit(default)
     val shadowReg   = Reg(default.cloneType)
     shadowReg.suggestName(s"${name}_mmreg")
-    val baseAddr    = Seq
```
abejgonzalez, line 149: TODO: Revert

abejgonzalez:
This is ready for review (it *should* pass CI tests soon)

joonho3020 (commented):
Can we try linux power off on F1 since this is a pretty big PR?
On a cursory look things seems fine to me, but I would like to see it being tested before clicking approve.
And of course, I'll take some more time to actually read through things.

.gitmodules
```diff
@@ -22,3 +22,15 @@
 [submodule "sim/toolchain-utilities/riscv-tests"]
 	path = sim/toolchain-utilities/riscv-tests
 	url = https://github.com/riscv-software-src/riscv-tests
+[submodule "sim/rocket-chip"]
+	path = sim/rocket-chip
+	url = https://github.com/chipsalliance/rocket-chip.git
+[submodule "sim/berkeley-hardfloat"]
```
joonho3020, line 28: Stupid question: did FireSim have a hardfloat dependency?

deploy/util/targetprojectutils.py
```diff
@@ -0,0 +1,14 @@
...
+    Eventually will be replaced by using YAML hooks.
+
+    Args:
+        targetproject: make arg corresponding to TARGET_PROJECT.
+        deploydir: Deploy directory of FireSim
+
+    Returns:
+        any extra args needed for FireSim's make command to function
+    """
+    if targetproject in ["firesim", "bridges"]:
```
joonho3020, line 12: So just to check my understanding...
If you run manager commands (e.g. `firesim infrasetup`), it calls into the makefiles inside chipyard to generate the target design + bridges to generate the FIRRTL & annotations, and then you call midas inside FireSim?

sim/firesim-lib/src/main/scala/bridges/PeekPokeBridge.scala
joonho3020, line 1: If we are operating under the assumption that all the target design specific bridges (NIC, CoSpike, BlockDev...) lives in Chipyard while bridges common to all targets (Reset, Clock, PeekPoke, FASED) lives inside FireSim, can we just remove the `firesim-lib` directory completely and place these bridges in `midas/src/main/scala/midas/widgets` ?

joonho3020: Or maybe we can move `firesim-lib` into `midas/src/main/scala/firesim-lib`?
I don't think I understand what the difference between the `firesim-lib` and `midas` directories.

joonho3020, line 1: So to my understanding, `firesim-lib` should be providing APIs to the target (CY) world for generating bridge APIs.
Can we move all these bridges inside `midas` and keep this `firesim-lib` interface minimal?

abejgonzalez:
> Can we try linux power off on F1 since this is a pretty big PR? On a cursory look things seems fine to me, but I would like to see it being tested before clicking approve. And of course, I'll take some more time to actually read through things.

Yup. I can try that right now. I think (keyword think) this won't affect Linux boot since in the CI it runs baremetal tests (Dhrystone + rv64ui-p-simple) on Rocket + Boom based configs (which are passing).

.gitmodules
abejgonzalez: This just copies how Chipyard set up it's RC SBT project. I'm not sure why Chipyard had it's own hardfloat but I figured I would keep it the same.

deploy/util/targetprojectutils.py
abejgonzalez: `firesim infrasetup` calls the `firesim/sim/Makefile`. To build RTL for Chipyard, it needs to have a reference to the makefrags in Chipyard. Those makefrags tell the FireSim `Makefile` what to do to get `.fir/.anno.json` and how to use the bridges it has. Those makefrags call the Chipyard `Makefile` directly to get the FIRRTL. So the order is:

`firesim/sim/Makefile` -> Chipyard MakeFrags -> Chipyard `Makefile` -> Got FIRRTL/ANNOs here -> Back to `firesim/sim/Makefile` to finish GG compilation.

sim/firesim-lib/src/main/scala/bridges/PeekPokeBridge.scala
abejgonzalez: This is as minimal as I could get it. Technically everything in that folder is needed to create `BridgeAnnotations` (which are the true API between FireSim/Chipyard).

deploy/util/targetprojectutils.py
joonho3020: Yikes

joonho3020:
Overall, looks fine to me pending CI + f1 runs

joonho3020 (approved):
Approved pending CI / F1 check

